PR #114, opened 2024-03-29
Add German translation for MoneyPrinterTurbo

Sw0rdf1sh1:
Add a valid German Translation. + Fix Subtile Middle -> Center translate var.

Cheers.


webui/i18n/de.json
```diff
@@ -0,0 +1,51 @@
+{
+  "Language": "German",
+  "Translation": {
+    "Video Script Settings": "**Video Einstellungen**",
+    "Video Subject": "Worum soll es in dem Video gehen? (Geben Sie ein Keyword an, :red[Dank KI wird automatisch ein Drehbuch generieren])",
+    "Script Language": "Sprache zum Generieren von Drehbüchern (KI gibt automatisch basierend auf der Sprache Ihres Keywords aus)",
+    "Generate Video Script and Keywords": "Klicken Sie hier, um mithilfe von KI ein „Drehbuch“ und „Video-Keywords“ basierend auf dem **Keyword** zu generieren.",
```
danicruz0415, line 7: I am not sure if this characters need to be changed. I see that they are kept in the other translations as well as in one line of your translation. Maybe to keep it consistent it could be better to keep them like they are in the other languages. What do you think?

image

Sw0rdf1sh1: @danicruz0415 I see, you're right I've fixed that :)
For consistency I should also use the parentheses

Thanks!

harry0703:
Thank you very much.

PR #146, opened 2024-04-02
feat: add Dockerfile

KevinZhang19870314:


harry0703:
thanks 👍

Ashes47:
Can we add like another container and also deploy streamlit frontend in docker?
Maybe with a simple auth?
Happy to colloborate for this :)

KevinZhang19870314:
@Ashes47 Once [157](https://github.com/harry0703/MoneyPrinterTurbo/pull/157) merged, you can run `docker-compose up` to launch both the web interface and api service.

> Maybe with a simple auth?

I think this can be done at streamlit part, write a simple auth key logic as a wrapper of current streamlit app.  @harry0703

PR #285, opened 2024-04-18
Update requirements.txt

vuisme:
Update streamlit to 1.33.0 Fix error when install
```
streamlit 1.32.0 depends on packaging<24 
git-changelog 2.5.2 depends on packaging>=24.0
```
Work well with me


harry0703:
thanks

PR #504, opened 2024-09-28
Fix the response 403 from pexels - search_videos_pexels - failed to download videos, maybe the network is not available. if you are in China, please use a VPN.

Dreyfi:
search_videos_pexels - failed to download videos, maybe the network is not available. if you are in China, please use a VPN.

### DIFF
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -161,11 +161,15 @@ def save_video(video_url: str, save_dir: str = "") -> str:
         logger.info(f"video already exists: {video_path}")
         return video_path
 
+    headers = {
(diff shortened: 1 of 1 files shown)


LucasHenriqueDiniz:
That fixed the problem getting the videos, but I still had problems downloading them too. This was solved when I added "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36" to the save_video function as well.

Dreyfi:
@LucasHenriqueDiniz Thank you for your test. I just updated the PR.  Cheers

PR #698, opened 2025-05-18
Add API management module

WITHOUTNAMESES:
The main interface now includes API management, making it easier for users to add, delete, and update their API keys. The feature has been localized and supports all five languages in the i18n setup.


drewsteele408:
Thank you for adding these changes.

harry0703:
Thanks for the contribution!  

PR #715, opened 2025-05-25
feat: add custom audio file support

michaeltmk:
allowing user to reuse an local audio file.
If using local audio file, subtitles will be disabled.


jansh7784 (approved):

harry0703:
Thanks for the contribution!  

PR #751, opened 2025-07-04
第一次提交

yanjianzao:
第一次提交

### DIFF
--- a/app/controllers/v1/llm.py
+++ b/app/controllers/v1/llm.py
@@ -38,8 +38,7 @@ def generate_video_script(request: Request, body: VideoScriptRequest):
 def generate_video_terms(request: Request, body: VideoTermsRequest):
     video_terms = llm.generate_terms(
         video_subject=body.video_subject,
-        video_script=body.video_script,
-        amount=body.amount,
+        video_script=body.video_script
     )
     response = {"video_terms": video_terms}
     return utils.get_response(200, response)
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -45,7 +45,7 @@ def search_videos_pexels(
         "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36",
     }
     # Build URL
-    params = {"query": search_term, "per_page": 20, "orientation": video_orientation}
+    params = {"query": search_term, "per_page": 20, "orientation": video_orientation,"orientation":"landscape","size":"large"}
     query_url = f"https://api.pexels.com/videos/search?{urlencode(params)}"
     logger.info(f"searching videos: {query_url}, with proxies: {config.proxy}")
 
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -45,7 +45,7 @@ def search_videos_pexels(
         "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36",
     }
     # Build URL
-    params = {"query": search_term, "per_page": 20, "orientation": video_orientation,"orientation":"landscape","size":"large"}
+    params = {"query": search_term, "page": 1, "per_page": 80, "orientation": "landscape", "size": "medium","locale":"en-US"}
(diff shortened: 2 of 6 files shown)


app/controllers/v1/llm.py
```diff
@@ -38,8 +38,7 @@ def generate_video_script(request: Request, body: VideoScriptRequest):
 def generate_video_terms(request: Request, body: VideoTermsRequest):
```
starkshade, line 38: ## ⚠️ Code Issue

**Lines:** 38-44

```python
def generate_video_terms(request: Request, body: VideoTermsRequest):
    video_terms = llm.generate_terms(
        video_subject=body.video_subject,
        video_script=body.video_script
    )
    response = {"video_terms": video_terms}
    return utils.get_response(200, response)
```

The removal of the `amount` parameter from the `llm.generate_terms` function call, without apparent justification in the diff, raises a maintainability concern. It's unclear if this parameter is no longer needed or if its omission is an oversight. This could lead to bugs if the `llm.generate_terms` function relies on this parameter.

app/services/material.py
```diff
@@ -45,7 +49,7 @@ def search_videos_pexels(
         "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36",
     }
     # Build URL
-    params = {"query": search_term, "per_page": 20, "orientation": video_orientation}
+    params = {"query": search_term, "page": 1, "per_page": 5, "orientation": video_orientation, "size": "large","locale":"en-US"}
```
starkshade, line 52: ## ⚠️ Code Issue

**Lines:** 52-52

```python
params = {"query": search_term, "page": 1, "per_page": 5, "orientation": video_orientation, "size": "large","locale":"en-US"}
```

This code snippet hardcodes pagination (`page: 1`, `per_page: 5`) and size (`size: "large"`) parameters in the `search_videos_pexels` function. This limits the number of results fetched and the quality of videos that can be selected, reducing the flexibility and effectiveness of the search. It would be better to make these parameters configurable or dynamically determined.

app/services/material.py
```diff
@@ -62,26 +66,58 @@ def search_videos_pexels(
...
+        
         videos = response["videos"]
-        # loop through each video in the result
+        
         for v in videos:
-            duration = v["duration"]
-            # check if video has desired minimum duration
-            if duration < minimum_duration:
+            duration = v.get("duration")
+            if not duration or duration < minimum_duration:
```
starkshade, line 74: ## ⚠️ Code Issue

**Lines:** 74-75

```python
if not duration or duration < minimum_duration:
    continue
```

In `search_videos_pexels`, this code checks if `duration` is less than `minimum_duration`. However, it does not explicitly handle the case where `duration` might be `None`. If `duration` is `None`, the comparison `duration < minimum_duration` will raise a `TypeError`. The preceding `if not duration` check should ideally catch this, but it's safer to ensure `duration` is a valid number before comparison.

**Fix:**
Ensure `duration` is a valid number before performing the comparison, or rely on the `if not duration` check to correctly handle `None` values.

app/services/material.py
```diff
@@ -62,26 +66,58 @@ def search_videos_pexels(
...
+
+                # Ensure width and height are available
+                if not width or not height:
+                    continue
+
+                # Check if the video orientation matches the requirement
+                is_landscape = width > height
+                is_portrait = height > width
+
+                if video_aspect == VideoAspect.landscape and not is_landscape:
```
starkshade, line 96: ## ⚠️ Code Issue

**Lines:** 96-99

```python
if video_aspect == VideoAspect.landscape and not is_landscape:
    continue
if video_aspect == VideoAspect.portrait and not is_portrait:
    continue
```

In `search_videos_pexels`, this aspect ratio filtering logic might incorrectly exclude square videos. If `width == height`, then `is_landscape` and `is_portrait` will both be `False`. Consequently, a square video will be skipped if `video_aspect` is `landscape` (because `not is_landscape` is true) and also skipped if `video_aspect` is `portrait` (because `not is_portrait` is true).

**Fix:**
Adjust the logic to correctly handle square videos. For example, consider square videos as fitting both landscape and portrait criteria, or add a specific check for square videos if they should be treated differently.

harry0703:
Validated against the current main baseline on April 2, 2026. This PR is in DIRTY merge state and cannot be merged cleanly onto the current base. Closing it because it does not pass merge validation.

PR #778, opened 2025-09-26
Optimize Dockerfile for China-friendly builds with fallback mirrors

Mxucc:
Optimize Dockerfile for China-friendly builds with fallback mirrors

- Prioritize Aliyun and Tsinghua mirrors for apt-get and pip to ensure fast and stable builds in China without VPN.
- Implement retry logic (3 attempts, 60s timeout) for both apt-get and pip to handle network instability.
- Add fallback to official Debian and PyPI sources for global compatibility.
- Retain arm64 support and ImageMagick security policy fix.
- Keep original structure with minimal changes for maintainability.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,12 +9,40 @@ RUN chmod 777 /MoneyPrinterTurbo
 
 ENV PYTHONPATH="/MoneyPrinterTurbo"
 
-# Install system dependencies
-RUN apt-get update && apt-get install -y \
-    git \
-    imagemagick \
-    ffmpeg \
-    && rm -rf /var/lib/apt/lists/*
+# Install system dependencies with domestic mirrors first for stability
+RUN echo "deb http://mirrors.aliyun.com/debian bullseye main" > /etc/apt/sources.list && \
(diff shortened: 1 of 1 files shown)


andrewpjlewis:
Nice work on making the Dockerfile more accessible. I’m new to Docker, so I’m curious: what exactly makes these builds “China-friendly,” and how do they compare to builds in other regions?

Mxucc:
> Nice work on making the Dockerfile more accessible. I’m new to Docker, so I’m curious: what exactly makes these builds “China-friendly,” and how do they compare to builds in other regions?

Thanks! The "China-friendly" part basically means we're using fast, local mirrors in China (like Aliyun and Tsinghua) for downloading packages during the build. In China, the Great Firewall can make pulling from default global servers super slow or flaky, so this speeds things up a ton and adds retries/failovers to make it reliable.
Compared to other regions? Builds there usually just use the standard Debian or PyPI servers without issues—no need for extras since connections are stable and quick. This setup falls back to defaults anyway if the mirrors fail, so it works everywhere but shines in China.

harry0703:
Thanks for the contribution!  

PR #826, opened 2026-02-24
[NEW FEATURE] Jimeng AI adaptation

LunaticLegacy:
Literally.
I just finished this...

(And modified almost ALL the spaghetti codes that without typing annotation, dataclass, etc.)

Fix 1: Fixed the issue that there's only one video will be generated and repeat from begin to end.

### DIFF
--- a/app/jimeng_video_handler.py
+++ b/app/jimeng_video_handler.py
@@ -94,8 +94,7 @@ class JiMengVideoHandler:
                 return get_result.content.video_url
             
             else:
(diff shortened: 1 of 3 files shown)


harry0703:
Validated against the current main baseline on April 2, 2026. This PR no longer passes merge validation on the current base: it conflicts with the current app/services/voice.py and requirements changes, including the local edge_tts 7.x synthesis/subtitle compatibility fix already applied to main. Closing it for now because it cannot be merged cleanly and safely onto the current baseline.

PR #827, opened 2026-02-24
fix: edge tts v1 fixed

mehmetgencdal:


harry0703:
Triage note after reviewing this branch against the current `main`: I do not recommend reopening or merging this patch in its current form.

The main reason is scope and regression risk. Besides the Edge TTS changes, this branch also replaces the existing subtitle post-processing with raw `sub_maker.get_srt()` output and broadens unrelated dependency versions (`fastapi`, `uvicorn`). That makes the patch much larger than the reported fix and increases the chance of behavior changes in subtitle formatting and runtime compatibility.

This branch also appears to be superseded by smaller follow-up fixes that are already in `main` (for example the dependency updates merged separately). There are no targeted tests here proving that subtitle generation still matches the current punctuation-aware flow.

Recommendation: keep this PR closed. If there is still an Edge TTS compatibility issue to address, please open a new minimal PR on top of the latest `main` with focused changes and regression tests for both audio generation and subtitle output.

PR #841, opened 2026-03-23
feat: Add gTTS as an alternative to Azure TTS and fix missing pydub dependency

liangjian704-rgb:
Currently, the Azure TTS v1 (`edge-tts`) is frequently throwing `403 Forbidden` errors due to recent Microsoft API restrictions. 

This PR:
- Adds `gTTS` (Google TTS) as a reliable and free fallback option for voice generation.
- Adds `gTTS` and `pydub` to `requirements.txt` to fix the `ModuleNotFoundError` crash that occurs when the app tries to process alternative TTS audio.
- Updates the relevant audio processing logic to support the new provider smoothly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ python-multipart==0.0.19
 pyyaml
 requests>=2.31.0
 pydub
+gtts>=2.5.0


harry0703:
Rejected after integration validation: this PR conflicts with the current TTS/subtitle fix baseline in app/services/voice.py and cannot be merged cleanly without a manual rework. The missing pydub dependency is already covered by smaller, cleaner changes.

PR #887, opened 2026-05-18
fix: guard against None message in LiteLLM response path

qizwiz:
## What

Add a `message is None` check in `app/services/llm.py` after the existing choices guard.

## Why

The existing guard:
```python
if not response or not getattr(response, "choices", None):
    raise ValueError(...)
content = response.choices[0].message.content
```

ensures choices is non-empty, but if `choices[0].message` is `None` (observed on Gemini via OpenAI-compatible endpoint on PROHIBITED_CONTENT, which returns HTTP 200 with `message: null`), accessing `.content` raises `AttributeError`.

## Fix

```python
if not response or not getattr(response, "choices", None):
    raise ValueError(f"[{llm_provider}] returned empty response")
if response.choices[0].message is None:
    raise ValueError(f"[{llm_provider}] returned empty message")
content = response.choices[0].message.content
```

No behaviour change for well-formed responses.

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -366,6 +366,8 @@ def _generate_response(prompt: str) -> str:
 
                 if not response or not getattr(response, "choices", None):
                     raise ValueError(f"[{llm_provider}] returned empty response")
+                if response.choices[0].message is None:
+                    raise ValueError(f"[{llm_provider}] returned empty message")
 
                 content = response.choices[0].message.content
                 return _normalize_text_response(content, llm_provider)


harry0703:
Thanks for the clear report and patch. The issue is valid.

I implemented the fix on `main` in commit 6c89e90 with a small adjustment to match the current code structure:

- LiteLLM now preserves the existing empty-response behavior for missing/empty `choices`.
- The branch now reuses the shared `_extract_chat_completion_text()` helper, which already validates `message is None` and empty content.
- Added a regression test for `choices[0].message is None` so this does not regress.

Validation performed:

- `uv run python -m unittest test.services.test_llm`
- `uv run python -m py_compile app/services/llm.py test/services/test_llm.py`
- `git diff --check`

I am closing this PR because the fix is now applied on `main` with tests.

harry0703:
Closed as addressed by 6c89e90 on main. Thanks again for the report and patch.

PR #889, opened 2026-05-20
fix: ensure audio_clip is properly closed in combine_videos

republic117:
## Summary
Fix a file descriptor leak in `combine_videos()` where `AudioFileClip` was created but never explicitly closed on certain exit paths.

## Bug Details
- **File:** `app/services/video.py`
- **Function:** `combine_videos()`
- **Issue:** `audio_clip = AudioFileClip(audio_file)` creates a file handle that was only closed at the end of the function, but early returns and exceptions could bypass the cleanup.

## Fix
Wrap the `audio_clip` usage in a `try/finally` block to ensure `close_clip(audio_clip)` is always called.

## Testing
- [ ] Tested with sample video generation
- [ ] Verified no regression in video concatenation

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -263,8 +263,11 @@ def combine_videos(
     max_clip_duration: int = 5,
     threads: int = 2,
 ) -> str:
-    audio_clip = AudioFileClip(audio_file)
-    audio_duration = audio_clip.duration
+        audio_clip = AudioFileClip(audio_file)
+    try:
+        audio_duration = audio_clip.duration
+    finally:
+        close_clip(audio_clip)
     logger.info(f"audio duration: {audio_duration} seconds")
     logger.info(f"maximum clip duration: {max_clip_duration} seconds")
 


harry0703:
Thanks for the PR. The resource-leak concern is valid: `combine_videos()` only needs `AudioFileClip` to read the narration duration, so that clip should be closed immediately after probing the duration.

I could not merge this PR as-is because the patch introduces an indentation error around `audio_clip = AudioFileClip(audio_file)`, which would break `app/services/video.py` syntax.

I implemented the equivalent fix on `main` in commit 79d7870 with a smaller scope:

- Read `audio_clip.duration` inside `try`.
- Always call `close_clip(audio_clip)` in `finally`.
- Close the audio clip immediately after duration probing because it is not used later in `combine_videos()`.
- Added a regression test that verifies the audio clip is closed even when reading `duration` raises an exception.

Validation performed:

- `uv run python -m unittest test.services.test_video.TestVideoService.test_combine_videos_closes_audio_clip_when_duration_read_fails test.services.test_video.TestVideoService.test_open_video_clip_quietly_suppresses_moviepy_stdout`
- `uv run python -m py_compile app/services/video.py test/services/test_video.py`
- `git diff --check`

Closing this PR as addressed by 79d7870. Thanks again for catching the leak.

harry0703:
Closed as addressed by 79d7870 on main. The original patch had an indentation error, so I applied the equivalent fix with regression coverage.

PR #890, opened 2026-05-21
fix: Redis pagination returns wrong results in get_all_tasks

republic117:
## Summary
Fix incorrect pagination logic in `RedisState.get_all_tasks()` that causes pages to return empty or partial results.

## Bug Details

`get_all_tasks()` uses a cumulative `total` counter for slice index calculations. When Redis SCAN returns data in multiple batches, the slice indices are computed against the cumulative total instead of the current batch's start position.

### Reproduction
With 18 tasks in Redis, `page_size=10`:
- **Page 1** (start=0, end=10): returns **0 results** ❌ (should be 10)
- **Page 2** (start=10, end=20): returns **2 results** ❌ (should be 8)

### Root Cause
```python
# Old: total is cumulative, slice uses wrong base
total += len(keys)
for key in keys[max(0, start - total):end - total]: # ← wrong

### DIFF
--- a/app/services/state.py
+++ b/app/services/state.py
@@ -70,22 +70,27 @@ class RedisState(BaseState):
         end = start + page_size
         tasks = []
         cursor = 0
-        total = 0
+        batch_start = 0                        # ← 追踪本批次起始索引
+        
         while True:
             cursor, keys = self._redis.scan(cursor, count=page_size)
-            total += len(keys)
-            if total > start:
-                for key in keys[max(0, start - total):end - total]:
-                    task_data = self._redis.hgetall(key)
-                    task = {
-                        k.decode("utf-8"): self._convert_to_original_type(v) for k, v in task_data.items()
-                    }
-                    tasks.append(task)
-                    if len(tasks) >= page_size:
-                        break
-            if cursor == 0 or len(tasks) >= page_size:
+            batch_len = len(keys)
+            batch_end = batch_start + batch_len  # ← 本批次结束索引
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the PR. The Redis pagination issue is valid: the old implementation used the cumulative `total` value to slice the current SCAN batch, which can return empty or partial pages when Redis returns keys in multiple batches.

I could not merge this PR as-is for two reasons:

- It still includes the earlier `combine_videos()` commit from #889, which conflicts with the current `main` branch and contains an indentation bug.
- The `app/services/state.py` patch contains a literal `...` placeholder inside the loop, so it would break the implementation if merged directly.

I implemented the equivalent Redis pagination fix on `main` in commit 4cdf511:

- Track `batch_start` and `batch_size` for each Redis SCAN batch.
- Slice only the overlapping part of the current batch for the requested page.
- Continue scanning until `cursor == 0` so the returned total count is accurate.
- Added a fake Redis regression test for the 18 tasks / page_size=10 case described in the PR.

Validation performed:

- `uv run python -m unittest test.services.test_state`
- `uv run python -m py_compile app/services/state.py test/services/test_state.py`
- `git diff --check`

Closing this PR as addressed by 4cdf511. Thanks for identifying the pagination bug.

harry0703:
Closed as addressed by 4cdf511 on main. The original branch contains conflicts and a placeholder in the patch, so I applied the equivalent fix with regression coverage.

PR #902, opened 2026-05-28
Refactor video task tests with dynamic parameters

krataratha:


harry0703:
Thanks for the contribution. I tested this patch and it currently cannot be merged because the test file is syntactically invalid.

Validation run:

```bash
uv run python -m py_compile test/services/test_task.py
```

Result:

```text
IndentationError: unexpected indent (test_task.py, line 80)
```

The patch leaves a duplicated fragment after the first `if __name__ == "__main__":` block, starting around `test/services/test_task.py:80`, so Python cannot import or run the test module at all.

I am closing this PR for now because it breaks the test suite at import time. A cleaned-up PR that removes the duplicated block and keeps the helper/test refactor focused would be welcome.

PR #907, opened 2026-05-28
Stabilize local TTS and video tests

CowCow7:
## Summary
- configure pydub to use the bundled imageio-ffmpeg binary when system ffmpeg is unavailable
- close VideoFileClip in the image preprocessing test so Windows can remove the generated file
- skip real SiliconFlow/Azure v2 tests when required API credentials are not configured

## Tests
- `python -m pytest -q` -> 34 passed, 3 skipped

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -4,6 +4,7 @@ import math
 import os
 import queue
 import re
+import shutil
 import threading
 import time
 from datetime import datetime
@@ -21,6 +22,20 @@ from app.config import config
 from app.utils import utils
 
 _DEFAULT_EDGE_TTS_TIMEOUT_SECONDS = 30.0
+
+
+def _configure_pydub_ffmpeg(audio_segment_cls):
+    configured_ffmpeg = os.environ.get("IMAGEIO_FFMPEG_EXE") or shutil.which("ffmpeg")
+    if not configured_ffmpeg:
+        try:
+            import imageio_ffmpeg
+
+            configured_ffmpeg = imageio_ffmpeg.get_ffmpeg_exe()
+        except Exception as exc:
+            logger.warning(f"failed to resolve bundled ffmpeg binary: {str(exc)}")
+
+    if configured_ffmpeg:
+        audio_segment_cls.converter = configured_ffmpeg
 
 
 def mktimestamp(time_unit: float) -> str:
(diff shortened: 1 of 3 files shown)


harry0703 (changes requested):
Thanks for the test-stabilization work. I ran the targeted tests and compile check:

```bash
uv run python -m unittest test.services.test_voice.TestVoiceService.test_siliconflow test.services.test_voice.TestVoiceService.test_azure_tts_v2 test.services.test_video.TestVideoService.test_preprocess_video
uv run python -m compileall app webui
git diff --check origin/main...HEAD
```

The tests pass locally with two skips, but one skip condition is checking the wrong config key.

Required change:

`test/services/test_voice.py::test_siliconflow` checks `vs.config.app.get("siliconflow_api_key")`, but the project stores SiliconFlow credentials under `[siliconflow].api_key` and the runtime reads `config.siliconflow.get("api_key", "")` in `app/services/voice.py`. With the current check, a correctly configured SiliconFlow test environment will still skip the real SiliconFlow test forever.

Please change the skip condition to check `vs.config.siliconflow.get("api_key")` instead. After that, the PR should be good to re-review.

harry0703:
Thanks for the PR. Because this is a cross-repository PR with maintainer edits disabled, I could not push the small follow-up fix directly to your branch.

I manually applied the PR changes to `main` and fixed the SiliconFlow skip condition to use the runtime config source:

```python
vs.config.siliconflow.get("api_key")
```

instead of `vs.config.app.get("siliconflow_api_key")`.

Applied on `main` in commit:

```text
ef5aba2 Stabilize local TTS and video tests
```

Validation run:

```bash
uv run python -m compileall app webui
uv run python - <<'PY'
import unittest
from app.config import config
config.siliconflow.clear()
config.azure.clear()
suite = unittest.TestSuite()
loader = unittest.defaultTestLoader
suite.addTests(loader.loadTestsFromName('test.services.test_voice.TestVoiceService.test_siliconflow'))
suite.addTests(loader.loadTestsFromName('test.services.test_voice.TestVoiceService.test_azure_tts_v2'))
suite.addTests(loader.loadTestsFromName('test.services.test_video.TestVideoService.test_preprocess_video'))
result = unittest.TextTestRunner(verbosity=2).run(suite)
raise SystemExit(0 if result.wasSuccessful() else 1)
PY
```

Result: passed, with the cloud credential tests skipped when credentials are absent.

Closing this PR because the equivalent fixed change has already landed on `main`.

PR #911, opened 2026-05-28
Optimize for OpenVINO (Intel CPU/GPU/NPU Acceleration)

DealsBeam:
Optimizes for Intel hardware by integrating OpenVINO GenAI, enabling high-performance local transcription and LLM inference.

Key Features & Changes:

- OpenVINO Whisper Integration: Added a new openvino subtitle provider. It uses the ov_genai.WhisperPipeline for significantly faster local transcription on Intel CPUs, Integrated GPUs, Discrete GPUs (Arc), and NPUs.
- Local LLM Support: Added openvino as an LLM provider. Users can now run models like Llama-3 or Qwen locally using ov_genai.LLMPipeline, avoiding cloud API costs and latency.
- Efficiency via Global Caching: Implemented global model caching for both Whisper and LLM pipelines. This ensures heavy models are loaded and compiled to the target device only once, preventing delays and OOM issues during multi-paragraph generation.
- WebUI Enhancements: Updated the "Basic Settings" panel to include OpenVINO as a provider choice and added a selector for target devices (CPU, GPU, NPU, AUTO).
- Dependency Management: Updated pyproject.toml and requirements.txt with stable OpenVINO 2025.0.0


LauraGPT:
Non-blocking downstream ASR compatibility note: this PR is a useful step because it turns subtitle generation into a selectable local transcription provider instead of keeping everything tied to Edge or faster-whisper.

A few provider-boundary details I would preserve so this remains easy to extend later to other local ASR engines such as FunASR/SenseVoice or video-clipping workflows like FunClip:

- Keep `subtitle_provider` as the routing boundary and avoid baking Whisper-only assumptions into shared UI labels, config names, or fallback paths. `openvino` can be the first hardware-accelerated local provider, but the interface should still be `audio -> SRT/timestamped segments`.
- Make the error path explicit before `subtitle.correct(...)`: if the OpenVINO model directory is missing, `openvino-genai` is not installed, or transcription returns no chunks, skip correction with a clear user-facing setup message rather than trying to correct a missing/empty SRT.
- Preserve the same output contract as the existing provider: valid UTF-8 SRT, monotonically increasing timestamps, and CJK punctuation/text untouched before the existing subtitle correction pass.
- Consider keeping language/model hints provider-neutral in config, so a later `funasr`/`sensevoice` provider can reuse the same UI shape without another settings migration.

Not asking this PR to add another ASR engine; just noting the compatibility seam that will matter for local/offline subtitle providers after OpenVINO lands.

harry0703:
Thank you for the contribution and for exploring OpenVINO acceleration. We will not merge this PR because the current implementation requires substantial correctness, compatibility, dependency, and testing work, and the local LLM portion overlaps with existing providers. We may revisit OpenVINO subtitle acceleration as a smaller optional integration if there is sufficient user demand. Thanks again!

PR #914, opened 2026-05-29
i18n: add 19 missing Turkish (tr) translation keys

ltianyi992:
## Summary

Adds 19 missing Turkish translation keys to `webui/i18n/tr.json`.

These keys exist in `en.json` but were absent from `tr.json`, causing the UI to fall back to English for Turkish users when using the API Key management features (Pexels/Pixabay key add/delete flows and the custom audio file option).

## Changes

- `webui/i18n/tr.json` — add 19 missing keys covering:
  - Custom audio file UI
  - API Key management panel (show/hide, current keys label)
  - Pexels API Key add / delete / success / error messages
  - Pixabay API Key add / delete / success / error messages


harry0703:
Merged, thank you. I verified the change locally before merging by applying it onto the current main branch, running `git diff --check`, validating `webui/i18n/tr.json` with `python3 -m json.tool`, and running the voice regression test suite together with the other accepted PRs. Everything passed; only the external-credential TTS cases were skipped.

PR #921, opened 2026-05-29
Update test_task.py

krataratha:


harry0703:
Thanks for the contribution. The idea of reducing duplicated test setup and adding stronger assertions is useful, but this PR cannot be merged in its current state.

I checked it against the current `main` branch and found several blocking issues:

- The PR is currently marked as conflicting with `main`.
- `test/services/test_task.py` does not compile:

```text
IndentationError: unexpected indent (test_task.py, line 83)
```

- The file contains leftover duplicated `VideoParams(...)` code after the `if __name__ == "__main__": unittest.main()` block, which appears to be an accidental edit artifact.
- `git diff --check` reports trailing whitespace.
- The new assertions are not aligned with the actual `tm.start()` return structure, so they do not provide reliable coverage yet.

Because of these issues, I am closing this PR. The testing direction is still valuable, so a clean follow-up PR based on the latest `main` would be welcome.

PR #929, opened 2026-05-29
fix: handle non-enum video_contact_mode in download_videos

ingamartinez:
## What

`download_videos` accessed `video_contact_mode.value` directly, which raises `AttributeError` when the argument is passed as a plain value instead of a `VideoConcatMode` enum.

## Fix

Use `getattr(video_contact_mode, "value", video_contact_mode)` so both the enum and a raw value are handled gracefully before comparing against `VideoConcatMode.random.value`.

## Impact

Prevents a crash in the video concatenation path when callers pass the mode as a string/value.

### DIFF
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -266,7 +266,8 @@ def download_videos(
     elif material_directory and not os.path.isdir(material_directory):
         material_directory = ""
 
-    if video_contact_mode.value == VideoConcatMode.random.value:
+    concat_mode_value = getattr(video_contact_mode, "value", video_contact_mode)
+    if concat_mode_value == VideoConcatMode.random.value:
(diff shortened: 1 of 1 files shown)


harry0703:
Merged, thank you for the focused fix.

I verified the failure locally before merging: calling `download_videos(..., video_contact_mode="random")` on the previous `main` raised:

```text
AttributeError: 'str' object has no attribute 'value'
```

Your change fixes that path by accepting either a `VideoConcatMode` enum value or a plain string value. After merging, I also added a small regression test so this does not regress again:

```text
uv run --with pytest pytest test/services/test_material.py -k string_concat_mode
```

Result: `1 passed`.

Merged commit: 27f1549
Regression test commit: 83f88fe

PR #931, opened 2026-05-30
fix: Docker Ollama detection, subtitle wrapping, BGM path validation, edge-tts rate, i18n & README

manh20051151:
### Changes

- **Docker detection**: auto-detect Docker env, use \host.docker.internal\ for Ollama base URL
- **Subtitle wrapping**: rewrite \wrap_text\ with PIL measurement; token-level & char-level splitting for long text
- **BGM path**: accept project-relative paths (\./resource/songs/xxx.mp3\), block paths outside whitelist
- **edge-tts**: emit \+0%\ instead of bare \

### DIFF
--- a/app/config/config.py
+++ b/app/config/config.py
@@ -5,6 +5,10 @@ import socket
 import toml
 from loguru import logger
 
+
+def in_docker() -> bool:
+    return os.path.isfile("/.dockerenv") or os.path.isfile("/proc/1/cgroup")
+
 root_dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))
 config_file = f"{root_dir}/config.toml"
 
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -99,7 +99,10 @@ def _generate_response(prompt: str) -> str:
                 model_name = config.app.get("ollama_model_name")
                 base_url = config.app.get("ollama_base_url", "")
                 if not base_url:
-                    base_url = "http://localhost:11434/v1"
+                    if config.in_docker():
+                        base_url = "http://host.docker.internal:11434/v1"
+                    else:
+                        base_url = "http://localhost:11434/v1"
             elif llm_provider == "openai":
                 api_key = config.app.get("openai_api_key")
                 model_name = config.app.get("openai_model_name")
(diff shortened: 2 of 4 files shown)


harry0703:
Thanks for the contribution. I reviewed the PR locally, but I cannot merge it in its current form.

There are two blocking issues:

1. `config.example.toml` now contains API keys and they are not quoted.

This makes the example config invalid TOML. I verified this locally on the PR branch:

```text
TOMLDecodeError Invalid value (at line 23, column 20)
```

Example config files must not contain real or personal API keys. Please restore these values to empty arrays, for example:

```toml
pexels_api_keys = []
pixabay_api_keys = []
```

2. The Docker detection is too broad.

The current implementation checks:

```python
os.path.isfile("/.dockerenv") or os.path.isfile("/proc/1/cgroup")
```

`/proc/1/cgroup` exists on many normal Linux systems, so this can incorrectly treat non-Docker Linux environments as Docker and change the default Ollama URL from `http://localhost:11434/v1` to `http://host.docker.internal:11434/v1`.

Please narrow the detection, for example by checking `/.dockerenv`, or by reading `/proc/1/cgroup` and only treating it as Docker/containerized when it contains clear container markers such as `docker`, `containerd`, or `kubepods`.

Also, the PR title mentions subtitle wrapping, BGM path validation, edge-tts rate, i18n, and README changes, but the actual diff is only about Ollama Docker detection and config changes. Several of those fixes already exist in `main`. Please rebase onto the latest `main` and keep this PR focused on the Ollama Docker default URL behavior.

harry0703:
Thanks again for the idea. I implemented the useful part of this PR directly in `main` with a safer and more complete approach, so I am closing this PR instead of merging it.

What changed in `main`:

- Ollama now keeps `http://localhost:11434/v1` as the default outside containers.
- When running inside a container, the default becomes `http://host.docker.internal:11434/v1` when that hostname is resolvable.
- For native Linux Docker environments where `host.docker.internal` may not exist, the code falls back to the container default gateway IP.
- A user-provided `ollama_base_url` always takes precedence.
- Added unit tests for non-container, Docker/Desktop-style, native Linux Docker fallback, and explicit user configuration cases.

I also verified this with a real local Docker container and a host Ollama instance using `qwen3.5:0.8b`; both the host path and container path returned a valid response.

Commit: 536efa3

I did not merge this PR because it also changed `config.example.toml` to include API keys and invalid TOML syntax, which would break new user configuration loading.

PR #943, opened 2026-05-31
perf(i18n): cache locale loading to avoid disk reads on every rerun

almhajer:
Streamlit re-runs the whole script on each interaction, so `load_locales()` re-parsed every locale JSON from disk each time. Wrap it in `functools.lru_cache` (locale files are static at runtime).

`app/utils/utils.py`. Identical return value; only repeated disk I/O is removed. No API or behaviour change.

### DIFF
--- a/app/utils/utils.py
+++ b/app/utils/utils.py
@@ -1,6 +1,7 @@
 import json
 import locale
 import os
+from functools import lru_cache
 from pathlib import Path
 import threading
 from typing import Any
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for the focused improvement.

I have absorbed this change into `main` in commit `8b2e530`. I kept the same behavior and added the locale cache with a project-style Simplified Chinese comment. I also included it in the local validation pass before pushing.

Validation run:
- `uv run --with pytest pytest test/services/test_task.py test/services/test_video.py -q` -> 15 passed
- `uv run --with ruff ruff check app/utils/utils.py app/models/const.py app/services/voice.py app/controllers/v1/video.py test/services/test_voice.py` -> passed
- `uv run python -m compileall -q app test/services/test_voice.py test/services/test_task.py test/services/test_video.py` -> passed

Closing this PR because the change is now on `main`.

almhajer:
thnaks so match @harry0703

PR #944, opened 2026-05-31
fix(subtitle): tolerant Arabic letter-form matching (avoid Whisper fallback)

almhajer:
For Arabic scripts, `edge-tts` returns cue text whose letter forms differ from the script (أ/إ/آ→ا, ى→ي, ة→ه) plus diacritics/tatweel, so the strict per-line match failed for every line, `sub_items` came out 0, and generation silently fell back to Whisper.

Files: `app/services/voice.py`, `app/models/const.py`. Adds `_normalize_arabic()` and a final tolerant tier in `_match_script_line()` (no-op for non-Arabic), plus Arabic punctuation (،؛؟) as split points. Verified with 12 unit cases.


harry0703:
Thank you for splitting this out. The Arabic subtitle matching issue is valid and useful.

I have absorbed the idea into `main` in commit `8b2e530`, with a few adjustments for this project:

- kept the implementation comments in Simplified Chinese, matching the repository's current collaboration rules
- preserved the original Arabic script text for final subtitle output
- used Arabic normalization only as the final fallback matching tier
- added regression tests for Arabic punctuation splitting, Arabic letter-form normalization, and Edge cue aggregation

Validation run:
- `uv run --with pytest pytest test/services/test_task.py test/services/test_video.py -q` -> 15 passed
- `uv run --with ruff ruff check app/utils/utils.py app/models/const.py app/services/voice.py app/controllers/v1/video.py test/services/test_voice.py` -> passed
- `uv run python -m compileall -q app test/services/test_voice.py test/services/test_task.py test/services/test_video.py` -> passed

One full `test_voice.py` run was blocked by an existing external SiliconFlow account issue: the configured API key currently returns `403: please complete identity verification`. The non-external regression suite passed.

Closing this PR because the fix is now on `main`.

almhajer:
thnaks so match @harry0703

PR #945, opened 2026-05-31
feat(subtitle): optional rounded semi-transparent subtitle background

almhajer:
Render the subtitle background as a semi-transparent (~55%) rounded box hugging the text, instead of a solid opaque rectangle. Text is drawn transparent and composited over a PIL rounded-rectangle RGBA background.

`app/services/video.py`. When the background is disabled, the original rendering path is unchanged. Verified by rendering a frame (transparent rounded corners + correct translucency).

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -20,7 +20,8 @@ from moviepy import (
     afx,
 )
 from moviepy.video.tools.subtitles import SubtitlesClip
-from PIL import Image, ImageFont
+import numpy as np
+from PIL import Image, ImageDraw, ImageFont
 
 from app.models import const
 from app.models.schema import (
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for the subtitle rendering improvement idea.

I am not merging this PR for now because it changes the default subtitle background rendering behavior from the current solid rectangle to a semi-transparent rounded background without a user-facing configuration switch. For a project with a large user base, that kind of visual behavior change should be opt-in and covered by rendering tests, otherwise existing users may get unexpected output changes.

The direction is still valuable. A safer version would be:

- keep the current default rendering unchanged
- add an explicit WebUI/API option for rounded or translucent subtitle background
- make opacity/radius configurable or at least tied to a documented default
- add a focused rendering test or screenshot/frame verification

Closing this PR for now. We can revisit this feature with a compatibility-preserving implementation.

harry0703:
I revisited this feature and absorbed the idea into `main` with a compatibility-preserving implementation.

Implemented in commit `f85a0ff`:

- added a WebUI option: `Enable rounded translucent subtitle background`
- kept the option disabled by default, so existing subtitle rendering behavior is unchanged
- when enabled, subtitle backgrounds use a tighter rounded translucent box that fits the text width
- added API/schema support via `rounded_subtitle_background`
- added a regression test for transparent rounded corners and translucent center rendering

Validation run:

- `uv run --with pytest pytest test/services/test_video.py test/services/test_task.py -q` -> 16 passed
- `uv run --with ruff ruff check app/models/schema.py app/services/video.py test/services/test_video.py` -> passed
- `uv run --with ruff ruff check --ignore E402 webui/Main.py` -> passed
- `uv run python -m json.tool webui/i18n/zh.json` / `webui/i18n/en.json` -> passed

I also rendered a local preview video/frame to verify the visual result before pushing.

Thanks again for the suggestion.

almhajer:
Reworked into a compatibility-preserving version (default unchanged, opt-in WebUI/API options + tests) here: https://github.com/harry0703/MoneyPrinterTurbo/pull/948 — couldn't reopen this one since the branch was force-pushed. Thanks for the guidance!

PR #948, opened 2026-05-31
feat(subtitle): optional rounded/translucent subtitle background (opt-in, configurable) — supersedes #945

almhajer:
This reworks #945 to be fully compatibility-preserving, per the review feedback there. Thanks again for the clear direction.

**Default behavior is unchanged** — subtitles still render a solid, fully opaque rectangle. The rounded/translucent background is strictly opt-in and configurable.

### What changed
- **API** (`app/models/schema.py`): `subtitle_background_rounded` (default `False`), `subtitle_background_opacity` (default `100`), `subtitle_background_radius` (default `20`) on `VideoParams` + `SubtitleRequest`. Defaults reproduce the original output exactly.
- **Rendering** (`app/services/video.py`): the default path keeps the original solid opaque `TextClip`; only when opted in is the text composited over a PIL rounded-rectangle RGBA background with configurable alpha/radius. Pure helpers `hex_to_rgb()` and `build_subtitle_background_rgba()` extracted for testing.
- **WebUI** (`webui/Main.py`): checkbox + opacity slider + corner-radius slider in the subtitle settings.
- **i18n**: 4 new keys across the existing locales.
- **Tests** (`test/services/test_subtitle_background.py`): verify defaults stay opaque and that the opt-in path yields transparent rounded corners + configurable translucency. (6 tests, all passing.)

Maps to the checklist from #945:
- [x] keep the current default rendering unchanged
- [x] explicit WebUI/API option for rounded / translucent background
- [x] configurable opacity / radius with documented defaults
- [x] focused rendering test (alpha/corner frame checks)

Closes #945 conceptually (that PR can't be reopened after a force-push).

### DIFF
--- a/app/models/schema.py
+++ b/app/models/schema.py
@@ -100,6 +100,11 @@ class VideoParams(BaseModel):
     font_name: Optional[str] = "STHeitiMedium.ttc"
     text_fore_color: Optional[str] = "#FFFFFF"
     text_background_color: Union[bool, str] = True
+    # Opt-in subtitle background styling. Defaults preserve the original solid,
+    # fully opaque rectangle so existing output is unchanged.
+    subtitle_background_rounded: Optional[bool] = False
+    subtitle_background_opacity: Optional[int] = 100  # 0-100 (%); 100 = opaque
+    subtitle_background_radius: Optional[int] = 20  # px corner radius when rounded
 
     font_size: int = 60
     stroke_color: Optional[str] = "#000000"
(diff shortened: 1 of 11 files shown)


almhajer:
thnaks so match

harry0703:
Thank you for reworking this into a compatibility-preserving version.

The direction is valid, and I appreciate that this version keeps the default output unchanged and adds explicit opt-in controls. However, we have already absorbed the core feature into `main` in commit `f85a0ff` with a smaller implementation:

- default subtitle rendering remains unchanged
- WebUI now has an opt-in rounded translucent subtitle background switch
- API/schema support was added through `rounded_subtitle_background`
- rendering behavior was verified with a local preview frame/video
- regression coverage was added for transparent rounded corners and translucent center rendering

This PR now conflicts with `main`, and it introduces a second set of parameter names plus additional opacity/radius controls. For now, I prefer to keep the first version simple and low-risk. If users need configurable opacity or corner radius later, we can add those as separate advanced options after the simpler switch has had some real usage.

Closing this PR because the main capability is already available on `main`. Thanks again for the careful follow-up.

almhajer:
heee it ok importnat add thanks @harry0703

PR #961, opened 2026-06-01
i18n: complete missing German (de) translations

viraatdas:
`webui/i18n/de.json` was missing about 15 keys that exist in `en.json`, so German users saw raw English labels for those controls (the `tr()` lookup falls back to the key when a translation is missing).

This adds natural German translations for the missing keys: advanced script settings, custom script requirements, custom system prompt, custom audio file, rounded subtitle background, and the MiMo TTS settings. The German `Translation` key set now matches `en.json`. Only `de.json` is touched.

Prepared with AI assistance.


harry0703:
Thank you for the contribution. I reviewed the PR and the direction is correct.

I have absorbed the German translation updates into `main` and also added the latest missing keys for the new advanced video encoder settings, so `webui/i18n/de.json` now matches the current `en.json` key set.

Validation performed:

- `python -m json.tool webui/i18n/de.json`
- Compared `de.json` and `en.json` translation keys; no missing or extra keys remain.
- `python -m py_compile webui/Main.py`

Implemented in: f034560

I will close this PR since the changes have been applied directly on `main`.

PR #964, opened 2026-06-01
feat(minimax): upgrade default model to MiniMax-M3

octo-patch:
## Summary

Bump the minimax default model from `MiniMax-M2.7` to the newly released `MiniMax-M3` in `config.example.toml`. M2.7 and M2.7-highspeed remain selectable so existing users can pin the older model if they want to.

## Changes

- `config.example.toml`: change `minimax_model_name` default to `MiniMax-M3`, update the Available models comment.

No code changes in `app/services/llm.py` — the minimax branch already reads `minimax_model_name` from config and talks to the OpenAI-compatible endpoint at `https://api.minimax.io/v1`, so the new model works out of the box.

## Test plan

- [x] TOML still parses
- [x] No tests reference the model name
- [ ] Local smoke test: set `minimax_api_key` + run a video task with the new default

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -134,10 +134,10 @@ qwen_model_name = "qwen-max"
 ########## MiniMax API Key
 # Visit https://platform.minimax.io to get your API key
 # MiniMax API is OpenAI-compatible
-# Available models: MiniMax-M2.7 (default), MiniMax-M2.7-highspeed
+# Available models: MiniMax-M3 (default), MiniMax-M2.7, MiniMax-M2.7-highspeed
 minimax_api_key = ""
 minimax_base_url = "https://api.minimax.io/v1"
-minimax_model_name = "MiniMax-M2.7"
+minimax_model_name = "MiniMax-M3"
 
 ########## Xiaomi MiMo API Key
 # Xiaomi MiMo API is OpenAI-compatible.


harry0703:
Thank you for the PR. The MiniMax M3 update is useful, but I did not merge this PR as-is because a real smoke test showed two additional details we should handle together:

- The provided MiniMax API key works with `https://api.minimaxi.com/v1`, while `https://api.minimax.io/v1` returns `401 invalid api key` for that key.
- `MiniMax-M3` may return reasoning content wrapped in `<think>...</think>`, which should not be included in generated scripts, subtitles, or voiceover text.

I have applied the change directly on `main` with a slightly broader fix:

- Updated the example MiniMax default model to `MiniMax-M3`.
- Documented the China endpoint fallback: `https://api.minimaxi.com/v1`.
- Added LLM response normalization to remove closed and unclosed `<think>` blocks.
- Added tests for the reasoning-block cleanup.

Validation performed:

- `uv run --with pytest pytest test/services/test_llm.py -q` -> `28 passed, 1 skipped`
- `python -m py_compile app/services/llm.py test/services/test_llm.py`
- `git diff --check`
- Real MiniMax M3 smoke test with the OpenAI-compatible client on `https://api.minimaxi.com/v1`; the final normalized result no longer contains `<think>` content.

Implemented in: dd36f6f

I will close this PR since the change has been absorbed into `main`.

PR #965, opened 2026-06-01
add awsbedrock as a llm model provider

minggeorgelei:
add awsbedrock as a llm model provider


harry0703:
Thank you for taking the time to add AWS Bedrock support.

We have decided not to merge this implementation because the current main branch already supports Bedrock through LiteLLM. Keeping Bedrock behind LiteLLM avoids maintaining a second provider implementation and lets AWS users rely on standard AWS credential sources, including environment variables, profiles, and IAM roles, instead of storing long-lived access keys in the application configuration. It also avoids making boto3 a required dependency for users who do not use AWS.

We may improve the Bedrock setup guidance or add a clearer WebUI entry in the future while continuing to use the LiteLLM backend.

Thanks again for the contribution and for helping us evaluate this integration.

minggeorgelei:
Thanks for you reply.

PR #970, opened 2026-06-02
Complete Turkish UI translations

erayusta:
Completes missing Turkish UI translation keys so the Turkish locale matches the English source translations.

Validation:
- Verified JSON syntax for en.json and tr.json
- Confirmed Turkish locale has the same translation keys as English


harry0703:
Thanks for completing the Turkish UI translations.

This PR was based on an older version of `webui/i18n/tr.json`, so it became conflicted after recent UI translation updates on `main`. I absorbed the missing Turkish translation keys directly into `main` instead of merging the PR as-is.

Validation completed:
- `tr.json` JSON syntax check passed
- all `webui/i18n/*.json` files parse correctly
- English and Turkish translation keys now match: 137 / 137, with no missing or extra keys
- `uv run --with pytest pytest -q test/services/test_subtitle_background_settings.py` passed

Implemented in commit `3ea1043`.

PR #975, opened 2026-06-02
feat: add pluggable render backend

a-essawy:
Hello,

Thanks for this cool project, it's been really helpful.

One thing I noticed is that the render step and caption burning can get heavy, especially for folks on older PCs or slower internet connections. So this PR adds an optional the option to run the render on a cloud backend instead of locally.

image

It adds a dropdown to choose the render method, Local by default, and the cloud backend as an option. The two are interchangeable, and the output of either should look almost the same. If the cloud render ever fails it just falls back to Local, so it can't get in the way of a run.

Happy to discuss anything and keep contributing to the project. (For transparency, [Rendobar](https://rendobar.com/ffmpeg/) is one of the projects I work on, so I'm happy to keep the integration generic for other backends or adjust whatever you'd prefer.)

I also think the cloud render idea could open the door to some fun directions later on, like a mobile app or an automation pipeline.

Thanks!!

### DIFF
--- a/app/models/schema.py
+++ b/app/models/schema.py
@@ -79,6 +79,7 @@ class VideoParams(BaseModel):
     video_clip_duration: Optional[int] = 5
     video_count: Optional[int] = 1
 
+    render_backend: Optional[str] = "local"
     video_source: Optional[str] = "pexels"
     video_materials: Optional[List[MaterialInfo]] = (
         None  # Materials used to generate the video
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -225,6 +225,23 @@ def save_video(video_url: str, save_dir: str = "") -> str:
     return ""
 
 
+class SourcedPath(str):
+    """A clip path that also remembers the URL it came from.
+
+    It IS a str, so it behaves like the path everywhere downstream. An optional
+    render backend can read `.source_url` to fetch the clip from its origin
+    instead of re-uploading it.
+    """
+
+    source_url: str = ""
+
+
+def _sourced(path: str, url: str) -> "SourcedPath":
+    sp = SourcedPath(path)
+    sp.source_url = url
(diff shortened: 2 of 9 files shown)


harry0703:
Thanks for the thoughtful PR and for being transparent about the Rendobar connection.

The idea is valuable: a pluggable render backend could help users on older machines avoid heavy local MoviePy/FFmpeg work, and it may also open up future workflows such as hosted automation or mobile clients.

I reviewed the implementation and ran the basic validation on the PR branch:
- `uv run python -m compileall app webui` passed
- `uv run --with pytest pytest -q test/services` passed: 79 passed, 3 skipped

However, I do not think this is safe to merge into `main` in its current form. The project now has a fairly large user base, and this PR changes the core video generation path while introducing a third-party cloud rendering service. Before we can accept something like this, we would need a smaller and more generic backend abstraction, explicit user consent and UI messaging for uploading local audio/subtitle/font/video files to a third party, clearer cost/credit behavior, stronger escaping and validation around remote FFmpeg command construction, request-scoped backend selection instead of relying on global config state, and dedicated tests for local fallback and cloud failure paths.

The branch is also based on an older `main` and currently conflicts with recent subtitle/background UI changes, so merging it directly would be risky.

For now I am going to close this PR rather than merge it. The concept is still interesting, but it needs to be redesigned in a more conservative way before it can be added to the main project.

Thanks again for contributing and for raising this direction.

a-essawy:
Thank you @harry0703 for taking the time to review it, and I totally get the decision.

The points you raised are fair too, If you ever feel like it could be a fit, I'd be happy to take another go at it in a smaller conservative way, No worries at all if it's not something you want though.

Either way thanks again for the feedback and for the work you put into this project.

Cheers!!

Issue #511, opened 2024-10-27
linux系统下的在打开“启用字幕”后会导致IMv7的错误，#418问题的后续


Lambda-zhou:
描述同#418
报错信息：
`OSError: MoviePy Error: creation of None failed because of the following error:
WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"
convert: unable to read font Georgia-Bold' @ warning/annotate.c/RenderType/1024. convert: delegate library support not built-in '/usr/share/fonts/type1/gsfonts/n019003l.pfb' (Freetype) @ warning/annotate.c/RenderFreetype/2112. convert: no decode delegate for this image format ' @ error/constitute.c/ReadImage/746.
convert: no images defined `PNG32:/tmp/tmp0xuf3v7x.png' @ error/deprecate.c/ConvertImageCommand/3368.
.This error can be due to the fact that ImageMagick is not installed on your computer, or (for Windows users) that you didn't specify the path to the ImageMagick binary. Check the documentation.`

尝试的办法：
1.修改了ImageMagick-7 和 ImageMagick-6下的文件policy.xml还是报错，在config.toml配置文件改了imagemagick_path = "/etc/ImageMagick-6",或是ImageMagick-7都一样
2.起一个新的环境，不用编译的，直接apt安装 apt install imagemagick libfreetype6-dev 然后 convert -list configure 确保DELEGATES这一栏有fontconfig freetype这两个

希望有彻底的解决办法（另外在实验过后发现似乎可以直接在app的页面中设定好不用修改config.toml配置文件？）

harry0703:
I checked this against the current main branch.

This report is tied to the older ImageMagick-based subtitle path. The current project now uses moviepy 2.1.2, and TextClip is no longer going through the old convert/magick workflow that produced the IMv7 warning in the original report.

What I verified locally on current main:
- TextClip renders successfully with a bundled font file from resource/fonts
- the failure mode is no longer the old ImageMagick convert error from this issue
- the original report also referenced Georgia-Bold, which is not a reliable default font in many Linux containers; using an explicit font file from the repo is the correct direction on modern main

So I am closing this as outdated or fixed by the current stack change. If someone can still reproduce a subtitle rendering failure on the latest main branch, please open a new issue with:
- the exact current commit or release
- the full traceback
- the configured font_name
- the container base image or Linux distribution
[...]

PR #977, opened 2026-06-02
docs(README-en): remove obsolete ImageMagick requirement and fix doc quality issues

jonathanzhan1975:
## Summary

Since the project upgraded to **MoviePy 2.x**, `TextClip` renders subtitles via **Pillow** rather than ImageMagick. ImageMagick is no longer a dependency, but `README-en.md` still instructed users to install it as a mandatory step and contained two FAQ entries that sent users down a broken troubleshooting path. This was confirmed obsolete in [issue #511](https://github.com/harry0703/MoneyPrinterTurbo/issues/511).

This PR also addresses several documentation quality issues identified by static analysis.

### ImageMagick (no longer required)
- **Remove** the "Install ImageMagick" step from Manual Deployment and renumber remaining steps (③→②, ④→③)
- **Replace** both ImageMagick FAQ entries (`ImageMagick is not installed` and `security policy prevents operations`) with a single clear notice: not required since MoviePy 2.x — run `git pull` / `update.bat` to get current code

### API access URL fix
- `http://0.0.0.0:8080/docs` → `http://127.0.0.1:8080/docs` — `0.0.0.0` is a bind address, not a valid URL for a browser on most systems

### Orphaned paragraph removed
- Removed an API URL paragraph that was stranded inside the Sponsors section (copy-paste residue)

### Voice Synthesis section expanded
- Added a config snippet showing how to configure Edge TTS vs Azure TTS voices, replacing the single stale v1.1.2 changelog line that provided no actionable guidance

### Subtitle comparison made concrete
- Replaced vague terms ("faster", "slower", "better performance", "more reliable quality") with specific information: timing estimates for whisper on CPU, model size comparison (250 MB turbo vs 3 GB large-v3), and a clear explanation of what each mode actually does

### Whisper error text cleaned up
- Wrapped raw error messages in fenced code blocks
- Fixed upstream typos preserved verbatim in the original: `LocalEntryNotfoundEror` → `LocalEntryNotFoundError`, `trafic` → `traffic`, `snapshotfolderfor` → `snapshot folder for`

## Test plan

- [ ] Verify subtitle rendering works without ImageMagick installed (MoviePy 2.x + Pillow only)
- [ ] Confirm `http://127.0.0.1:8080/docs` correctly reaches the API docs after `uv run python main.py`
- [ ] Review rendered Markdown for formatting correctness

### DIFF
--- a/README-en.md
+++ b/README-en.md
@@ -264,9 +264,11 @@ Picwish focuses on the **image processing field**, providing a rich set of **ima
 
 A list of all supported voices can be viewed here: [Voice List](./docs/voice-list.txt)
 
-The default TTS provider is **Edge TTS** (free, no API key required). To switch voices, set `voice_name` in `config.toml` or select one from the WebUI voice dropdown.
+The default TTS provider is **Edge TTS** (free, no API key required). In the WebUI it appears as **"Azure TTS V1"** — this is the same thing. To switch voices, set `voice_name` in `config.toml` or select one from the WebUI voice dropdown.
 
-To use higher-quality **Azure TTS** voices, configure your Azure Speech credentials in `config.toml`:
+> **Note:** "Azure TTS V1" (Edge TTS, free) and "Azure TTS V2" (paid Azure Speech SDK) are two different options in the WebUI. Only V2 requires an Azure API key.
+
+To use higher-quality **Azure TTS V2** voices, configure your Azure Speech credentials in `config.toml`:
 
 ```toml
 [azure]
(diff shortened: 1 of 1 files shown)


harry0703:
Thanks for the careful documentation cleanup.

I merged this PR into `main`. The ImageMagick requirement was indeed obsolete after the project moved to MoviePy 2.x/Pillow-based subtitle rendering, and the `127.0.0.1:8080` API docs URL is clearer for users than `0.0.0.0:8080`.

I also added a follow-up commit to sync the same documentation cleanup into the Chinese README so the bilingual docs stay consistent:

- PR merge commit: `797b5bc`
- Chinese README sync commit: `788f52a`

Thanks again for improving the docs.

PR #981, opened 2026-06-03
feat: implement silent voice mode and refactor ffmpeg logic

yanglinzhen1022:
1. Added 'None' voice option to skip synthesis and generate silent audio based on text length.
2. Refactored ffmpeg binary resolution: moved to utils.py and updated callers for consistency.
3. Enabled backend hot-reload by default (reload_debug=True) for better developer experience.
4. Added MiniMax to the LLM provider list in the WebUI.
5. Updated i18n translations (zh/en) to support the new silent voice mode.

### DIFF
--- a/app/config/config.py
+++ b/app/config/config.py
@@ -181,7 +181,7 @@ project_description = _cfg.get(
     "<br><small>Supported by <a href='https://aihubmix.com/?aff=CEve'>AIHubMix</a></small>",
 )
 project_version = _cfg.get("project_version", "1.2.9")
-reload_debug = False
+reload_debug = True
 
 app["redis_host"] = os.getenv(
     "MPT_APP_REDIS_HOST",
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -123,29 +123,6 @@ def _prioritize_unique_source_clips(
     return primary_items + overflow_items
 
 
-def get_ffmpeg_binary():
-    # 优先复用用户在 config.toml / 环境变量里显式指定的 ffmpeg，可避免
-    # Windows 便携包、Docker、自定义安装目录等场景下 PATH 不一致。
-    configured_ffmpeg = os.environ.get("IMAGEIO_FFMPEG_EXE")
-    if configured_ffmpeg:
-        return configured_ffmpeg
-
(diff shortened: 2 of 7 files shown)


harry0703:
Thank you for the contribution. The ideas in this PR are useful, especially the no-voice mode, shared FFmpeg resolution, and adding MiniMax to the WebUI provider list.

I have selectively absorbed the low-risk parts into `main` in commit `76d7e09`:

- Added an explicit `No Voice` / `no-voice` TTS mode in the WebUI.
- Implemented silent audio generation as a timeline placeholder, so the existing video, subtitle, and composition pipeline keeps working without narration.
- Kept empty `voice_name` as a configuration error instead of silently treating it as no-voice mode.
- Added multilingual duration estimation for no-voice scripts, including non-ASCII languages such as Russian and Arabic.
- Moved FFmpeg binary resolution into a shared utility while keeping a compatibility wrapper in `video.py`.
- Added MiniMax to the WebUI LLM provider list.
- Added i18n entries for all existing WebUI locale files.
- Added regression tests for no-voice mode, FFmpeg resolution, i18n keys, and related edge cases.

I did not merge this PR directly because it also changes `reload_debug` to `True` by default, which is not suitable for normal production usage, and the original implementation mixed several unrelated changes in one PR. Keeping the final change smaller makes it easier to review and safer for existing users.

Validation completed:

- Python compile check for the changed service and WebUI files.
- Focused unit tests for no-voice mode, FFmpeg resolution, i18n keys, and video codec fallback.
- Real local smoke test generating a video with no narration, subtitles, and a silent audio track.
- Real SiliconFlow TTS smoke test with the current API key.

Thanks again for pointing the project in this direction.

harry0703:
Closing this PR now.

We have selectively absorbed the useful parts into `main` in commit `76d7e09`, including the explicit No Voice mode, shared FFmpeg resolution, MiniMax in the WebUI provider list, full i18n coverage, and regression tests.

I did not merge this PR directly because it also changed `reload_debug` to `True` by default and mixed several unrelated changes in one PR. The final implementation keeps the production default unchanged and narrows the behavior so normal TTS settings are not affected.

PR #982, opened 2026-06-03
Update config / __init__.py

hmd37:
I just replace os with object oriented pathlib module to increase readablity

### DIFF
--- a/app/config/__init__.py
+++ b/app/config/__init__.py
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
 import sys
 
 from loguru import logger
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for the contribution.

I understand the intent: replacing `os.path` with `pathlib` can improve readability in many places. However, I am going to close this PR without merging it because this specific change is mostly a style-only refactor and it slightly changes the logger path-formatting behavior.

The current code uses:

```python
os.path.relpath(file_path, root_dir)
```

That always produces a relative path from the project root, even when the logged file path is outside the project directory.

The PR changes it to:

```python
file_path.relative_to(root_dir)
```

When the file is outside the project root, `relative_to()` raises `ValueError`, and the fallback keeps the absolute path. Because the formatter later prefixes it with `./`, logs could become less stable or less clean for external/library paths.

Since this does not fix a bug or add a feature, and the project currently prioritizes avoiding unnecessary behavior changes, I do not want to merge this refactor as-is.

Thanks again for taking the time to improve the codebase.

hmd37:
Thank you for clarification, I understand.
Actually I am new to open source contributions that's why I tried this, adding new feature is hard for me, I need to understand the code fully first, maybe I can try fixing bugs someday.

PR #987, opened 2026-06-03
fix: keep last subtitle when SRT has no trailing blank line

Bortlesboat:
## Problem

`app/services/subtitle.py::file_to_subtitles` only records a subtitle block when it encounters a **following blank line**:

```python
elif line.strip() == "" and current_times:
    index += 1
    times_texts.append((index, current_times.strip(), current_text.strip()))
    current_times, current_text = None, ""
```

If the SRT file's **last block is not terminated by a trailing blank line** (very common — many tools and hand-written files omit it), that branch never fires for the final block, so the last subtitle is **silently dropped**.

```python
# SRT ending in "...World" with no trailing newline:
file_to_subtitles(path)   # -> only ["Hello"], "World" is lost
```

This feeds `correct()`, `task.py` subtitle handling, so a dropped last line propagates into the rendered video / corrected SRT.

## Fix

Flush the pending block after the read loop:

```python
if current_times:
    index += 1
    times_texts.append((index, current_times.strip(), current_text.strip()))
```

Files that *do* end with a blank line are unaffected (the block is already flushed inside the loop, and `current_times` is reset to `None`).

## Tests

Added two cases to `test/services/test_subtitle.py`:
- `..._keeps_last_block_without_trailing_newline` — **fails on `main`** (1 subtitle instead of 2) and passes with the fix.
- `..._parses_blocks_with_trailing_newline` — guards against regressing the normal case.

All subtitle tests pass.

### DIFF
--- a/app/services/subtitle.py
+++ b/app/services/subtitle.py
@@ -161,6 +161,13 @@ def file_to_subtitles(filename):
                 current_times, current_text = None, ""
             elif current_times:
                 current_text += line
+
+    # Flush the final block. SRT files whose last subtitle is not followed by a
+    # trailing blank line never hit the blank-line branch above, so without this
+    # the last subtitle would be silently dropped.
+    if current_times:
+        index += 1
+        times_texts.append((index, current_times.strip(), current_text.strip()))
     return times_texts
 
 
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for the clear report and focused fix.

I verified the behavior and tested the subtitle parser with both cases:

- SRT files whose final block does not have a trailing blank line
- Normal SRT files that do end with a trailing blank line

The focused subtitle tests passed:

```bash
uv run --with pytest pytest test/services/test_subtitle.py -q
# 3 passed
```

This is a valid low-risk fix, and it has been merged into `main`.

PR #991, opened 2026-06-04
 video generator

hessinhoss20-oss:
This workflow automates the daily generation of videos using a Python script, with dependencies installed from requirements.txt and environment variables for API keys.

### DIFF
--- a/.github/auto-generate-yml
+++ b/.github/auto-generate-yml
@@ -1,0 +1,31 @@
+name: Daily Video Generator
+
+on:
+  schedule:
+    # 
+    - cron: '0 7 * * *'
+  workflow_dispatch: # هذا يتيح لك تشغيل الأتمتة يدوياً في أي وقت للتجربة
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for the suggestion.

I cannot merge this workflow as-is because it would not run correctly in the current project.

Key issues:

- The file is added as `.github/auto-generate-yml`, but GitHub Actions only loads workflow files from `.github/workflows/*.yml`.
- `actions/setup-python` uses `python-framework`, which is not a valid input; it should use `python-version`.
- The command `python main.py --prompt ...` is not supported. In this project, `main.py` starts the API server and does not provide a CLI video-generation interface.
- A scheduled workflow that generates videos every day could consume user API credits unexpectedly, so this kind of automation needs explicit design, configuration, and artifact handling before it can be safely included.

The idea of automation is useful, but it should be built on top of a proper CLI or API-based job runner first. I will close this PR for now.

Faizal89079:
You can cite projects that explicitly describe this setup:

### Source 1: AI Video Generator (GitHub)

This project uses:

* Python scripts for video generation
* Dependencies installed via `requirements.txt`
* Environment variables such as `OPENAI_API_KEY`
* Automated video generation workflow

### Source 2: Gemini YouTube Automation

This project describes:

* A daily automated workflow
* Python-based video creation pipeline
* Installation using `pip install -r requirements.txt`
* Configuration of API credentials

Source: ([[SourcePulse](https://www.sourcepulse.org/)])

PR #994, opened 2026-06-05
fix: handle empty Qwen choices

### LINKED ISSUE #984: bug: Qwen chat parser may crash with IndexError on empty choices list (commit ae52fec)
## Bug Report

**Commit:** `ae52fec` — *Fix Qwen chat response parsing*

---

### What was fixed
The PR correctly fixes `AttributeError: 'NoneType' object has no attribute 'replace'` when `output.text` is `None` by adding a fallback to `output.choices[0].message.content`.

### Remaining risk

The new fallback path assumes `choices` is a non-empty list:

```python
if choices:
    first_choice = choices[0]
    message = _get_response_field(first_choice, "message")
```

If the Qwen API returns `{"output": {"choices": []}}` (empty list), `choices` evaluates to falsy — `if choices:` is skipped — control falls through to `output.text` which may also be `None`, ultimately returning a silent error string instead of raising a clear exception.

### Steps to reproduce

```python
# Simulate Qwen returning empty choices
mock_response = {"output": {"choices": [], "text": None}}
result = _get_qwen_response(mock_response)
# Expected: clear error / exception
# Actual: silently returns "Error: returned empty text content"
```

### Suggested fix

```python
if choices and len(choices) > 0:
    first_choice = choices[0]
    ...
else:
    logger.warning("Qwen returned empty choices list")
    raise ValueError("Qwen API returned no choices")
```

Sushanth012:
## Summary
- detect explicit empty Qwen choices lists before falling back to output.text
- return a clear Qwen empty choices error instead of an unclear empty text path
- add a regression test for empty choices responses

Fixes #984

## Testing
- uv run --with pytest python -m pytest test/services/test_llm.py

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -101,7 +101,11 @@ def _extract_qwen_generation_text(response) -> str:
     """
     output = _get_response_field(response, "output")
     choices = _get_response_field(output, "choices") if output else None
-    if choices:
+    if choices is not None:
+        if not choices:
+            logger.warning("Qwen returned an empty choices list")
+            raise ValueError("[qwen] returned empty choices")
+
         first_choice = choices[0]
         message = _get_response_field(first_choice, "message")
         content = _get_response_field(message, "content") if message else None
--- a/test/services/test_llm.py
+++ b/test/services/test_llm.py
@@ -277,6 +277,18 @@ class TestLiteLLMProvider(unittest.TestCase):
         self.assertIn("returned empty text content", result)
         self.assertNotIn("NoneType", result)
 
+    def test_qwen_provider_reports_empty_choices(self):
+        """Qwen chat 响应 choices 为空时应返回明确错误。"""
+        self._use_qwen_provider()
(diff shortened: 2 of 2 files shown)


harry0703:
Thank you for the focused contribution.

I reviewed the patch and verified it locally with:

```bash
uv run --with pytest pytest test/services/test_llm.py -k qwen -q
# 4 passed

uv run --with pytest pytest test/services/test_llm.py -q
# 45 passed, 1 skipped
```

The change is small and targeted: Qwen responses with an explicit empty `choices` list now produce a clear `[qwen] returned empty choices` error, and the regression test covers that path.

Merged into `main`.

kapil971390:
Thanks for reviewing and addressing this so quickly. Glad the finding was useful and resulted in a fix. Really appreciate the discussion and contribution from everyone involved. 🙌

This issue was originally surfaced during analysis with VeriQA, so it's encouraging to see the feedback loop work in a real open-source project.

PR #999, opened 2026-06-06
Concurrent video download + real-time WebUI log display

Ileetking:
app/services/material\.py: Concurrent download, retry, and timeout quantization
 webui/Main. py: log queue + session_ state, background task + main thread polling to update logs in real time


Ileetking:
The conflict has been resolved.

PR #1003, opened 2026-06-08
fix: add video duration safety buffer

### LINKED ISSUE #985: bug: >= comparison in video clip selection may produce audio trailing silence (commit 1e56568)
## Bug Report

**Commit:** `1e56568` — *Reduce repeated video material usage*

---

### What changed
The deduplication PR introduced `_prioritize_unique_source_clips()` and changed the clip accumulation condition from `>` to `>=`:

```python
# Before
if video_duration > audio_duration:
    break

# After
if video_duration >= audio_duration:
    break
```

### The risk

The old code stopped only when video was **strictly longer** than audio, giving a small safety buffer. The new code stops the moment video duration **equals** audio duration exactly.

Clip durations are floats. If accumulated clips reach exactly `audio_duration` (e.g. `10.0s == 10.0s`), the loop breaks with no safety margin for floating-point drift during FFmpeg concatenation. The final video may come out fractionally short, causing the last frames to be silent or cut off.

### Reproduction scenario

```
audio_duration = 10.0s
clips: [3.0s, 4.0s, 3.0s]  →  sum = 10.0s exactly
```

With `>=`, loop breaks after 3rd clip. FFmpeg may produce `9.97s` due to frame rounding → trailing silence.

With `>` (old), loop would continue adding one more clip as buffer.

### Suggested fix

```python
if video_duration >= audio_duration + 0.1:  # small safety margin
    break
```

Sushanth012:
## Summary
- require a small 0.1s safety buffer before stopping video clip accumulation
- apply the same buffered check when looping processed clips
- add a regression test for exact audio/video duration boundaries

Fixes #985

## Testing
- uv run --with pytest python -m pytest test/services/test_video.py

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -314,7 +314,11 @@ def _format_ffmpeg_concat_path(file_path: str) -> str:
 
 
 def concat_video_clips_with_ffmpeg(
-    clip_files: List[str], output_file: str, threads: int, output_dir: str
+    clip_files: List[str],
+    output_file: str,
+    threads: int,
+    output_dir: str,
+    max_duration: float | None = None,
 ):
     concat_list_file = os.path.join(output_dir, "ffmpeg-concat-list.txt")
     with open(concat_list_file, "w", encoding="utf-8") as fp:
@@ -322,7 +326,7 @@ def concat_video_clips_with_ffmpeg(
             fp.write(f"file '{_format_ffmpeg_concat_path(clip_file)}'\n")
 
     def build_command(codec: str) -> list[str]:
-        return [
+        command = [
             utils.get_ffmpeg_binary(),
             "-y",
             "-f",
@@ -337,8 +341,11 @@ def concat_video_clips_with_ffmpeg(
             str(threads or 2),
             "-pix_fmt",
             "yuv420p",
-            output_file,
         ]
+        if max_duration is not None and max_duration > 0:
+            command.extend(["-t", f"{max_duration:.3f}"])
+        command.append(output_file)
+        return command
 
     def run_concat(codec: str):
         command = build_command(codec)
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for the focused PR.

I checked the patch and ran the video tests locally:

```bash
uv run --with pytest pytest test/services/test_video.py -q
# 25 passed
```

The concern in #985 is reasonable, and the helper/test are heading in the right direction. However, I do not want to merge this exact implementation yet because it can introduce a different behavior issue.

With the current approach, if the accumulated video duration is exactly equal to the audio duration, the code requires `audio_duration + 0.1s` before stopping. Since the pipeline appends whole clips, being short by only `0.1s` may add another full 3-5 second clip. That can make the final video noticeably longer than the audio and create a silent tail at the end.

A safer fix would need to keep the small safety margin while also preventing the final output from becoming significantly longer than the audio. For example:

- add a small buffer during source accumulation, but trim the final composed video to the audio duration; or
- trim the last added clip precisely instead of appending a full extra clip; and
- add a regression test that verifies the final video duration does not noticeably exceed the audio duration.

So I am not merging this as-is, but the issue is valid and this is a good starting point. A revised PR that avoids the long silent-tail case would be welcome.

Sushanth012:
Updated this based on your feedback. The branch is rebased onto current main, and the fix now keeps the small source-duration safety margin while capping the final ffmpeg concat output to the audio duration. This avoids the case where adding one extra whole clip creates a noticeably long silent tail.

I also removed the single-clip direct-copy shortcut so the same duration cap is applied consistently, added a regression check that combine_videos passes the audio duration cap to concat, and added a direct ffmpeg command test for the -t limit.

Validation: uv run --with pytest pytest test/services/test_video.py -q (28 passed, 11 warnings, 2 subtests passed), plus git diff --check.

harry0703:
Thanks for the update. I tested the boundary case with real generated media: three clips totaling exactly 10s and a 10s audio track.\n\nBefore this change, the safety margin could loop one extra full clip, making the combined and final video about 13s long. With this PR, the concat output is capped to the audio duration, and the final video stays at 10s as expected.\n\nThe approach keeps the small source-duration safety buffer while avoiding a visible silent tail, so I have merged it. Thanks again for the contribution.

PR #1011, opened 2026-06-09
Add ContentFactoryRequest and ContentFactoryResponse models

rahtk121-code:


harry0703:
Thanks for the contribution.

I reviewed this PR. At the moment it only adds `ContentFactoryRequest` and `ContentFactoryResponse` schema models, but there is no related API endpoint, service implementation, documentation, or test coverage using these models.

Because of that, it is hard to evaluate the actual behavior or user-facing value of this change, and merging unused schema models would add maintenance surface without enabling a feature yet.

Could you please share the intended use case for `ContentFactory`, or update the PR with the corresponding endpoint/service logic and tests? Once there is a complete feature flow, I can review it again.

harry0703:
Thank you, @rahtk121-code, for the contribution. As noted in the earlier review, this PR currently adds only unused request and response schema models, without a corresponding API endpoint, service implementation, documentation, typed output contract, or tests. Since there has been no follow-up completing the feature flow, merging these standalone models would add maintenance surface without providing user-facing functionality. We would be happy to reconsider a future PR that includes the complete ContentFactory use case and its tests, but I am closing this PR for now. We appreciate your effort.

PR #1014, opened 2026-06-10
fix(groq): add default model name + clarify unvalidated manual entry warning

### LINKED ISSUE #1013: Groq: manually entered model name not validated when model list fails to load
## What I found

In `webui/Main.py`, when `get_groq_model_ids()` fails to fetch the model list (network issue, invalid API key, temporary outage), the UI correctly falls back to a text input:

```python
except Exception as e:
    logger.warning(f"failed to fetch groq models: {e}")
    return []
```

```python
else:
    st_llm_model_name = st.text_input(
        tr("Model Name"),
        value=llm_model_name,
        key="groq_model_name_input",
    )
    if effective_api_key:
        st.caption("Unable to load Groq model list right now. You can still enter a model name manually.")
```

The problem: the manually entered model name is never validated before it reaches `_generate_response()`. If the user types a model name that doesn't exist on Groq (e.g. a typo, or a deprecated model), the generation request fails with a generic OpenAI SDK error — not a helpful "invalid model name" message.

## Reproduction

1. Set `llm_provider = "groq"` with a valid API key
2. Block network access to `https://api.groq.com/openai/v1/models` (or use a bad `base_url`)
3. UI shows text input — type any string as model name (e.g. `non-existent-model`)
4. Submit a generation request
5. Error returned is a raw API error, not "model not found"

There's also a minor related issue: failed model fetches are cached for 5 minutes (`ttl=300`), so even after the API recovers, the fallback text input stays visible until the cache expires.

## Suggested fix

Two small changes would address this cleanly:

**Option A** — Add a warning in the UI when showing the fallback text input:
```python
st.caption("Unable to load Groq model list. Enter a model name manually — note it won't be validated until generation.")
```

**Option B** — In `llm.py`, when `llm_provider == "groq"`, add a default model fallback similar to other providers:
```python
elif llm_provider == "groq":
    api_key = config.app.get("groq_api_key")
    model_name = config.app.get("groq_model_name")
    if not model_name:
        model_name = "llama-3.3-70b-versatile"  # same default as WebUI
    base_url = config.app.get("groq_base_url", "")
    if not base_url:
        base_url = "https://api.groq.com/openai/v1"
```

Happy to submit a PR if the fix direction looks right to maintainers.

kapil971390:
Fixes #1013

## Changes

**`app/services/llm.py`**
- Added default `model_name = "llama-3.3-70b-versatile"` when `groq_model_name` is not set in config — consistent with how other providers handle missing model names (`g4f`, `gemini`, `mimo`, `aihubmix` all set defaults here rather than relying solely on the WebUI default)

**`webui/Main.py`**
- Updated caption shown when model list fails to load — now explicitly says the manually entered name won't be validated until generation, so users understand a wrong model name will fail at request time rather than immediately

## Why

When `get_groq_model_ids()` fails (network issue, outage, bad API key), the UI falls back to a text input. The previous caption said "you can still enter a model name manually" but didn't warn that the name would go unvalidated. Combined with `@st.cache_data(ttl=300)`, a user could spend 5 minutes trying a wrong model name before realising the issue.

The `llm.py` default also helps users who configure Groq via `config.toml` directly without the WebUI — they get a working default instead of a validation error if they forget to set `groq_model_name`.

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -212,6 +212,8 @@ def _generate_response(prompt: str) -> str:
             elif llm_provider == "groq":
                 api_key = config.app.get("groq_api_key")
                 model_name = config.app.get("groq_model_name")
+                if not model_name:
+                    model_name = "llama-3.3-70b-versatile"
                 base_url = config.app.get("groq_base_url", "")
                 if not base_url:
                     base_url = "https://api.groq.com/openai/v1"
--- a/webui/Main.py
+++ b/webui/Main.py
@@ -590,7 +590,7 @@ if not config.app.get("hide_config", False):
                         )
                         if effective_api_key:
                             st.caption(
-                                "Unable to load Groq model list right now. You can still enter a model name manually."
+                                "Unable to load Groq model list right now. You can still enter a model name manually — note it won't be validated until generation."
                             )
                         else:
                             st.caption(


kapil971390:
Thanks for the quick review and merge, 
@harry0703! Happy to contribute. 🙏

PR #1015, opened 2026-06-10
Fix video duration buffer without silent tail

### LINKED ISSUE #985: bug: >= comparison in video clip selection may produce audio trailing silence (commit 1e56568)

Auglengroup:
## Summary
- keep a small 0.1s source-video buffer before stopping clip accumulation to avoid frame rounding making the final video slightly shorter than narration
- trim the combined video back to the narration duration during ffmpeg concat or the single-clip path, so the buffer does not create a long silent tail
- add regression coverage for the buffer helper and ffmpeg `-t` trimming command

Fixes #985

## Testing
- `.venv/bin/python -m pytest test/services/test_video.py -q`
- `.venv/bin/python -m compileall app test`
- `git diff --check`

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -78,6 +78,11 @@ _SUPPORTED_VIDEO_CODECS = (
     "h264_videotoolbox",
 )
 _runtime_disabled_video_codecs = set()
+_VIDEO_DURATION_SAFETY_BUFFER_SECONDS = 0.1
+
+
+def _has_enough_video_duration(video_duration: float, audio_duration: float) -> bool:
+    return video_duration >= audio_duration + _VIDEO_DURATION_SAFETY_BUFFER_SECONDS
 
 
 def _prioritize_unique_source_clips(
@@ -279,7 +284,11 @@ def _format_ffmpeg_concat_path(file_path: str) -> str:
 
 
 def concat_video_clips_with_ffmpeg(
-    clip_files: List[str], output_file: str, threads: int, output_dir: str
+    clip_files: List[str],
+    output_file: str,
+    threads: int,
+    output_dir: str,
+    max_duration: float | None = None,
 ):
     concat_list_file = os.path.join(output_dir, "ffmpeg-concat-list.txt")
     with open(concat_list_file, "w", encoding="utf-8") as fp:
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the focused fix. The idea makes sense: keep a tiny video-duration buffer to avoid frame rounding issues, then trim the combined output back to the narration duration so there is no silent tail.

I tested the branch locally, but the current test suite fails:

```bash
uv run --with pytest pytest test/services/test_video.py -q
# 1 failed, 25 passed
```

The failing test is:

```text
TestVideoService.test_concat_video_clips_trims_to_audio_duration
```

Root cause: the test mocks `subprocess.run` with a helper that only accepts `command, capture_output, text, check`, but the existing encoder-detection path calls `subprocess.run(..., timeout=10)`. That causes:

```text
TypeError: fake_run() got an unexpected keyword argument 'timeout'
```

Could you please update the test mock to accept `**kwargs` or otherwise handle the encoder-detection call? After the tests pass, I would also like to validate this with a real short video generation case, because this PR changes the final duration/trim behavior in the main composition path.

harry0703:
Thank you, @Auglengroup, for the thoughtful fix and the clear regression coverage. Since this PR was opened, the same behavior has been incorporated into `main` through #1003 and follow-up maintainer refinements: video selection keeps a 0.1-second safety margin for FFmpeg frame rounding, and the final concat output is trimmed back to the narration duration to avoid a silent tail. The current implementation also uses a unified concat path, and the corresponding regression tests pass. Since there is no remaining incremental change to absorb, I am closing this PR as superseded. We appreciate your contribution.

PR #1021, opened 2026-06-11
fix: rename misspelled video_contact_mode parameter to video_concat_mode

reyys:
## What

Renames the `download_videos()` parameter `video_contact_mode` to `video_concat_mode` ("contact" was a misspelling of "concat").

## Why

The parameter definition in `app/services/material.py`, its call site in `app/services/task.py`, and the test in `test/services/test_material.py` were all consistently misspelled, so the code worked — but the name was inconsistent with `video_concat_mode` used everywhere else in the codebase (`VideoParams` schema, video service, webui). This makes the API consistent and less confusing.

## Changes

- `app/services/material.py`: parameter definition + internal usage
- `app/services/task.py`: keyword argument at the call site
- `test/services/test_material.py`: keyword argument in test

## Testing

All 11 tests in `test/services/test_material.py` pass:

```
Ran 11 tests in 0.003s
OK
```

No functional change — pure rename, no external API or config key involved.

### DIFF
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -306,7 +306,7 @@ def download_videos(
     search_terms: List[str],
     source: str = "pexels",
     video_aspect: VideoAspect = VideoAspect.portrait,
-    video_contact_mode: VideoConcatMode = VideoConcatMode.random,
+    video_concat_mode: VideoConcatMode = VideoConcatMode.random,
     audio_duration: float = 0.0,
     max_clip_duration: int = 5,
 ) -> List[str]:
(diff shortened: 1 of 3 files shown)


harry0703:
Thanks for the cleanup.

I verified the merge result and confirmed that the old `video_contact_mode` spelling is no longer used internally. The related material/task tests pass, and this change keeps the internal naming consistent with the existing `video_concat_mode` field used by the schema, WebUI, and video service.

This is a small and useful maintainability fix, so I am merging it. Much appreciated.

PR #1026, opened 2026-06-11
fix(llm): robustly parse JSON from fenced / multi-line LLM responses in generate_terms

xlab-nyc:
**What**

`generate_terms()` can return an empty list when the LLM wraps its JSON array
in a markdown ```` ```json ```` fence or spreads it across multiple lines:

- the primary `json.loads(response)` raises on the fence / prose prefix, and
- the regex recovery `re.search(r"\[.*]", response)` has no `re.DOTALL`, so a
  multi-line array never matches.

The result is `search_terms == []`, which means **no stock footage is fetched**
and the generated video is degraded/blank — with only a `warning` in the log.

Non-OpenAI providers (Claude, Gemini, …) emit fenced / multi-line JSON far more
often than GPT, so this disproportionately hits those backends.

Notably, `_parse_social_metadata()` already anticipates this exact shape (it
uses `re.DOTALL` and a comment about models wrapping JSON in a markdown fence) —
this PR just brings `generate_terms()` in line with that established handling.

**Change**

- add a small `_strip_code_fence()` helper that removes a surrounding
  ```` ```json … ``` ```` fence before the first `json.loads()`
- add `re.DOTALL` to the `generate_terms` array-recovery regex
- apply the fence strip in `_parse_social_metadata()` too, for consistency

No behavioural change for already-valid responses; purely additive recovery.

**Test (standalone repro)**

| input shape | before | after |
|---|---|---|
| single-line fenced array | OK (via fallback) | OK |
| **multi-line fenced array** | **empty** | OK |
| **prose + multi-line fenced array** | **empty** | OK |

Reproduced live against an OpenAI-compatible Anthropic endpoint (`claude-haiku-4-5`),
where the subject "the quiet discipline of compounding habits" triggered the
`failed to generate video terms: Expecting value: line 1 column 1 (char 0)`
warning before this change.

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -694,6 +694,22 @@ def generate_script(
     return final_script.strip()
 
 
+def _strip_code_fence(text: str) -> str:
+    """Strip a surrounding markdown code fence from an LLM response.
+
+    Non-OpenAI providers (Claude, Gemini, …) frequently wrap JSON output in a
+    ```json … ``` fence even when asked to return raw JSON. Removing it lets the
+    first json.loads() succeed instead of falling through to the regex recovery
+    path (and spuriously logging a warning). Mirrors the DOTALL handling already
+    used in _parse_social_metadata().
+    """
+    t = (text or "").strip()
+    if t.startswith("```"):
+        t = re.sub(r"^```[a-zA-Z0-9]*\s*", "", t)
+        t = re.sub(r"\s*```$", "", t)
+    return t.strip()
+
+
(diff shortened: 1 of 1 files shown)


harry0703:
Thanks for the focused fix.

I verified the merge result with the existing LLM test suite and additional mocked responses covering fenced JSON arrays, multi-line arrays, prose plus fenced JSON, and fenced social metadata JSON. Everything behaved as expected, so I am merging this.

Much appreciated.

PR #1029, opened 2026-06-12
fix: write moviepy temp audio to system temp dir to avoid Windows Defender interference

### LINKED ISSUE #217: 连续调用接口`/api/v1/videos`，生成视频错误
连续调用`/api/v1/videos`接口多次（差不多每隔几秒调用一次，连续4~5次），只有一个生成成功了，其他的都失败。同时发现一个奇怪的现象，临时mp3文件(类似`final-1.mp4.tempTEMP_MPY_wvf_snd.mp3`)生成到根目录了（结束后又自动删掉）。

附上log：
[2024-04-10.txt](https://github.com/harry0703/MoneyPrinterTurbo/files/14926706/2024-04-10.txt)

mertfurkandgn:
## Summary

On Windows, MoviePy writes a TEMP_MPY_wvf_snd file next to the output video in the task directory. Windows Defender scans files written to this directory and can lock/delay the temp file, causing the final MP4 to be written as 0 bytes (moov atom not found error).

Fix: write the temp audio file to tempfile.gettempdir() (system temp dir) which Defender does not actively scan during video generation.

## Related

Fixes #217

## Test plan

- Generate a video on Windows with Windows Defender active
- Confirm final-1.mp4 is written correctly (non-zero size, playable)
- Confirm no regression on Linux/macOS (tempfile.gettempdir() returns the standard temp dir on all platforms)

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -6,6 +6,7 @@ import random
 import gc
 import shutil
 import subprocess
+import tempfile
 from contextlib import redirect_stdout
 from functools import lru_cache
 from typing import List
@@ -973,6 +974,7 @@ def generate_video(
     # 显式沿用输入音频的采样率；如果取不到，再回退到 MoviePy 默认的 44100Hz。
     # 这样可以减少不同运行环境，尤其是 Docker 环境中再次重采样带来的音质波动。
     output_audio_fps = int(getattr(audio_clip, "fps", 0) or 44100)
+    temp_audio_dir = tempfile.gettempdir()
     _write_videofile_with_codec_fallback(
         video_clip,
         output_file=output_file,
@@ -980,7 +982,7 @@ def generate_video(
         audio_codec=audio_codec,
         audio_fps=output_audio_fps,
         audio_bitrate=audio_bitrate,
-        temp_audiofile_path=output_dir,
+        temp_audiofile_path=temp_audio_dir,
         threads=params.n_threads or 2,
         logger=None,
         fps=fps,
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -6,6 +6,7 @@ import random
 import gc
 import shutil
 import subprocess
+import sys
 import tempfile
 from contextlib import redirect_stdout
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for the PR and for documenting the Windows Defender angle in detail.

The idea is useful, but I do not want to merge this exact change directly for now. The current `temp_audiofile_path=output_dir` behavior was originally added for #217, and that issue was later confirmed as fixed by the reporter. This PR changes the temp audio directory for all platforms, not only Windows, and it also reverses the intent of the existing comment around the previous Windows temp-file handling.

My main concern is that using the system temp directory globally could introduce new behavior on macOS/Linux/Docker or in environments where the system temp directory is on a different filesystem, has different cleanup behavior, or has different permissions. The Windows Defender hypothesis is still worth considering, but I think we should absorb it more conservatively if we change this area.

A safer direction would be something like:

- keep the current default behavior unchanged;
- only consider the system temp directory for Windows-specific cases;
- put the directory selection behind a small helper so the behavior is explicit and testable;
- update the existing comment so it does not conflict with the implementation;
- add a focused test for the temp audio directory selection.

So for now I will not merge this PR as-is, but I appreciate the investigation and may reuse the idea in a more narrowly scoped Windows-specific fix.

mertfurkandgn:
Thank you for the detailed feedback. I have updated the PR to address all your concerns:

- Added a dedicated `_get_temp_audio_dir(output_dir)` helper that makes the platform-specific logic explicit and testable.
- The system temp directory is now used **only on Windows** (`sys.platform == 'win32'`). Linux, macOS, and Docker continue to use `output_dir` exactly as before.
- The existing comment around `#217` is preserved and the new helper's docstring explains the Windows Defender rationale separately.
- Two focused unit tests are added for the helper covering both the Windows path and the non-Windows fallback.

harry0703:
Thank you for the update and for addressing the review feedback.

The revised version is much safer now: it keeps the existing behavior on macOS, Linux, and Docker, and only uses the system temp directory on Windows to avoid possible Windows Defender locks on MoviePy's temporary audio file.

We also verified this locally with the video service test suite, and the Windows-side validation passed as well. This should improve video generation stability for Windows users without changing the normal generation workflow.

I will merge this PR. Thanks again for the contribution.

PR #1030, opened 2026-06-12
Prevent Streamlit first-run prompt from blocking WebUI startup

BinaryOutlook:
## Summary
- Disable Streamlit's first-run email prompt across WebUI launch paths so startup does not block before the dashboard is available.
- Update README launch examples to include the same Streamlit flag.
- Clarify API key quote formatting in `config.example.toml` in English and Chinese to avoid TOML parse errors from smart quotes.

## Validation
- `python -c "import tomllib; tomllib.load(open('config.example.toml','rb')); print('config.example.toml parsed OK')"`
- `git diff --cached --check`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,7 @@ COPY . .
 EXPOSE 8501
 
 # Command to run the application
-CMD ["streamlit", "run", "./webui/Main.py","--browser.serverAddress=127.0.0.1","--server.enableCORS=True","--browser.gatherUsageStats=False"]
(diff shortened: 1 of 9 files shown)


harry0703:
Thanks for the contribution.

This is a useful startup reliability fix: disabling Streamlit's first-run email prompt prevents the WebUI launch path from blocking before the dashboard becomes available. I also verified that `--server.showEmailPrompt=False` is supported by the project's current Streamlit version and that the WebUI can start successfully with this option.

Merged. Thanks again!

PR #1031, opened 2026-06-13
Add basic CI and address targeted review items

BinaryOutlook:
## Summary
- add a basic GitHub Actions CI workflow for Python compile and unittest discovery
- make `MemoryState` thread-safe and return isolated task snapshots
- validate `custom_audio_file` paths before use, fail explicitly for unsupported `VideoAspect`, and document Redis state trust boundaries
- fill two missing Russian WebUI translation keys so full unittest discovery stays green

## Before / After
Before:
- no repository CI workflow covered compile or unit test discovery
- in-memory task state used an unlocked mutable dictionary
- custom audio paths were checked directly from caller input
- unsupported video aspects silently fell back to portrait resolution
- Redis task-state deserialization trust assumptions were implicit

After:
- CI runs `uv sync --frozen`, `compileall`, and `python -m unittest discover -s test`
- `MemoryState` uses an `RLock` and returns deep-copied snapshots
- custom audio must resolve inside the task directory before duration/open checks
- unsupported aspect values raise `ValueError`
- Redis trust boundary is documented in code and example config

## Testing
- `git diff --check origin/main...HEAD`
- `./.venv/Scripts/python.exe -m compileall app cli.py main.py webui test`
- `./.venv/Scripts/python.exe -X utf8 -m unittest test.services.test_state test.services.test_task test.services.test_schema test.services.test_webui_i18n`
- `./.venv/Scripts/python.exe -X utf8 -m unittest discover -s test`

Full suite result: 137 tests passed, 3 skipped.


harry0703:
Thanks for the detailed cleanup and for adding a first CI workflow. The direction is useful, but I do not think this PR is ready to merge as-is yet.

A couple of points need to be addressed first:

1. The CI workflow currently runs full unittest discovery. That includes tests which may call external TTS/network services, so unrelated PRs could fail because of transient network or provider issues. The CI suite should be narrowed to deterministic tests, or those integration-style tests should be skipped unless explicitly enabled.

2. The `custom_audio_file` change restricts audio files to the generated task directory. That improves safety, but the current API does not provide a way to upload audio into that per-task directory before task creation. Existing API clients that pass a server-side audio path would silently fall back to TTS, which is a behavior regression.

Could you update the PR to handle those two points? After that I can review it again. Thanks again for working on this.

BinaryOutlook:
Thanks for the review. I pushed a follow-up commit that addresses both blockers:

- CI now runs a deterministic unittest slice instead of full discovery, so unrelated PRs should not fail on live TTS/network/provider behavior.
- Live provider tests are gated behind `MPT_RUN_INTEGRATION_TESTS=1`, with the test README documenting how to opt in.
- `custom_audio_file` now keeps task-local file support and also accepts existing server-side audio paths, preserving compatibility for API clients that already pass server-side files.
- Invalid/missing custom audio paths now fail the audio step explicitly instead of silently falling back to TTS.

Local verification:

- `git diff --check`
- `python -m compileall app cli.py main.py webui test`
- `python -X utf8 -m unittest test.services.test_state test.services.test_task test.services.test_schema test.services.test_webui_i18n`
- `python -X utf8 -m unittest discover -s test` -> 138 tests passed, 5 skipped

Could you please review again when you have a chance?

harry0703:
Thanks for the follow-up update.

I reviewed the latest changes again and verified them locally with:

- `uv run python -m compileall app cli.py main.py webui test`
- `uv run python -X utf8 -m unittest test.services.test_state test.services.test_task test.services.test_schema test.services.test_webui_i18n`
- `uv run python -X utf8 -m unittest discover -s test`

The two earlier blockers are addressed:

- CI now runs a deterministic test subset, while live provider tests are gated behind `MPT_RUN_INTEGRATION_TESTS=1`.
- `custom_audio_file` no longer silently falls back to TTS for invalid paths, while existing server-side audio file compatibility is preserved.

This is a useful improvement for the project's baseline quality and regression protection, so I have merged it. Thanks again for the careful work and for responding to the review feedback.

PR #1033, opened 2026-06-13
fix(cli): validate --video-source local constraints at parse time

### LINKED ISSUE #1032: cli: --video-source local has two validation gaps — silent empty terms and late failure when --video-materials is missing
## Summary

The new CLI (`cli.py`, added in #1005) has two related validation gaps when `--video-source local` is used.

---

### Bug 1 — `--stop-at terms` silently returns empty terms for local source

**Reproduction:**
```bash
python cli.py \
  --video-subject "cats" \
  --video-source local \
  --video-materials "cat.mp4" \
  --stop-at terms
```

**Output:**
```json
{"task_id": "...", "result": {"script": "...", "terms": ""}}
```

**Why it happens:**  
In `task.py`, `generate_terms()` is intentionally skipped for local sources (`if params.video_source != "local"`). However, `"terms"` is still listed as a valid `--stop-at` choice in the CLI without any warning. The pipeline reaches `if stop_at == "terms"` and returns `{"terms": ""}` — no error, no indication that terms are not generated for local sources.

**Expected behaviour:**  
Either raise an error when `--stop-at terms` is combined with `--video-source local`, or add a note in the help text that terms are not generated for local sources.

---

### Bug 2 — Missing `--video-materials` is only caught after expensive LLM + TTS steps

**Reproduction:**
```bash
python cli.py \
  --video-subject "cats" \
  --video-source local
  # --video-materials omitted
```

**What happens:**  
The pipeline runs through:
1. LLM script generation (Step 1) ✓
2. TTS audio generation (Step 3) ✓
3. Subtitle generation (Step 4) ✓
4. **Fails at Step 5** — "no valid materials found"

`preprocess_video(materials=None)` returns `[]`, which triggers a failure — but only after all the preceding steps have already completed. For local source, the absence of `--video-materials` makes the run unrecoverable from the start.

**Expected behaviour:**  
Fail fast at argument-parsing time:
```python
if args.video_source == "local" and not args.video_materials:
    parser.error("--video-materials is required when --video-source is local")
```

---

Both issues stem from the CLI not encoding the constraints that `task.py` already enforces internally. Happy to submit a small PR for both if this looks reasonable.

kapil971390:
Fixes #1032

## What

Two validation gaps in `cli.py` when `--video-source local` is used:

**Bug 1 — late failure when `--video-materials` is missing**

Without this fix, omitting `--video-materials` with `--video-source local` allows the pipeline to run through LLM script generation, TTS audio, and subtitle generation before failing at the materials step with a generic "no valid materials found" message. All that work is wasted.

**Bug 2 — `--stop-at terms` silently returns empty terms**

`"terms"` is listed as a valid `--stop-at` choice, but `generate_terms()` is intentionally skipped for local sources in `task.py`. This caused the pipeline to return `{"terms": ""}` with no error or warning.

## Fix

Both constraints are now checked in `parse_args()` via `parser.error()`, which exits immediately with a clear message before any work begins:

```
error: --video-materials is required when --video-source is local
error: --stop-at terms has no effect with --video-source local (search terms are not generated for local sources)
```

## Changes

- `cli.py` — two `parser.error()` guards added at the end of `parse_args()`
- `test/services/test_cli.py` — two tests covering the new validation paths

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -52,7 +52,18 @@ def parse_args(argv: Sequence[str] | None = None) -> argparse.Namespace:
         help="enable subtitles (default: enabled, use --no-subtitle-enabled to disable)",
     )
     parser.add_argument("--task-id", default="", help="custom task id")
-    return parser.parse_args(argv)
+    args = parser.parse_args(argv)
+
+    if args.video_source == "local" and not (args.video_materials or "").strip():
+        parser.error("--video-materials is required when --video-source is local")
+
+    if args.video_source == "local" and args.stop_at == "terms":
+        parser.error(
+            "--stop-at terms has no effect with --video-source local "
+            "(search terms are not generated for local sources)"
+        )
+
+    return args
 
 
 def build_video_params(args: argparse.Namespace) -> VideoParams:
--- a/test/services/test_cli.py
+++ b/test/services/test_cli.py
@@ -105,5 +105,21 @@ class TestCli(unittest.TestCase):
                 os.remove(test_filepath)
 
 
+    def test_local_source_requires_video_materials(self):
+        with self.assertRaises(SystemExit) as cm:
+            cli.parse_args(["--video-subject", "test", "--video-source", "local"])
+        self.assertNotEqual(cm.exception.code, 0)
+
+    def test_local_source_stop_at_terms_rejected(self):
+        with self.assertRaises(SystemExit) as cm:
+            cli.parse_args([
+                "--video-subject", "test",
+                "--video-source", "local",
+                "--video-materials", "a.mp4",
+                "--stop-at", "terms",
+            ])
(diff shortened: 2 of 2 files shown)


harry0703:
Thanks for the contribution.

I verified this locally with:

- `uv run python -m unittest test.services.test_cli`
- `uv run python cli.py --video-subject cats --video-source local`
- `uv run python cli.py --video-subject cats --video-source local --video-materials cat.mp4 --stop-at terms`

Both invalid local-source CLI combinations now fail early with clear parser errors instead of wasting work or returning empty terms. Merged. Thanks again!

kapil971390:
Happy to help! The local-source constraints were easy to miss since task.py enforces them internally — makes sense to surface them early at the CLI layer so users get a clear message before any work begins. Let me know if anything else comes up.

PR #1048, opened 2026-06-17
[codex] Run full unit test suite in CI

rongdu88668-ux:
## Summary

- Run the full unittest discovery suite in CI instead of a fixed subset of four test modules.
- Ensure Gemini TTS creates the output directory before exporting MP3 audio and closes the export handle.

## Why

`python -m unittest discover -s test` currently exercises the existing deterministic tests, but the CI workflow only ran a small subset. Enabling discovery keeps future `test_*.py` files covered automatically. While validating the full suite, the Gemini TTS test exposed that its MP3 export path did not create the target directory first.

## Validation

- `.venv/bin/python -m compileall app cli.py main.py webui test`
- `.venv/bin/python -X utf8 -m unittest discover -s test`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,5 +31,5 @@ jobs:
       - name: Compile Python sources
         run: uv run python -m compileall app cli.py main.py webui test
 
-      - name: Run deterministic unit tests
-        run: uv run python -X utf8 -m unittest test.services.test_state test.services.test_task test.services.test_schema test.services.test_webui_i18n
+      - name: Run unit tests
+        run: uv run python -X utf8 -m unittest discover -s test
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -1019,7 +1019,10 @@ def gemini_tts(
             return None
         
         # 导出为MP3格式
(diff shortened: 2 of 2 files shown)


harry0703:
Thank you for the PR.

Both changes look useful:

- Running `unittest discover -s test` in CI should give us better coverage as new `test_*.py` files are added.
- The Gemini TTS export change is also reasonable. While checking the current test suite, I saw the same unclosed file warning around `audio_segment.export(...)`, so explicitly closing the returned handle is a good cleanup.

Since this PR is still marked as draft, I will not merge it yet. Before merging, I would like to keep an eye on CI runtime and possible flaky/network-sensitive tests when switching from the current deterministic subset to full discovery.

Thanks again for improving the test coverage path.

harry0703:
Thank you, @rongdu88668-ux, for contributing this and for highlighting the gap in our CI test coverage. We have incorporated the core idea and expanded it into the complete pytest suite, including branch coverage enforcement and additional regression tests for task pipelines, controllers, queues, media handling, and application startup. The maintainer-side commit also preserves your authorship with a Co-authored-by trailer. We will maintain the final version directly, so I am closing this PR. We really appreciate your contribution.

Issue #1049, opened 2026-06-17
bug: LLM error path leaks embedded credentials from base_url in exception messages


kapil971390:
## Description

When a user configures a custom `base_url` with embedded credentials (e.g. `https://user:pass@proxy.example.com/v1`), any LLM API failure causes the exception message — which includes the full URL — to be returned to the caller unsanitized via the outer `except` block in `app/services/llm.py`.

This was noted by @ether-btc in #986 (closed). Opening as a separate issue as requested.

## Affected code

**`app/services/llm.py` — outer exception handler (~line 559):**

```python
except Exception as e:
    return f"Error: {str(e)}"   # ← str(e) may contain the full base_url with user:pass
```

The OpenAI SDK (and requests library) includes the target URL in exception messages when a connection or auth error occurs. For providers like `groq`, `openai`, `deepseek`, `minimax`, `mimo`, and `modelscope`, the `base_url` is read directly from user config:

```python
base_url = config.app.get("groq_base_url", "")   # user-supplied, may contain credentials
# ...
client = OpenAI(api_key=api_key, base_url=base_url)
response = client.chat.completions.create(...)    # if this raises, str(e) includes base_url
```

## Steps to reproduce

```python
# config.toml
groq_base_url = "https://myuser:mypassword@proxy.example.com/openai/v1"
groq_api_key  = "gsk_..."
llm_provider  = "groq"
```

1. Set `groq_base_url` to a URL with embedded credentials pointing to a non-existent host
2. Trigger any LLM generation
3. Observe the returned string contains `mypassword` from the URL

## Affected providers

Any provider that reads `*_base_url` from user config and passes it to `OpenAI(base_url=...)`:
`groq`, `openai`, `deepseek`, `minimax`, `mimo`, `modelscope`, and others.

## Suggested fix

Add a helper to strip `user:pass@` from URLs in error messages before returning:

```python
import re

def _sanitize_error_message(msg: str) -> str:
    """Strip embedded credentials from any URLs in the message."""
    return re.sub(r'(https?://)([^:@/\s]+:[^@/\s]+@)', r'\1***:***@', str(msg))

# in the outer except block:
except Exception as e:
    return f"Error: {_sanitize_error_message(e)}"
```

Happy to submit a PR with this fix.

harry0703:
Thank you for splitting this out and for the clear report.

This has been fixed on `main` in commit `a810d67`.

What changed:

[...]

Issue #986, opened 2026-06-03
bug: FFmpeg failure in no-voice mode silently corrupts video output (commit 76d7e09)


kapil971390:
## Bug Report

**Commit:** `76d7e09` — *feat: add no voice synthesis mode*

---

### The issue

`generate_silent_audio()` uses `subprocess.run(..., check=False)` — meaning FFmpeg failures are **not raised as exceptions**. The function logs the error and returns `False`:

```python
def generate_silent_audio(output_file, duration):
    result = subprocess.run([ffmpeg_binary, ...], check=False)
    if result.returncode != 0:
        logger.error(f"Failed to generate silent audio: {result.stderr}")
        return False   # ← silent failure
```

The caller in video composition **does not validate** this return value before passing the audio path downstream. FFmpeg then fails at the final composition stage with a cryptic file-not-found error — far from the actual root cause.

### Impact

- User sets `voice_name = "no-voice"` or `"none"`
- FFmpeg is misconfigured or `IMAGEIO_FFMPEG_EXE` points to wrong path
- `generate_silent_audio` returns `False`
- Video generation proceeds, fails at composition with no clear error message
- User has no idea the issue was FFmpeg, not their script/config

### Steps to reproduce

```bash
# Point to non-existent FFmpeg
export IMAGEIO_FFMPEG_EXE=/tmp/nonexistent-ffmpeg

# Run generation with no-voice mode
# Expected: clear error "FFmpeg not found at /tmp/nonexistent-ffmpeg"
# Actual:   cryptic failure during video composition
```

### Suggested fix

```python
if not generate_silent_audio(voice_file, duration):
    raise RuntimeError(
        f"Failed to generate silent audio — check FFmpeg installation. voice_file={voice_file}"
    )
```

ether-btc:
Hi @kapil971390 — we hit this in our [ether-btc/mpt-minimax](https://github.com/ether-btc/mpt-minimax) fork and have a fix.

[...]

PR #1050, opened 2026-06-17
fix(llm): sanitize embedded credentials from LLM error messages

kapil971390:
## Summary

Closes / references #1049 (as requested in the discussion on #986).

When a provider is configured with a `*_base_url` that contains embedded credentials (`https://user:pass@host/...`), the OpenAI SDK may raise an exception whose `str()` representation includes the raw URL. The outer `except Exception as e: return f"Error: {str(e)}"` block in `_generate_response` was surfacing that string verbatim, leaking credentials into:

- API responses returned to callers
- Any logging layer that records return values

## Fix

Added `_sanitize_error_message()` — a one-liner regex that strips the `userinfo@` component from any `http://` or `https://` URL found in the exception message — and routed the except block through it:

```python
def _sanitize_error_message(msg) -> str:
    """Strip embedded credentials from any URLs that appear in exception messages."""
    return re.sub(r"(https?://)([^:@/\s]+:[^@/\s]+@)", r"\1***:***@", str(msg))

# before:
except Exception as e:
    return f"Error: {str(e)}"

# after:
except Exception as e:
    return f"Error: {_sanitize_error_message(e)}"
```

`re` is already imported in the module so no new dependency is introduced.

## Test plan

- [ ] Configure any provider with `base_url = "https://myuser:mysecret@proxy.example.com/v1"` and trigger an exception (e.g. bad API key, network timeout)
- [ ] Confirm the returned error string shows `https://***:***@proxy.example.com/v1` instead of the raw credentials
- [ ] Confirm normal error messages (no URL) are returned unchanged

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -547,7 +547,12 @@ def _generate_response(prompt: str) -> str:
 
         return _normalize_text_response(content, llm_provider)
     except Exception as e:
-        return f"Error: {str(e)}"
+        return f"Error: {_sanitize_error_message(e)}"
+
+
+def _sanitize_error_message(msg) -> str:
+    """Strip embedded credentials from any URLs that appear in exception messages."""
+    return re.sub(r"(https?://)([^:@/\s]+:[^@/\s]+@)", r"\1***:***@", str(msg))
 
 
 def _limit_script_text(text: str | None, max_length: int, field_name: str) -> str:


harry0703:
Thank you for opening the PR and for following up on #1049.

This is the same issue we just fixed on `main` in commit `a810d67`, so I am going to close this PR to avoid duplicating the change.

The merged fix covers the same core behavior from this PR:

- LLM errors are sanitized before returning `Error: ...`.
- Embedded URL credentials such as `https://user:pass@example.com/v1` are redacted to `https://***:***@example.com/v1`.

The version merged on `main` also adds a bit more coverage:

- common sensitive query parameters such as `api_key`, `access_token`, `token`, `key`, `secret`, and `password` are redacted as well;
- regression tests were added for both the sanitizer helper and the OpenAI-compatible provider error path.

Validation already passed with:

- `uv run python -X utf8 -m unittest test.services.test_llm`
- `uv run python -m compileall app/services/llm.py test/services/test_llm.py`

Thanks again for the security-focused contribution. The report and PR helped confirm the right fix path.

PR #1057, opened 2026-06-18
feat(ui): add optional step-by-step Wizard Mode for settings

Shivay4226:
Problem:
--------
The settings page displays all 5 categories (Script, Video, Audio, Subtitles, Configuration) simultaneously in tabs. First-time users frequently feel overwhelmed by the number of options and are unsure where to begin, which increases the chance of misconfiguration and leads to a poor first-run experience.

Solution:
---------
Introduce an optional Wizard Mode that guides users through settings one step at a time while preserving the original tab layout for power users.

Detailed Changes:
-----------------
1. Session State Management
   - Added `use_wizard` (bool, defaults to True) to track whether the user prefers the wizard or classic tab view.
   - Added `wizard_step` (int, defaults to 0) to remember the current wizard position across rerenders.
   - Added `video_params_dict` (dict) to persist `VideoParams` values when navigating between steps.

2. Conditional Layout Rendering
   - Replaced the hard-coded `st.tabs(...)` call with a conditional: * Wizard ON  -> creates 5 empty `st.container()` placeholders and renders only the container matching the current step. * Wizard OFF -> creates the original 5 tabs so experienced users see no difference.

3. Step Indicators & Navigation
   - Added a `st.progress(...)` bar showing "Step X / 5: <Category>".
   - Added a caption explaining how to navigate.
   - Added Previous / Next buttons at the bottom of the page.
   - Navigation updates `st.session_state["wizard_step"]` and triggers a `st.rerun()` so the correct step renders immediately.

4. Generate Button Placement
   - The "Generate Video" primary button now only appears on the final step when Wizard Mode is active. When Wizard Mode is disabled, the button remains in its original position right after the tabs.

5. State Persistence Across Steps
   - Before every script rerun, `VideoParams` is serialized (via `model_dump()` or the legacy `.dict()`) and stored in `st.session_state["video_params_dict"]`.
   - On the next run, `params` is reconstructed from this dict instead of starting from defaults, so user inputs survive step changes.
   - File uploaders now use explicit `key` arguments (`local_video_files`, `custom_audio_file_uploader`) so Streamlit keeps their state in `session_state` even when the widget is not currently rendered.

Why This Approach:
------------------
- Reduces cognitive load for newcomers by surfacing only relevant fields per step, similar to onboarding flows in modern web apps.
- Zero breaking change: the toggle defaults to wizard mode, but unchecking it restores the exact previous tab behaviour.
- State-safe by design: because Streamlit reruns the entire script on every interaction, persisting values in `session_state` is the idiomatic way to keep data alive across renders.

Testing:
--------
1. Start the app: streamlit run webui/Main.py
2. Observe the progress bar and step-by-step layout.
3. Fill values on Step 1, click Next, then Previous — all inputs should remain.
4. Upload local video files on Step 2, go to Step 3 and back — files should still be present.
5. Disable Wizard Mode — the classic tabs should appear exactly as before.
image

### DIFF
--- a/webui/Main.py
+++ b/webui/Main.py
@@ -252,439 +252,22 @@ def get_groq_model_ids(api_key: str, base_url: str) -> list[str]:
         logger.warning(f"failed to fetch groq models: {e}")
         return []
 
-# 创建基础设置折叠框
+# 创建基础设置折叠框 - 只保留日志设置
 if not config.app.get("hide_config", False):
     with st.expander(tr("Basic Settings"), expanded=False):
-        config_panels = st.columns(3)
-        left_config_panel = config_panels[0]
-        middle_config_panel = config_panels[1]
-        right_config_panel = config_panels[2]
+        # 是否隐藏配置面板
+        hide_config = st.checkbox(
+            tr("Hide Basic Settings"), value=config.app.get("hide_config", False)
+        )
+        config.app["hide_config"] = hide_config
 
-        # 左侧面板 - 日志设置
-        with left_config_panel:
-            # 是否隐藏配置面板
-            hide_config = st.checkbox(
-                tr("Hide Basic Settings"), value=config.app.get("hide_config", False)
-            )
-            config.app["hide_config"] = hide_config
-
-            # 是否禁用日志显示
-            hide_log = st.checkbox(
-                tr("Hide Log"), value=config.ui.get("hide_log", False)
-            )
-            config.ui["hide_log"] = hide_log
-
-        # 中间面板 - LLM 设置
-
-        with middle_config_panel:
-            st.write(tr("LLM Settings"))
-            # 下拉框需要展示“AIHubMix（推荐）”这类面向用户的文案，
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for the contribution and for focusing on the first-run experience.

After reviewing and running the submitted WebUI, we have decided not to merge this PR. The implementation does not currently match the described Wizard Mode: there is no wizard toggle, step state, progress indicator, Previous/Next navigation, or cross-step parameter snapshot. The actual change reorganizes the existing settings into five tabs, while the Generate Video button remains permanently visible.

The current main branch has also evolved substantially and now includes a one-time onboarding tour, a dedicated settings dialog, task restoration, upload dependency handling, and centralized widget-state management. This PR conflicts with that structure, and adapting the old tab-based implementation would risk regressing those newer capabilities for a relatively limited usability benefit.

We appreciate the idea and the effort. If a full wizard is reconsidered in the future, it should be designed against the current modular WebUI and include automated state-navigation and upload-lifecycle tests.

PR #1059, opened 2026-06-20
feat: add Korean localization

remo-creator-ai:
## Summary
- Add Korean WebUI localization via `webui/i18n/ko.json`
- Add `ko-KR` as a supported script language option
- Add Korean README and link it from existing README language switchers
- Include Korean OpenAI setup guidance in the new README

## Test Plan
- `uv run python -m py_compile main.py cli.py webui/Main.py`
- `uv run pytest test/services/test_webui_i18n.py -q`
- Verified `ko` locale loads through `utils.load_locales` and WebUI responds with HTTP 200


harry0703:
Thank you for contributing Korean localization support. Korean is a valuable language for the project, but this PR is now behind the current WebUI: the locale is missing many labels added on the main branch, and the Korean README is also based on an older version of the documentation. Merging it as-is would leave users with an incomplete and inconsistent interface.

We plan to add and maintain additional languages against the current localization baseline with complete coverage and review, so we will close this PR for now. We sincerely appreciate the work and the direction you proposed.

PR #1062, opened 2026-06-21
feat: Add support for Codex / Coze LLM provider

Rishi-rsg07:
This PR adds support for the Codex / Coze LLM platform. It introduces configuration parsing for codex_api_key, codex_model_name, and codex_base_url, and implements direct API integration to handle Coze’s custom v3 chat routing endpoints without breaking the standard OpenAI client layout.

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -184,14 +184,43 @@ def _generate_response(prompt: str) -> str:
                 base_url = config.app.get("ollama_base_url", "")
                 if not base_url:
                     base_url = config.get_default_ollama_base_url()
-            elif llm_provider == "openai":
-                api_key = config.app.get("openai_api_key")
-                model_name = config.app.get("openai_model_name")
-                base_url = config.app.get("openai_base_url", "")
+            elif llm_provider == "codex":
+                api_key = config.app.get("codex_api_key") or ""
+                bot_id = config.app.get("codex_model_name") or ""
+                base_url = config.app.get("codex_base_url", "").rstrip("/")
                 if not base_url:
-                    base_url = "https://api.openai.com/v1"
+                    base_url = "https://api.coze.com/v3/chat"
+
+                headers = {
+                    "Authorization": f"Bearer {api_key}",
+                    "Content-Type": "application/json"
+                }
+                payload = {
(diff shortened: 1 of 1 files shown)


harry0703:
Thanks for the contribution.

I am not going to merge this in the current form yet, mainly because the provider boundary is unclear and the implementation is incomplete for users.

The biggest issue is naming: this PR adds a provider named `codex`, but the implementation is actually calling the Coze v3 chat API. That can easily confuse users, especially because “Codex” may be understood as OpenAI Codex rather than Coze. If this is intended to support Coze, the provider should be named `coze`, not `codex`.

There are also a few functional gaps that need to be addressed before this can be merged:

- Add the provider to the WebUI provider list, otherwise most users cannot select or configure it.
- Add `config.example.toml` entries and clear usage instructions.
- Follow the official Coze v3 chat response flow. The current implementation assumes the final answer is directly available in `data.messages`, which may not be correct for Coze v3 depending on the API mode.
- Avoid hardcoded wording such as `Invalid Mock API Key` in production error messages.
- Keep error handling consistent with the existing LLM providers.

Also, for users who only need broader LLM provider support, the project already includes LiteLLM, which is currently the preferred path for many non-native providers.

I appreciate the contribution, but this needs a more complete and clearly named implementation before it can be accepted.

Rishi-rsg07:
Thank You so much for the feedback will be getting to work with reference
to your valuable feedback.

harry0703:
Thank you again for the contribution and for acknowledging the earlier review feedback.

Since the requested changes have not been implemented, we are going to close this PR. The current patch still identifies the Coze integration as `codex`, does not provide WebUI or example configuration support, and does not follow the required Coze v3 non-streaming flow of polling the chat until completion and then retrieving its messages. It also replaces the existing OpenAI configuration branch, which causes the OpenAI provider to fail at runtime.

The direct Coze Bot use case is currently quite limited for this project compared with the maintenance cost of a dedicated provider. If stronger user demand appears in the future, it would be better to revisit it as a clean, independently tested `coze` adapter built around the official API behavior.

We appreciate the time you spent exploring this integration.

PR #1071, opened 2026-06-23
Add ElevenLab to to Audio setting

nhatanh2996:


harry0703:
Thank you for the contribution. ElevenLabs TTS support is a useful direction for the project, and I appreciate the amount of work you put into this PR.

I tested the branch locally. The basic checks and voice service tests are passing:

```text
uv run python -X utf8 -m unittest test.services.test_voice
```

The WebUI i18n test is still failing because two new translation keys are used in `webui/Main.py` but are not present in all required locale files:

```text
ElevenLabs API Key
ElevenLabs Model
```

Please add these keys to the relevant i18n JSON files so this passes:

```text
```

One more request: please remove the `docs/superpowers/...` design/implementation plan files from this PR. They look like internal implementation notes and are not necessary for users of this project. Keeping the PR focused on the actual ElevenLabs code, config, WebUI, and tests will make it easier to review and merge.

If you have also tested this with a real ElevenLabs API key, it would be helpful to add a short note with the result, without exposing the key.

Thanks again. Once the i18n issue and extra docs are cleaned up, I can take another look.

nhatanh2996:
hi @harry0703, thanks for the feedback — I've updated the coding based on your suggestions, and I also threw in 2 more fonts that support Vietnamese natively, both free from Google Fonts. Good luck. I'm waiting on that TikTok video source fetcher feature of yours haha

harry0703:
Thank you for the update. I tested this PR again with a real ElevenLabs API key.

What works now:

- The API key can be validated successfully.
- The WebUI voice loading flow can fetch favorite ElevenLabs voices.
- `get_elevenlabs_voices()` returns the expected voice list.
- Real TTS generation works with an available voice. I successfully generated an MP3 with `eleven_multilingual_v2`.
- The local checks are passing:

```text
uv run python -X utf8 -m unittest test.services.test_voice
```

There is one remaining issue I found during real testing:

ElevenLabs may still return voices that are no longer usable. For example, one favorite voice was returned by the voices API, but the TTS request failed with `voice_disabled` / `voice_access_denied` because the voice had been disabled by its owner.

Could you please improve this edge case? A simple solution is fine, for example:

- filter unavailable/disabled voices if the voices API exposes a usable status field; or
- handle this error clearly during synthesis so users know they should choose another ElevenLabs voice; and
- avoid silently retrying the same disabled voice three times when the provider has already returned a non-retryable authorization/access error.

Once this edge case is handled and the tests still pass, I think this PR should be ready to merge.

nhatanh2996:
@harry0703 I've fixed it, go ahead and check again. But there's one thing I'm curious about — are you overthinking it, or is that just how production-minded people think? I'm usually pretty detail-oriented myself, but meeting you made me realize there's always a mountain higher than the one you're standing on, haha

harry0703:
Thank you for the quick update and for handling the disabled/inaccessible voice edge case.

I tested the latest version again with a real ElevenLabs API key:

- favorite voices can be loaded from the WebUI flow
- an unavailable voice now fails fast with a clear error instead of retrying repeatedly
- an available ElevenLabs voice can generate a valid MP3 successfully
- local checks are passing:

```text
uv run python -X utf8 -m unittest test.services.test_voice
```

I have merged this PR. Thanks again for the contribution.

PR #1074, opened 2026-06-24
Add missing Vietnamese translations for script-order material matching

sirnvad:
## Summary

- Sync `webui/i18n/vi.json` with `en.json` by adding 2 keys missing from the Vietnamese translation
- `"Match Materials to Script Order"` → `"Khớp tư liệu theo thứ tự kịch bản"`
- `"Match Materials to Script Order Help"` → full Vietnamese description of the feature

## Background

These keys were introduced in commit f2e6d80 but were not included in `vi.json`, causing the Vietnamese UI to fall back to the raw key string for those options.

## Test plan

- [ ] Launch WebUI with language set to Tiếng Việt
- [ ] Verify "Match materials to script order" toggle displays correctly in Vietnamese under Advanced Video Settings

### DIFF
--- a/webui/i18n/vi.json
+++ b/webui/i18n/vi.json
@@ -43,6 +43,8 @@
     "Landscape": "Ngang 16:9",
     "Clip Duration": "Thời Lượng Tối Đa Của Đoạn Video (giây)",
     "Number of Videos Generated Simultaneously": "Số Video Được Tạo Ra Đồng Thời",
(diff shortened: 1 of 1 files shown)


harry0703:
Thank you for adding these missing Vietnamese translations. The two strings have since been added independently to the current main branch as part of the updated WebUI localization, so this PR is now redundant. We appreciate your contribution and will close the PR.

PR #1077, opened 2026-06-26
feat(tts): add Chatterbox TTS provider (self-hosted, OpenAI-compatible)

pcbntt:
Chatterbox (resemble-ai/chatterbox, MIT) is an open-source, locally hosted TTS model with zero-shot voice cloning — a self-hostable alternative to ElevenLabs. This adds it as a selectable TTS server, mirroring the existing ElevenLabs/SiliconFlow integrations.

- voice.py: is_chatterbox_voice(), get_chatterbox_voices(), chatterbox_tts() calling an OpenAI-compatible {base_url}/audio/speech endpoint, plus a dispatch branch in tts() for the "chatterbox:<voice>" voice_name format.
- config.py: register the [chatterbox] config table (load + save).
- config.example.toml: documented [chatterbox] section (base_url/api_key/ model_id/voices).
- webui/Main.py: "Chatterbox TTS" option in the TTS server dropdown, its voice list, and friendly-name handling.

Like ElevenLabs, Chatterbox returns no word-level timestamps, so subtitles use the existing full-text SubMaker fallback (subtitle_provider = "whisper" for tighter sync). Works with community servers such as devnen/Chatterbox-TTS-Server and travisvn/chatterbox-tts-api.


harry0703:
Thank you for the contribution. Adding Chatterbox as a self-hosted TTS provider is a useful direction, especially for users who want a local or self-managed alternative to hosted TTS services.

I checked the PR locally. The basic checks are passing:

```text
uv run python -X utf8 -m unittest test.services.test_voice
```

Before merging, I think a few things should be improved so the feature is usable and safer for regular WebUI users:

1. WebUI configuration is not complete yet

The PR adds `Chatterbox TTS` to the TTS server dropdown, but the WebUI does not expose the required Chatterbox settings yet:

```text
base_url
api_key
model_id
voices
```

Right now a user can select Chatterbox in the WebUI, but unless they manually edit `config.toml`, synthesis will fail because `base_url` is empty. It would be better to add a small Chatterbox settings section in the WebUI, similar to the other TTS providers.

2. Please add focused tests for the new provider

The existing test suite passes, but this PR does not add Chatterbox-specific tests. Please add basic tests for:

```text
is_chatterbox_voice()
get_chatterbox_voices()
chatterbox_tts() success path with mocked requests.post
chatterbox_tts() missing base_url / HTTP failure path
```

3. Real service validation would be helpful

Since this depends on a self-hosted OpenAI-compatible `/audio/speech` endpoint, please also mention which Chatterbox server implementation you tested with and the result, for example `devnen/Chatterbox-TTS-Server` or `travisvn/chatterbox-tts-api`.

One smaller note: `voice_volume` is passed into `chatterbox_tts()` but is not used in the request payload. If the backend does not support volume control, that is acceptable, but it should be clear in the implementation or UI behavior so users do not expect the volume setting to take effect for this provider.

Thanks again. Once the WebUI settings and provider-specific tests are added, I can take another look.

pcbntt:
Thanks for the detailed review — all four points are addressed in the latest commit.

**1. WebUI configuration.** Added a Chatterbox settings section that appears when Chatterbox is selected (or the current voice is a `chatterbox:` voice), exposing **Base URL, API Key, Model, and Voices**. Users no longer need to hand-edit `config.toml`. The four new labels are added to the `en` and `ru` locales so `test_webui_i18n` stays green.

**2. Provider tests.** Added focused tests in `test/services/test_voice.py`:
- `is_chatterbox_voice()` and `get_chatterbox_voices()` (normalisation to `chatterbox:<name>`, comma-string input, empty-config default)
- `chatterbox_tts()` success path with mocked `requests.post` + `AudioFileClip` (asserts URL, payload, bearer auth, written audio, legacy SubMaker)
- `chatterbox_tts()` missing `base_url` (no network call) and HTTP-error (retried 3x -> `None`)

**3. Real service validation.** Tested against **[`travisvn/chatterbox-tts-api`](https://github.com/travisvn/chatterbox-tts-api)** (CPU, OpenAI-compatible `/v1/audio/speech`). End-to-end through the MoneyPrinterTurbo API:

```
POST /api/v1/audio   voice_name="chatterbox:default-Female"   ->  task state=1 (complete)
app/services/voice.py chatterbox_tts - start chatterbox tts, voice: default, try: 1
app/services/voice.py chatterbox_tts - SUCCESS: .../audio.mp3   (4.62s, mono 24kHz)
```

One note for awareness: this server returns WAV content even when `response_format=mp3` is requested (the `.mp3` file contains RIFF/WAVE). moviepy/ffmpeg reads it fine so the pipeline works end-to-end, but it is a minor mismatch worth being aware of.

**4. `voice_volume`.** It is intentionally not sent — the OpenAI `/audio/speech` contract has no volume field, so self-hosted Chatterbox servers ignore it. This is now documented both in `chatterbox_tts()` and in the WebUI info panel, which points users to Speech Rate instead.

Local checks all pass:
- `python -m compileall` on the changed files
- `python -m unittest test.services.test_voice` -> 39 passed (includes the 4 new Chatterbox tests)
- `python -m unittest test.services.test_webui_i18n` -> 4 passed

Happy to adjust naming or wording if you'd like anything changed.

harry0703:
Thank you for the updates. I re-tested the latest PR with a local real Chatterbox environment before merging.

Validation performed:

```text
uv run python -X utf8 -m unittest test.services.test_voice test.services.test_webui_i18n
```

Result: 43 tests passed, 3 skipped.

I also started a local `travisvn/chatterbox-tts-api` CPU server and verified both layers:

```text
GET /health -> status=healthy, model_loaded=true
POST /v1/audio/speech -> 200 OK, audio/wav, about 5.14s
MoneyPrinterTurbo chatterbox_tts() -> success, SubMaker returned, about 3.98s audio generated
```

I checked the WebUI as well. After selecting `Chatterbox TTS`, the Base URL, API Key, Model, and Voices settings are shown correctly.

The only note is the same one you mentioned: the tested server returns WAV content even when `response_format=mp3` is requested. The current pipeline can read it successfully through ffmpeg/moviepy, so I do not think this blocks the PR.

Merging this now. Thanks again for the contribution.

PR #1046, opened 2026-06-16
feat(upload): add YouTube Shorts support via Upload-Post API

lsopenad:
Body:
  ## Summary
  
  - Extends `UploadPostService` to accept `youtube_extra` kwargs and inject YouTube-specific fields into the Upload-Post API payload (`youtube_title`, `youtube_description`,
  `tags[]`, `privacyStatus`, `containsSyntheticMedia=true`)
  - In the task pipeline, when `"youtube"` is present in `upload_post_platforms`, calls `generate_social_metadata(platform="youtube_shorts")` to auto-generate
  title/description/tags before cross-posting 
  - Adds `upload_post_youtube_privacy_status` config key (`"public"` by default) for per-platform privacy control without breaking TikTok/Instagram
  - `containsSyntheticMedia` is hardcoded to `true` — all videos generated by this tool are AI-generated content
  
  ## Changes
  
  | File | Change |
  |---|---|
  | `app/services/upload_post.py` | `upload_video()` + `cross_post_video()` accept `youtube_extra: dict = None`; YouTube fields injected only when `"youtube" in platforms` |
  | `app/services/task.py` | Generates YouTube social metadata before cross-post when YouTube is in platforms |
  | `config.example.toml` | New `upload_post_youtube_privacy_status` key; platforms comment updated |
  | `test/services/test_upload_post.py` | 4 unit tests covering YouTube payload fields, `containsSyntheticMedia` enforcement, and TikTok/Instagram isolation |
  | `README.md` | Cross-platform publishing feature added to feature list |
  
  ## Test plan
  
  - [✅] `uv run python -X utf8 -m unittest test.services.test_upload_post` — 4 tests pass
  - [✅] Existing CI suite passes: `test_state`, `test_task`, `test_schema`, `test_webui_i18n`
  - [✅] Set `upload_post_platforms = ["tiktok", "instagram", "youtube"]` + valid API key → verify YouTube upload on generated video
  - [✅] Set `upload_post_platforms = ["tiktok", "instagram"]` → verify no YouTube fields sent, no regression

PR #1078, opened 2026-06-26
fix(upload): youtube_shorts platform name not matched — metadata silently dropped on upload

kapil971390:
## What's the bug

The YouTube Shorts upload feature (added in #1046) builds `youtube_extra` — containing title, description, tags, and privacy status — only when this check passes:

```python
# app/services/task.py  line 429
if "youtube" in platforms:
```

But the config example documents `"youtube_shorts"` as the platform value:

```toml
# config.example.toml  line 392
# Platforms to cross-post to (options: "tiktok", "instagram", "youtube")
upload_post_platforms = ["tiktok", "instagram"]
```

A user enabling YouTube Shorts sets `upload_post_platforms = ["youtube_shorts"]`. Python list membership `"youtube" in ["youtube_shorts"]` is **`False`**, so `youtube_extra` stays `None` and the video uploads without any metadata — title, description, tags, and privacyStatus are all silently dropped.

The same broken check exists in `upload_post.py` line 62 where the extra fields are appended to the multipart request.

## Root cause

```python
# task.py — youtube_extra never built when platform is "youtube_shorts"
if "youtube" in platforms:          # False for ["youtube_shorts"]
    youtube_extra = { ... }         # never reached

# upload_post.py — extra fields never appended
if youtube_extra and "youtube" in platforms:   # same False
    data.append(('youtube_title', ...))        # never reached
```

## Fix

Replace both exact-match checks with a prefix check so that both `"youtube"` and `"youtube_shorts"` (and any future `youtube_*` variant) are handled:

```python
# task.py
if any(p.startswith("youtube") for p in platforms):

# upload_post.py
if youtube_extra and any(p.startswith("youtube") for p in platforms):
```

## Edge cases verified

**Manually traced all platform list combinations against both the old and new check:**

| `platforms` value | old `"youtube" in platforms` | new `any(p.startswith(...))` |
|---|---|---|
| `["youtube"]` | `True` ✅ | `True` ✅ |
| `["youtube_shorts"]` | **`False` ❌ bug** | `True` ✅ |
| `["tiktok", "youtube_shorts"]` | **`False` ❌ bug** | `True` ✅ |
| `["tiktok", "youtube"]` | `True` ✅ | `True` ✅ |
| `["instagram"]` | `False` ✅ | `False` ✅ |
| `[]` | `False` ✅ | `False` ✅ |

Non-YouTube platforms (`tiktok`, `instagram`) are completely unaffected — the check only gates the `youtube_extra` metadata block, not the upload itself.

### DIFF
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -426,7 +426,7 @@ def start(task_id, params: VideoParams, stop_at: str = "video"):
         logger.info(f"\n\n## cross-posting videos to {', '.join(platforms)}")
 
         youtube_extra = None
-        if "youtube" in platforms:
+        if any(p.startswith("youtube") for p in platforms):
             metadata = llm.generate_social_metadata(
                 video_subject=params.video_subject,
                 video_script=video_script,
--- a/app/services/upload_post.py
+++ b/app/services/upload_post.py
@@ -59,7 +59,7 @@ class UploadPostService:
                 for platform in platforms:
                     data.append(('platform[]', platform))
 
-                if youtube_extra and "youtube" in platforms:
+                if youtube_extra and any(p.startswith("youtube") for p in platforms):
                     if "youtube_title" in youtube_extra:
                         data.append(('youtube_title', youtube_extra["youtube_title"][:100]))
                     if "youtube_description" in youtube_extra:


kapil971390:
### Reproduction screenshots

**Before fix** — `platforms = ["youtube_shorts"]`, check evaluates to `False`, `youtube_extra` stays `None`:
mpt-youtube-BEFORE


**After fix** — same config, `any(p.startswith("youtube") ...)` evaluates to `True`, metadata built correctly:

mpt-youtube-AFTER

harry0703:
Thank you for the clear fix and the detailed explanation.

I reviewed and tested the change locally. The issue makes sense: `youtube_shorts` should also receive the YouTube metadata fields, and the previous exact `"youtube" in platforms` check could silently skip them.

The relevant tests passed on my side:

```text
uv run python -X utf8 -m unittest test.services.test_upload_post test.services.test_llm
```

Merged. Thanks again for the contribution.

kapil971390:
thank you! @harry0703

PR #1084, opened 2026-06-27
feat(material): add optional TwelveLabs video-AI integration (Marengo + Pegasus)

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## 中文简介

为 MoneyPrinterTurbo 新增**可选**的 TwelveLabs 视频 AI 接入（完全 opt-in，不配置 key 时行为零变化）：

- **Marengo**（多模态向量，512 维）：在下载素材前，按视频主题对 LLM 生成的搜索关键词做语义重排，让最贴合主题的素材优先被下载——当时间轴预算紧张时尤其有用。
- **Pegasus**（视频理解）：提供一个可复用的 `analyze_clip()` 服务，可对素材片段做内容 QA / 描述（作为辅助函数提供，未强制接入任何流程，因此不会改变默认行为）。

接入方式与现有素材源（Pexels / Pixabay / Coverr）一致:在 `config.toml` 的 `[app]` 段配置 `twelvelabs_api_keys`，并把开关 `twelvelabs_rerank_terms` 设为 `true` 即可启用重排。SDK 是可选额外依赖（`uv sync --extra twelvelabs`），且仅在真正使用时才惰性导入。未配置 key 时所有函数都是无副作用的 no-op；“素材按文案顺序匹配”模式下会跳过重排（该模式下关键词顺序本身就是脚本叙事顺序）。

可在 https://twelvelabs.io 免费申请 API key（有较慷慨的免费额度）。

## English

This adds an **opt-in** TwelveLabs (https://twelvelabs.io) video-AI integration. With no `twelvelabs_api_keys` configured, behavior is completely unchanged.

**What it adds**
- **Marengo** (multimodal embeddings, 512-dim) — semantically reorders the auto-generated B-roll search terms by relevance to the `video_subject` *before* footage is downloaded, so the most on-topic material lands first when the timeline budget is tight (instead of whatever order the LLM happened to emit).
- **Pegasus** (video understanding) — a reusable `analyze_clip()` service that QA's / describes a clip from a public URL. It is exposed as a helper for callers (e.g. sanity-checking a downloaded clip against the script); it is intentionally **not** auto-invoked in the pipeline, so it can't change default behavior. Happy to wire it into an explicit QA step as a follow-up if that's of interest.

**Why it helps**
The pipeline already auto-generates search terms via the LLM, but their order is arbitrary. When the video runs out of timeline before all terms are used, the *first* terms win — so ordering them by semantic relevance to the subject directly improves which B-roll makes the final cut. This slots in right where terms are generated (`generate_terms` in `app/services/task.py`).

**Opt-in / non-breaking**
- Wired in exactly like the existing material providers (Pexels/Pixabay/Coverr): config lives in the `[app]` section of `config.toml`, keys are rotated via the same `material.get_api_key` helper.
- Gated twice: needs `twelvelabs_api_keys` **and** `twelvelabs_rerank_terms = true`. Any API error falls back to the original term order — it can never make results worse.
- Skipped in "match materials to script" mode, where term order is intentionally the script's narrative order.
- The `twelvelabs` SDK is an **optional** extra (`uv sync --extra twelvelabs`) and imported lazily, so the base install and existing CI (`uv sync --frozen`) are unaffected.

**How tested**
- New `test/services/test_twelvelabs.py` (stdlib `unittest`, mirroring the repo's other service tests):
  - No-network unit tests with the SDK mocked: disabled-no-op, flag-off skip, cosine ordering, fallback-on-failure, single-term no-op, `analyze_clip` text + `max_tokens` clamping.
  - A live contract test gated on `MPT_RUN_INTEGRATION_TESTS=1` + `TWELVELABS_API_KEY` (skipped by default, matching the repo's existing live-test convention).
- Ran the repo's exact CI steps locally: `python -m compileall app cli.py main.py webui test` and the deterministic unit suite (incl. `test_task`) under `uv sync --frozen` **without** the extra — all green, confirming the lazy import keeps CI unaffected.
- Live-verified the TwelveLabs contract against the real API: Marengo returns a 512-dim vector and the rerank puts the on-topic term first.
- `uv lock` updated so `uv sync --frozen` stays consistent with the new optional extra.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/README-en.md
+++ b/README-en.md
@@ -104,7 +104,7 @@ materials, video subtitles, and video background music before synthesizing a hig
 - [x] Supports **background music**, either random or specified music files, with adjustable `background music volume`
 - [x] Video material sources are **high-definition** and **royalty-free**, and you can also use your own **local materials**
 - [x] Supports multiple stock video providers: **Pexels**, **Pixabay**, and **Coverr**
-- [x] Optional **TwelveLabs** video AI integration: use **Marengo** embeddings to semantically rank B-roll search terms against your subject, and **Pegasus** to QA/describe clips ([free API key](https://twelvelabs.io))
+- [x] Optional **TwelveLabs** video AI integration: use **Marengo** embeddings to semantically rank B-roll search terms against your subject, and **Pegasus** to QA/describe clips ([TwelveLabs API key](https://twelvelabs.io))
 - [x] Supports integration with various models such as **OpenAI**, **AIHubMix**, **AIML API**, **EvoLink**, **Moonshot**, **Azure**, **gpt4free**, **one-api**, **Qwen**, **Google Gemini**, **Ollama**, **DeepSeek**, **MiniMax**, **ERNIE**, **Pollinations**, **ModelScope** and more
 
 ## Video Demos 📺
--- a/README-en.md
+++ b/README-en.md
@@ -104,6 +104,7 @@ materials, video subtitles, and video background music before synthesizing a hig
 - [x] Supports **background music**, either random or specified music files, with adjustable `background music volume`
 - [x] Video material sources are **high-definition** and **royalty-free**, and you can also use your own **local materials**
 - [x] Supports multiple stock video providers: **Pexels**, **Pixabay**, and **Coverr**
+- [x] Optional **TwelveLabs** video AI integration: use **Marengo** embeddings to semantically rank B-roll search terms against your subject, and **Pegasus** to QA/describe clips ([free API key](https://twelvelabs.io))
 - [x] Supports integration with various models such as **OpenAI**, **AIHubMix**, **AIML API**, **EvoLink**, **Moonshot**, **Azure**, **gpt4free**, **one-api**, **Qwen**, **Google Gemini**, **Ollama**, **DeepSeek**, **MiniMax**, **ERNIE**, **Pollinations**, **ModelScope** and more
 
 ## Video Demos 📺
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@
 - [x] 支持 **背景音乐**，随机或者指定音乐文件，可设置`背景音乐音量`
 - [x] 视频素材来源 **高清**，而且 **无版权**，也可以使用自己的 **本地素材**
 - [x] 支持多种素材源:**Pexels**、**Pixabay**、**Coverr**
-- [x] 可选接入 **TwelveLabs** 视频 AI:用 **Marengo** 多模态向量按主题语义重排素材关键词,用 **Pegasus** 对素材做内容 QA/描述([免费 API key](https://twelvelabs.io))
+- [x] 可选接入 **TwelveLabs** 视频 AI:用 **Marengo** 多模态向量按主题语义重排素材关键词,用 **Pegasus** 对素材做内容 QA/描述([TwelveLabs API key](https://twelvelabs.io))
 - [x] 支持 **OpenAI**、**AIHubMix**、**AIML API**、**EvoLink**、**Moonshot**、**Azure**、**gpt4free**、**one-api**、**通义千问**、**Google Gemini**、**Ollama**、**DeepSeek**、**MiniMax**、 **文心一言**, **Pollinations**、**ModelScope** 等多种模型接入
 - [x] 支持一键 **跨平台发布**，生成完成后自动上传至 **TikTok**、**Instagram** 和 **YouTube Shorts**（需 [Upload-Post](https://upload-post.com) 账号；YouTube 发布时自动标注 AI 生成内容）；在 `config.toml` 中配置 `upload_post_platforms`、`upload_post_youtube_privacy_status` 等参数
 
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@
 - [x] 支持 **背景音乐**，随机或者指定音乐文件，可设置`背景音乐音量`
 - [x] 视频素材来源 **高清**，而且 **无版权**，也可以使用自己的 **本地素材**
 - [x] 支持多种素材源:**Pexels**、**Pixabay**、**Coverr**
+- [x] 可选接入 **TwelveLabs** 视频 AI:用 **Marengo** 多模态向量按主题语义重排素材关键词,用 **Pegasus** 对素材做内容 QA/描述([免费 API key](https://twelvelabs.io))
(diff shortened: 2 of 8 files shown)


harry0703:
Thank you for the contribution. The direction is interesting, especially the opt-in Marengo reranking before material download. That could improve which B-roll terms are prioritized when the video duration budget is tight.

I tested the PR locally and found one blocker before we can merge it.

Default environment test failure:

```text
uv run python -X utf8 -m unittest test.services.test_twelvelabs
```

This fails in the default environment without the optional extra installed:

```text
FAIL: test_analyze_clip_returns_model_text
TwelveLabs analyze_clip failed: No module named 'twelvelabs'
```

The root cause appears to be that `analyze_clip()` imports `twelvelabs.types.VideoContext_Url` directly inside the function. Even though `_client()` is mocked in the unit test, the test still requires the optional `twelvelabs` SDK to be installed. Since the PR intends the SDK to be optional and lazily imported, the deterministic unit tests should also pass without `uv sync --extra twelvelabs`.

For comparison, this does pass when the optional extra is installed:

```text
uv run --extra twelvelabs python -X utf8 -m unittest test.services.test_twelvelabs
```

Other checks I ran:

```text
uv run python -X utf8 -m unittest test.services.test_task
```

Those passed.

Could you please adjust the test or the helper so the default test suite remains green without installing the optional TwelveLabs SDK? For example, the `VideoContext_Url` construction could be wrapped behind a small helper that is mockable, or the unit test could patch the module import path explicitly.

One smaller request: please also make the README/config wording more neutral. Phrases like “free API key” and “generous free tier” read a bit promotional for the main project documentation. It would be better to keep it factual, for example: “Configure a TwelveLabs API key from the TwelveLabs dashboard if you want to enable this optional integration.”

Once the default-environment test failure is fixed and the documentation wording is toned down, I can take another look. Thanks again.

mohit-twelvelabs:
感谢 @harry0703 的 review！已修复合并前的唯一阻塞项。

**问题根因**：`analyze_clip()` 内部惰性导入 `from twelvelabs.types import VideoContext_Url`，因此即使 mock 了 `_client`，确定性单测在默认环境（未执行 `uv sync --extra twelvelabs`）下仍需真实 SDK，导致 `test_analyze_clip_returns_model_text` 报 `No module named 'twelvelabs'`。

**修复**：在该单测中通过 `sys.modules` 注入轻量级的 `twelvelabs` / `twelvelabs.types` stub（提供一个最小的 `VideoContext_Url`），让内部导入解析到 stub，其余逻辑仍由被 mock 的 `_client` 处理。这样确定性单测无需真实依赖即可运行。运行时导入保持惰性，SDK 仍为可选 extra，**未**添加任何硬依赖。Marengo 重排序相关用例本就完全不依赖 SDK，保持不变。

验证（两种环境均通过）：
- 默认环境（SDK 未安装，已确认 `find_spec('twelvelabs') is None`）：`uv run python -X utf8 -m unittest test.services.test_twelvelabs` → OK (skipped=2)
- 带 extra：`uv run --extra twelvelabs python -X utf8 -m unittest test.services.test_twelvelabs` → OK (skipped=2)

`uv.lock` 未改动，其他确定性单测无回归。提交：`b8cd49b`。

---

Thanks for the review, @harry0703! Fixed the one merge blocker.

**Root cause**: `analyze_clip()` lazily does `from twelvelabs.types import VideoContext_Url` inside the function, so even with `_client` mocked the deterministic unit test still needed the real SDK in the default environment (no `uv sync --extra twelvelabs`) — hence `No module named 'twelvelabs'` in `test_analyze_clip_returns_model_text`.

**Fix**: the unit test now injects lightweight stub modules for `twelvelabs` / `twelvelabs.types` (a minimal `VideoContext_Url`) via `sys.modules`, so the internal import resolves to the stub and the mocked `_client` handles the rest. The test runs deterministically with no real dependency. The runtime import stays lazy and the SDK stays an optional extra — **no** hard dependency added. The Marengo-rerank tests were already fully SDK-free and are unchanged.

Verified (both pass):
- Default env (SDK not installed; confirmed `find_spec('twelvelabs') is None`): `uv run python -X utf8 -m unittest test.services.test_twelvelabs` → OK (skipped=2)
- With extra: `uv run --extra twelvelabs python -X utf8 -m unittest test.services.test_twelvelabs` → OK (skipped=2)

`uv.lock` unchanged, no regressions in other deterministic tests. Commit: `b8cd49b`.

— Mohit (@mohit-twelvelabs, TwelveLabs)

harry0703:
Thanks for the quick update. I re-tested the latest commit locally.

The previous test blocker is fixed now:

```text
uv run python -X utf8 -m unittest test.services.test_twelvelabs
uv run --extra twelvelabs python -X utf8 -m unittest test.services.test_twelvelabs
uv run python -X utf8 -m unittest test.services.test_task
```

All of the above passed on my side.

There is still one remaining documentation issue before merge: the marketing-style wording has not been fully toned down yet. I still see phrases such as:

```text
README-en.md: "free API key"
README.md: "免费 API key"
config.example.toml: "Get a free key (generous free tier)"
app/services/twelvelabs.py: "Get a free API key ... (generous free tier)"
```

Could you please make those references neutral? For example, something like:

```text
Configure a TwelveLabs API key from the TwelveLabs dashboard to enable this optional integration.
```

or simply link to the TwelveLabs dashboard/documentation without mentioning free tiers or promotional wording.

After that, I think this should be ready for another final pass. Thanks.

mohit-twelvelabs:
感谢 review！已将 4 处（README.md / README-en.md / config.example.toml / app/services/twelvelabs.py）的"免费 / generous free tier"等推广性措辞改为中性表述——直接指向 TwelveLabs 控制台，不再提及免费层级（commit `b1d4732`）。

Done — toned down all four references (README.md, README-en.md, config.example.toml, app/services/twelvelabs.py) to neutral wording that just points to the TwelveLabs dashboard, with no free-tier/promotional phrasing (commit `b1d4732`). Tests still pass as before. Thanks for the careful review — ready for your final pass whenever you are.

— Mohit (@mohit-twelvelabs, TwelveLabs)

harry0703:
Merged. Thank you for the contribution and for iterating on the review feedback.

I rechecked the updated PR before merging: the default-environment tests now pass without the optional TwelveLabs SDK, the extra-enabled tests pass as well, and the documentation wording is now neutral. I also verified the live TwelveLabs path with a private key, including Marengo embedding/reranking and Pegasus clip analysis.

This is a useful opt-in addition for users who want better B-roll term ordering without changing the default behavior for everyone else. Thanks again.

PR #1085, opened 2026-06-27
feat: import .txt/.md file into the WebUI video script field

### LINKED ISSUE #1002: [Feature]: 能否新增根据md/doc生成视频？
### 需求描述 | Problem Statement

我有现成的文案，不是需要根据关键词来生成视频。

有更详细的内容+模型能力应该能生成更好的视频吧。

mvanhorn:
## Summary
Add a Streamlit `st.file_uploader` accepting `txt` and `md` directly above the existing Video Script `text_area` in `webui/Main.py` (the `tr("Video Script")` widget). When a file is uploaded, decode it as UTF-8 and run it through a new `strip_markdown(text: str) -> str` helper in `app/utils/utils.py` that removes Markdown headings, list bullets/numbering, horizontal-rule separators, emphasis markers, and link/image syntax while preserving the prose and line structure. Write the cleaned text into `st.session_state["video_script"]` so it populates the script field.

## Why this matters
The reporter has finished scripts and wants to feed an existing document into MoneyPrinterTurbo rather than generating a script from keywords. The maintainer (harry0703) confirmed pasting into the Video Script box already works, but there is no file-upload entry point for documents. He explicitly endorsed adding this as an enhancement and sketched the design: upload `.txt`/`.md`, auto-read the body into the video script, auto-clean Markdown formatting (headings, lists, separators), with longer-document segmentation/summarization and `.docx` as later steps. This plan implements the first, well-bounded slice of that design.

See https://github.com/harry0703/MoneyPrinterTurbo/issues/1002.

## Testing
- Happy path: a `.md` file with `# Heading`, `- bullet`, `---` separators, and `**bold**` produces clean prose with formatting markers removed and paragraph breaks intact. - A plain `.txt` file passes through `strip_markdown` essentially unchanged (no Markdown tokens to remove). - Edge: empty file or whitespace-only file yields an empty/stripped string without raising. - Edge: links `[text](url)` and images `alt` reduce to their visible text (or are removed for images) rather than leaking raw URLs.

Fixes #1002

### DIFF
--- a/app/utils/utils.py
+++ b/app/utils/utils.py
@@ -243,6 +243,45 @@ def split_string_by_punctuations(s):
     return result
 
 
+def strip_markdown(text: str) -> str:
+    text = text or ""
+    cleaned_lines = []
+    removed_separator = False
+
+    for line in text.splitlines():
+        line = re.sub(r"!\[[^\]]*\]\([^)]+\)", "", line)
+        line = re.sub(r"\[([^\]]+)\]\([^)]+\)", r"\1", line)
+        line = re.sub(r"\[([^\]]+)\]\[[^\]]*\]", r"\1", line)
+
+        stripped_line = line.strip()
+        if any(
+            re.fullmatch(pattern, stripped_line)
+            for pattern in (r"(?:-\s*){3,}", r"(?:\*\s*){3,}", r"(?:_\s*){3,}")
+        ):
+            removed_separator = True
+            continue
+
+        heading_match = re.match(r"^\s{0,3}#{1,6}\s+(.*?)\s*#*\s*$", line)
+        if heading_match:
+            line = heading_match.group(1)
+        else:
+            line = re.sub(r"^\s*(?:[-*+]\s+|\d+[.)]\s+)", "", line)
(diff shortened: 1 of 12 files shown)


harry0703:
Thanks for the PR. I tested this locally and the implementation works as expected, but after reviewing the WebUI flow I do not plan to merge this version for now.

The current WebUI already allows users to paste an existing script directly into the Video Script field. Adding a visible file uploader above that field is a small convenience for `.txt` / `.md` files, but it also makes the main workflow look a bit more complex and may give new users the impression that uploading a file is required.

If we revisit this later, I would prefer to keep it as a lower-priority advanced option, probably behind an expander, and focus on more meaningful document-import support such as longer document handling, segmentation, summarization, or `.docx` support.

Thank you again for the contribution and for implementing it cleanly.

mvanhorn:
Makes sense - the paste flow already covers this, and keeping the main workflow simple is the right call. I'll close this out. If document import gets revisited later (expander placement, .docx, longer docs with segmentation), I'd take another pass at it in that shape.

PR #1086, opened 2026-06-27
fix: warn at startup when the API key is empty on a non-loopback bind

### LINKED ISSUE #1018: Code Review Observations & Questions
## Overview

I did a code review pass across the core source files — `task.py`, `video.py`, `material.py`, `state.py`, `llm.py`, `voice.py`, `schema.py`, `file_security.py`, and the controller layer. Overall the project is well-organized and clearly the result of sustained, thoughtful development. The comments (especially the Chinese-language inline notes) show real care for edge-case reasoning. A few things stood out that I wanted to raise constructively.

---

## Strengths Worth Noting

- **Path traversal protection** in `file_security.resolve_path_within_directory` is solid — uses `os.path.realpath` + `os.path.commonpath` rather than naive string prefix checks, handles Windows drive letters correctly.
- **`_prioritize_unique_source_clips`** is a clever heuristic. The decision to select the longest clip per source before shuffling fallbacks is non-obvious but well-motivated in the inline comment.
- **`_normalize_text_response`** stripping `<think>...</think>` blocks from reasoning models before passing text downstream is the right call and easy to miss.
- **Video codec whitelist** in `_get_configured_video_codec` — limiting to a known-safe set rather than passing arbitrary strings to FFmpeg is good defensive practice.
- **TLS verify escape hatch** in `material.py` with a loud warning log is a reasonable compromise for proxy environments.

---

### 1. `MemoryState` is not thread-safe

`task.py` calls `sm.state.update_task(...)` from background threads via FastAPI `BackgroundTasks`, and `MemoryState._tasks` is a plain `dict` with no lock. Python's GIL protects individual bytecode operations but not compound read-modify-write sequences like `update_task`'s `{..., **kwargs}` merge. Concurrent tasks could silently drop updates.

**Suggestion:** Wrap `_tasks` mutations in a `threading.Lock()`, similar to how `_api_key_lock` is used in `material.py`.

### 2. `RedisState._convert_to_original_type` uses `ast.literal_eval` on untrusted data

All task state fields are stored as strings in Redis and deserialized with `ast.literal_eval`. If Redis is accessible to other processes or users, a crafted string value in a task field could trigger unexpected behavior. `literal_eval` is safe against code execution but it will happily deserialize arbitrary nested data structures of unbounded depth.

**Suggestion:** Consider an explicit allow-list of types (int, float, str, list of strings) rather than open-ended `literal_eval`. At minimum, document the trust boundary assumption.

### 3. `start()` in `task.py` silently swallows the return value on most error paths

Most failure branches call `sm.state.update_task(..., state=TASK_STATE_FAILED)` and then `return` (bare return, returning `None`). The caller in the controller layer checks `if not final_video_paths` but the earlier stages just fall through. The `stop_at` partial-pipeline feature returns structured dicts on success but `None` on failure, making it hard for callers to distinguish "stopped early successfully" from "failed silently."

**Suggestion:** Either raise a typed exception or return a consistent `{"error": "..."}` dict on failure so callers can surface a meaningful error to the API response.

### 4. Auto cross-post runs unconditionally in the video generation hot path

In `task.py`, the TikTok/Instagram cross-post block runs synchronously inside `start()`, which is already a background task. If the Upload-Post API is slow or times out, it stalls the task completion signal. The `cross_post_results` are stored in task state but there is no retry, timeout cap, or way for the caller to opt out per-request (only a global `auto_upload` flag).

**Suggestion:** Run cross-posting in a separate fire-and-forget thread or make it an explicit API endpoint so users can trigger it selectively after reviewing the generated video.

### 5. `custom_audio_file` path is validated via `os.path.exists` in `task.py` but the path originates from user-supplied `VideoParams`

`generate_audio()` calls `os.path.exists(custom_audio_file)` directly on the user-provided path before using it. Unlike the BGM and task-output paths, this path is not run through `file_security.resolve_path_within_directory`. A user (or API client) with access to the generate endpoint could probe the filesystem for file existence.

**Suggestion:** Validate `custom_audio_file` against the upload directory using the existing `file_security` module before checking existence or opening the file.

### 6. `VideoAspect.to_resolution()` falls through to a default instead of raising

```python
def to_resolution(self):
    if self == VideoAspect.landscape.value:
        return 1920, 1080
    elif self == VideoAspect.portrait.value:
        return 1080, 1920
    elif self == VideoAspect.square.value:
        return 1080, 1080
    return 1080, 1920  # silent default
```

If a new enum value is added (e.g. `ultrawide`) and `to_resolution` is not updated, video generation silently uses portrait resolution. The comparisons also use `.value` on `self` against `.value` — since `VideoAspect` is a `str` enum, `self == VideoAspect.landscape.value` compares a member to its own string value, which works but is confusing.

**Suggestion:** Add an `else: raise ValueError(f"no resolution for {self}")` and simplify comparisons to `self == VideoAspect.landscape`.

### 7. API authentication is commented out

In `app/controllers/v1/video.py`:
```python
# router = new_router(dependencies=[Depends(base.verify_token)])
router = new_router()
```

This means anyone with network access can submit video generation tasks, consume API keys configured in `config.toml`, and enumerate task outputs. The `config.example.toml` mentions an API key setting but the enforcement is disabled.

**Question:** Is this intentional for local/demo use only? If so, it would help to add a prominent note in the README or a startup warning when the server binds to a non-loopback address with no auth enabled.

---

## Minor Items

- `get_api_key()` in `material.py` uses a global `_api_key_counter` for round-robin rotation, but the counter is never reset and grows without bound. With `int` in Python this is harmless but slightly untidy.
- `RedisState.get_all_tasks` does a full `SCAN` to compute `total` even when only page 1 is needed. For large task counts this could be slow. Consider maintaining a separate counter key or using `DBSIZE` as an approximation.
- The `stop_at` parameter in `task.py:start()` uses magic strings (`"script"`, `"terms"`, etc.) with no enum or constant definitions. A `StopAt` enum in `schema.py` would make this easier to document and validate at the API boundary.

---

Thanks for building this — the pipeline composition pattern (script → terms → audio → subtitle → materials → video) is clean and the incremental `stop_at` feature is genuinely useful for debugging and partial workflows.

mvanhorn:
## Summary
Add a small, testable helper in `app/asgi.py`, e.g. `warn_if_api_unprotected(api_key: str, listen_host: str) -> str | None`, that returns a warning message when `api_key` is empty/blank and `listen_host` is not a loopback address (`127.0.0.1`, `localhost`, `::1`), and `None` otherwise. Invoke it once at application construction in `get_application()` (or immediately after `app = get_application()`), logging the returned message via the already-imported `loguru` `logger.warning(...)` so operators see it on startup.

## Why this matters
A reviewer filed a multi-point code-review issue; the maintainer (harry0703) agreed several items were worth tracking and, in a follow-up, confirmed he had already addressed `MemoryState` locking, `VideoAspect.to_resolution()` explicit failure, and `custom_audio_file` path resolution himself. One agreed item remains unaddressed: "documenting or warning more clearly when API authentication is disabled, especially for non-loopback deployments." In `app/controllers/base.py`, `verify_token` compares the request header against `config.app.get("api_key", "")`, and `config.listen_host` defaults to `0.0.0.0` (a non-loopback bind), so a deployment left with an empty `api_key` exposes the API on all interfaces with no clear operator warning.

See https://github.com/harry0703/MoneyPrinterTurbo/issues/1018.

## Testing
- Empty `api_key` with `listen_host="0.0.0.0"` returns a non-empty warning string. - Empty `api_key` with `listen_host="127.0.0.1"` (and `localhost`, `::1`) returns `None` (loopback is considered safe). - Non-empty `api_key` with any host returns `None`. - Whitespace-only `api_key` is treated as empty (still warns on non-loopback).

Fixes #1018

### DIFF
--- a/app/asgi.py
+++ b/app/asgi.py
@@ -15,6 +15,24 @@ from app.router import root_api_router
 from app.utils import utils
 
 
+_LOOPBACK_HOSTS = {"127.0.0.1", "localhost", "::1"}
+
+
+def warn_if_api_unprotected(api_key: str, listen_host: str) -> str | None:
+    if str(api_key or "").strip():
+        return None
+
+    normalized_host = str(listen_host or "").strip().lower()
+    if normalized_host in _LOOPBACK_HOSTS:
+        return None
+
+    return (
+        "API authentication is disabled because app.api_key is empty while "
+        f"listen_host is '{listen_host}'. Set app.api_key or bind to a loopback "
+        "host before exposing this service."
+    )
+
+
 def exception_handler(request: Request, e: HttpException):
     return JSONResponse(
         status_code=e.status_code,
@@ -38,6 +56,12 @@ def get_application() -> FastAPI:
        FastAPI: Application object instance.
 
     """
+    auth_warning = warn_if_api_unprotected(
+        config.app.get("api_key", ""), config.listen_host
+    )
+    if auth_warning:
+        logger.warning(auth_warning)
+
     instance = FastAPI(
         title=config.project_name,
         description=config.project_description,
--- a/test/services/test_asgi_auth_warning.py
+++ b/test/services/test_asgi_auth_warning.py
@@ -1,0 +1,34 @@
+import sys
+import unittest
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent))
+
+from app.asgi import warn_if_api_unprotected
+
+
+class TestAsgiAuthWarning(unittest.TestCase):
+    def test_empty_api_key_warns_for_non_loopback_host(self):
+        warning = warn_if_api_unprotected("", "0.0.0.0")
+
+        self.assertTrue(warning)
+        self.assertIn("API authentication is disabled", warning)
+
+    def test_empty_api_key_allows_loopback_hosts(self):
+        for listen_host in ("127.0.0.1", "localhost", "::1"):
+            with self.subTest(listen_host=listen_host):
+                self.assertIsNone(warn_if_api_unprotected("", listen_host))
+
+    def test_non_empty_api_key_never_warns(self):
+        for listen_host in ("0.0.0.0", "127.0.0.1", "localhost", "::1"):
+            with self.subTest(listen_host=listen_host):
+                self.assertIsNone(warn_if_api_unprotected("secret", listen_host))
+
+    def test_whitespace_only_api_key_warns_for_non_loopback_host(self):
+        warning = warn_if_api_unprotected("  \t\n", "0.0.0.0")
+
+        self.assertTrue(warning)
+
+
+if __name__ == "__main__":
+    unittest.main()


harry0703:
Thank you for following up on the security observation and for keeping the proposed change small and well tested.

At this stage, our intended deployment setup keeps the API on the loopback interface, and the provided Docker configurations publish port 8080 only on `127.0.0.1`. Because of that, this startup warning would provide limited practical value for the deployments we currently support.

I also found that a warning conditioned only on whether `app.api_key` is empty would not accurately represent the current protection state, because the authentication dependencies on the API routers are not enabled. Supporting non-loopback or public deployments safely would require a broader authentication and task-output access-control design rather than only a startup message. We are not planning to take on that larger change at this time.

I am therefore closing this PR instead of asking you to rework it further. I appreciate the careful implementation, tests, and the attention you brought to the deployment risk. Thank you again for the contribution.

PR #1087, opened 2026-06-28
Keep credentials in local place

RyanFeiluX:
Split config.toml into two parties, parameters + secrets.


harry0703:
Thanks for the PR.

I reviewed the scope, but I am going to close this one rather than merge it.

The PR title/body mentions keeping credentials in a local place, which is a reasonable direction. However, the actual diff is much broader than that:

- 300+ commits and more than 130 changed files
- a large Vue frontend migration
- multi-scene video generation changes
- logging/WebSocket/settings API changes
- Docker/script/documentation changes
- project/version/name changes from a forked branch
- a committed `config.toml` containing real or real-looking API keys

Because of that, this PR is too large and risky to review or merge safely. It also includes sensitive configuration data, which should never be committed to the repository.

If you want to contribute the credentials/config separation idea, please open a much smaller PR that only changes that part. Ideally it should:

- keep the current project structure and UI unchanged
- avoid unrelated frontend, video pipeline, Docker, or documentation rewrites
- keep `config.toml` ignored and out of git
- update only `config.example.toml` or add a safe local override mechanism
- include a short explanation of the migration path for existing users

Thanks again for taking the time to contribute. The config/secrets direction is worth discussing, but it needs to be separated into a focused, reviewable PR.

PR #1088, opened 2026-06-28
add CLAUDE.md and fix .gitignore for session context persistence

baohua01:


harry0703:
Thanks for the PR.

I am going to close this one because `CLAUDE.md` is a local AI-assistant/session context file and should not be committed to the public project. The current content also includes local workflow details, branch names, and account-specific information that are not useful for general users.

Keeping `CLAUDE.md` ignored is intentional so contributors can use their own local assistant notes without adding them to the repository. Thanks for understanding.

PR #1090, opened 2026-06-29
新增火山引擎供应商

zengwangdong:
新增火山引擎供应商，并采用推广链接
image

image
image

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -286,6 +286,13 @@ def _generate_response(prompt: str) -> str:
                 base_url = config.app.get("deepseek_base_url")
                 if not base_url:
                     base_url = "https://api.deepseek.com"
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for the contribution.

We have absorbed the core idea from this PR and added VolcEngine Ark as a built-in LLM provider on `main` in a more neutral and minimal form:

- added `volcengine` as an OpenAI-compatible provider
- added WebUI provider selection and configuration hints
- added default Ark base URL and model settings
- added unit coverage for the provider path
- verified the provider with a real Ark API request before merging the local change

I am going to close this PR instead of merging it directly because the final implementation was adjusted to keep the UI copy and provider integration more consistent with the rest of the project. Thanks again for proposing this integration.

PR #1094, opened 2026-07-02
fix: estimate no-voice duration correctly for Cyrillic scripts

### LINKED ISSUE #1093: [Bug]: No Voice mode overestimates video duration for Cyrillic scripts
### 问题描述 | Current Behavior

When No Voice is selected, MoneyPrinterTurbo significantly overestimates the duration of videos whose scripts use Cyrillic characters.

A Bulgarian script containing approximately 68 words generated a silent video approximately 90 seconds long. The same script would normally take approximately 30 seconds to narrate.

The problem occurs because estimate_no_voice_duration() treats Cyrillic text using the fallback character-based calculation of approximately four characters per second, instead of counting whitespace-delimited words.

### 重现步骤 | Steps to Reproduce

1. Start MoneyPrinterTurbo.
2. Enter a Bulgarian script using Cyrillic characters.
3. Select No Voice.
4. Disable subtitles.
5. Generate the video.
6. Observe that a script of approximately 68 words produces a video approximately 90 seconds long.

**Example script:**

"Инфлацията е когато цените на стоките и услугите се покачват с времето. Със същите 100 лева купуваме по-малко храна, гориво или дрехи отпреди. Това означава, че парите ни губят част от стойността си. Ако доходите ни не растат със същия темп, реално ставаме по-бедни. Затова е важно да следим бюджета си, да сравняваме цени и да пазим спестяванията си разумно, например чрез депозит, инвестиции или други подходящи решения." 

**Expected behavior**

The Bulgarian script should produce a silent timeline of approximately 28–32 seconds, based on its word count and sentence pauses.

### 错误日志 | Error Logs

None. Video generation completes successfully, but the generated silent audio and final video are approximately 90 seconds long due to the wrong word calculation.

3.12.1

### 操作系统 | Operating System

Windows 11

v1.3.0

The issue is caused by the fallback logic in estimate_no_voice_duration(). Cyrillic letters are counted as generic non-ASCII characters and divided by four characters per second.

Counting Cyrillic words at approximately 2.4 words per second fixes the Bulgarian example while preserving the existing CJK character-based estimation.

The current main implementation does indeed count CJK characters separately, ASCII text by words, but other scripts—including Cyrillic—by individual characters at four characters per second.

stoyandg:
## Summary

Fixes #1093.

This PR corrects the silent-video duration estimation for Cyrillic scripts in `No Voice` mode.

Previously, Cyrillic text was processed by the generic non-ASCII fallback and estimated at approximately four characters per second. This caused Bulgarian and other Cyrillic scripts to produce videos substantially longer than their expected spoken duration.

For example, a Bulgarian script of approximately 68 words generated a video approximately 90 seconds long instead of approximately 30 seconds.

## Changes

* Detect whitespace-delimited Cyrillic words.
* Estimate Cyrillic duration at approximately 2.4 words per second.
* Exclude Cyrillic characters from the generic character-based fallback.
* Preserve the existing CJK, ASCII and other-script estimation behavior.
* Add regression tests for Bulgarian Cyrillic text.
* Add a regression test to ensure CJK estimation remains unchanged.

## Before

The example Bulgarian script generated a silent timeline of approximately 86–90 seconds.

## After

The same script generates a timeline of approximately 28–32 seconds which is the correct duration.

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -269,39 +269,87 @@ def is_no_voice(voice_name: str | None) -> bool:
 
 def estimate_no_voice_duration(text: str) -> float:
     """
-    为无配音模式估算一个稳定的视频时间轴长度。
+    Estimate a stable video timeline duration for No Voice mode.
 
-    无配音仍需要一个音频占位来驱动现有素材裁剪、字幕时间轴和最终合成。
-    估算策略尽量简单：
-    1. 中文等 CJK 字符按约 4.2 字/秒估算；
-    2. 英文/数字按约 2.7 词/秒估算；
-    3. 其他语种文字按约 4.0 字符/秒兜底估算，覆盖俄语、阿拉伯语、
-       日文假名、韩文等非 ASCII 文本；
-    4. 每个断句补一点停顿，让字幕切换不至于过于紧凑；
-    5. 最少 3 秒，避免极短脚本生成 0 秒音频。
+    Estimation strategy:
+    1. CJK characters are estimated at approximately 4.2 characters/second.
+    2. ASCII words and numbers are estimated at approximately 2.7 words/second.
+    3. Cyrillic words are estimated at approximately 2.4 words/second.
+    4. Other writing systems fall back to approximately 4 characters/second.
+    5. A short pause is added between sentences.
+    6. The minimum generated duration is 3 seconds.
     """
     normalized_text = (text or "").strip()
+
     if not normalized_text:
         return 3.0
 
-    cjk_chars = len(re.findall(r"[\u4e00-\u9fff]", normalized_text))
-    words = len(re.findall(r"[A-Za-z0-9]+", normalized_text))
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for reporting this issue and for providing a clear Bulgarian example with a focused implementation.

After reviewing the change across a broader set of languages and edge cases, I have decided not to merge this PR. While the reported Cyrillic overestimation is real, this approach adds a script-specific rule to a duration estimator that has similar limitations for other non-ASCII, whitespace-delimited languages. It also changes the existing Russian estimate enough to fail the current regression test, while the PR does not include the Bulgarian and CJK regression tests described in its summary. Long unbroken Cyrillic tokens and some extended or decomposed Unicode forms can also be underestimated or handled inconsistently.

Addressing this safely would require a broader duration-estimation design and a diverse multilingual test matrix rather than another script-specific branch. We are not planning to take on that larger change at this time, so I am closing the PR instead of asking you to rework it further.

I appreciate the detailed reproduction steps and the time you spent investigating the problem. Thank you again for the contribution.

PR #1097, opened 2026-07-03
fix(llm): generate_terms return type, provider redacted defaults, structured errors

techdou:
## Summary

A batch of code-review-driven fixes targeting the LLM service, pipeline robustness, and API surface. All changes are covered by unit tests and backward-compatible with existing call sites.

## Background

A focused code review surfaced several real bugs and maintainability issues. This PR addresses the generic ones (local/personal branding and Windows tooling are kept out).

### Bug fixes
- **`generate_terms` return type** (`app/services/llm.py`): when the provider returned an error string, `generate_terms` propagated it as a `str` despite its `List[str]` signature. The downstream task pipeline only checks `if not video_terms:`, so a truthy `"Error: ..."` string passed through and was eventually sent to Pexels/Pixabay as a search term. Now returns `[]` so the task layer's failure branch fires correctly.
- **Redacted provider defaults** (`app/services/llm.py`): `qwen`/`cloudflare`/`ernie` had `base_url`/`model_name` left as literal `"***"` (sanitization residue), which made these providers effectively unusable. Real defaults restored.

### Refactor
- **Provider registry** (`app/services/llm.py`): replaced the ~460-line `if/elif` provider dispatch chain with a data-driven `_PROVIDER_REGISTRY` (22 providers) plus 10 `_call_*` functions. Adding a provider is now one `ProviderConfig` entry instead of editing 4 scattered locations (config read, hardcoded validation whitelist, provider branch, OpenAI fallback). Behavior and the public `_generate_response(prompt) -> str` contract are unchanged.
- **Typo**: `generate_terms` logged `"failed to generate video script"`; corrected to `"... terms"`.

### New capabilities
- **Structured LLM result** (`app/services/llm.py`): `generate_response_structured(prompt) -> LLMResult` returns `{ok, text, error_code, error_message}` with `error_code` classified as `MISSING_API_KEY` / `MISSING_MODEL` / `MISSING_BASE_URL` / `PROVIDER_DISABLED` / `EMPTY_RESPONSE` / `PROVIDER_ERROR` / `UNKNOWN_PROVIDER`. Lets programmatic callers decide retry vs. fallback vs. config prompt instead of string-matching on `"Error: ..."`. The legacy string-returning `_generate_response` is kept as a thin wrapper for backward compatibility.
- **Pipeline artifact persistence + resume** (`app/services/task.py`): each stage (script/terms/audio/subtitle/materials/video) writes its artifacts to `storage/tasks/<task_id>/manifest.json`. `start(..., resume=True)` reuses completed stages whose artifact files still exist. Note: the TTS `sub_maker` is in-memory only, so resume regenerates audio for the TTS path but fully reuses `custom_audio_file` and all pure-data/file artifacts. Defaults to `resume=False` (existing behavior unchanged).
- **Health endpoint** (`app/router.py`): mounts the existing but previously-unwired `/ping` controller (returns `"pong"`).
- **Async LLM endpoints** (`app/controllers/v1/llm.py`): `/scripts`, `/terms`, `/social-metadata` are now `async def` and offload the synchronous provider SDKs to `asyncio.to_thread`, so long LLM calls no longer consume FastAPI worker threads.

### Config
- **`config.example.toml`**: added root-level `listen_host` / `listen_port` / `project_name` / `project_version` / `log_level`. Previously these were absent from the example, so `load_config()` copying example → `config.toml` on first run produced a config without them, falling back to hardcoded defaults silently.

## Test plan

- [x] `python -m unittest test.services.test_llm` — 63 tests pass (1 integration test skipped)
- [x] `python -m unittest test.services.test_task` — 15 tests pass
- [x] `python -m unittest test.services.test_schema test.services.test_cli test.services.test_material test.services.test_state` — no regressions
- [x] `curl http://127.0.0.1:18080/ping` returns `"pong"`
- [x] Verified `generate_terms` with a provider error returns `[]` (not a string)
- [x] Verified `_PROVIDER_REGISTRY` has 22 entries with no `"***"` placeholders
- [x] Verified `resume=True` reuses manifest artifacts and skips completed LLM stages

## Notes

- A pre-existing test (`test_voice.py::test_gemini_tts_uses_legacy_submaker_fields`) fails in isolation when `storage/temp/` does not exist; this is unrelated to this PR and not touched here.

### DIFF
--- a/app/controllers/v1/llm.py
+++ b/app/controllers/v1/llm.py
@@ -1,3 +1,5 @@
+import asyncio
+
 from fastapi import Request
 
 from app.controllers.v1.base import new_router
@@ -17,13 +19,21 @@ from app.utils import utils
 router = new_router()
 
 
+# 这些 LLM 端点声明为 async，并通过 asyncio.to_thread 在后台线程执行同步的
+# generate_* 调用。原因：现有 20+ provider 的 SDK（dashscope / g4f / requests
+# 类的 cloudflare/ernie/pollinations/litellm 等）都是同步阻塞接口，无法直接
+# 改成原生 async。如果端点本身是同步 def，FastAPI 会把请求交给 anyio 线程池
+# （默认 40 个 worker），LLM 等待期间会占满线程池，挤占其它同步端点。改为
+# async def 后，等待期间不占线程池容量；同步 SDK 调用仍由 to_thread 隔离，
+# 避免阻塞事件循环。WebUI / CLI 走 task.start 同步直调路径，不受影响。
 @router.post(
     "/scripts",
     response_model=VideoScriptResponse,
     summary="Create a script for the video",
 )
-def generate_video_script(request: Request, body: VideoScriptRequest):
-    video_script = llm.generate_script(
+async def generate_video_script(request: Request, body: VideoScriptRequest):
+    video_script = await asyncio.to_thread(
+        llm.generate_script,
         video_subject=body.video_subject,
         language=body.video_language,
         paragraph_number=body.paragraph_number,
@@ -39,8 +49,9 @@ def generate_video_script(request: Request, body: VideoScriptRequest):
     response_model=VideoTermsResponse,
     summary="Generate video terms based on the video script",
 )
-def generate_video_terms(request: Request, body: VideoTermsRequest):
-    video_terms = llm.generate_terms(
+async def generate_video_terms(request: Request, body: VideoTermsRequest):
+    video_terms = await asyncio.to_thread(
+        llm.generate_terms,
         video_subject=body.video_subject,
         video_script=body.video_script,
         amount=body.amount,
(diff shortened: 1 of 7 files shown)


harry0703:
Thank you for identifying the `generate_terms()` return-type issue.

I have adapted the focused fix into `main` as commit `8cbaa0e`, while preserving your authorship. Provider errors now return an empty list instead of a truthy `"Error: ..."` string, so the task stops at term generation and does not continue into audio generation or material searches.

I also added regression coverage at both the LLM service and task orchestration levels. The targeted tests, the current deterministic CI test set, and the complete LLM/task test suites all pass.

I did not absorb the broader provider registry, pipeline resume, async endpoint, or configuration changes from this PR because the current `main` already has a newer shared provider registry, while the resume behavior needs separate design and validation.

Thanks again for the useful bug report and implementation.

PR #1099, opened 2026-07-07
fix(voice): read duration for non-mp3 custom audio files

pcbntt:
## Problem

Uploading a custom audio file that isn't `.mp3` (e.g. an `.m4a` from the WebUI, or `.wav`/`.aac`) aborts video generation:

```
ERROR | app.services.voice:get_audio_duration - Invalid target type: <class 'str'>
ERROR | app.services.task:generate_audio - failed to get audio duration from custom audio file.
ERROR | __main__ - Video Generation Failed
```

`get_audio_duration()` only routes string paths ending in `.mp3` to the moviepy/ffmpeg reader; every other extension falls through to the `else` branch, logs `Invalid target type`, and returns `0.0`, which fails the task.

## Fix

`moviepy`'s `AudioFileClip` already decodes any ffmpeg-supported container, so the `.mp3` guard is unnecessary. Route every string path to the reader and rename the helper `_get_audio_duration_from_mp3` → `_get_audio_duration_from_file` to reflect it is not mp3-specific. `SubMaker` handling is unchanged.

## Verification

- Reproduced with a real `.m4a` upload; with the fix `get_audio_duration()` returns the true duration (`28.89s`) instead of `0.0`.
- Added focused tests in `test/services/test_voice.py`:
  - non-mp3 paths (`.m4a` / `.wav` / `.aac`) return the real duration (mocked `AudioFileClip`)
  - a missing file returns `0.0`
- `python -m unittest` for the two new tests passes.

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -1670,32 +1670,32 @@ def _get_audio_duration_from_submaker(sub_maker: SubMaker):
         return 0.0
     return legacy_offsets[-1][1] / 10000000
 
-def _get_audio_duration_from_mp3(mp3_file: str) -> float:
+def _get_audio_duration_from_file(audio_file: str) -> float:
     """
-    获取MP3音频时长
+    获取音频文件时长（支持 mp3/m4a/wav/aac 等 ffmpeg 可解码的格式）
     """
-    if not os.path.exists(mp3_file):
-        logger.error(f"MP3 file does not exist: {mp3_file}")
+    if not os.path.exists(audio_file):
+        logger.error(f"audio file does not exist: {audio_file}")
         return 0.0
 
     try:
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the clear fix and the focused tests.

I verified this against the current main branch and confirmed the issue: custom audio paths were only accepted when they ended with .mp3, so formats like .m4a, .wav, and .aac could fail duration detection and stop video generation.

I also tested the PR after merging it with the latest main branch:

- CI Python tests passed
- test/services/test_voice.py passed
- the focused audio duration tests passed
- the full test/services suite passed

This is a small and useful improvement, so I have merged it. Thanks again for the contribution.

PR #1101, opened 2026-07-09
feat(webui): upload custom background music from the UI

pcbntt:
## Summary

Adds an `st.file_uploader` to the **Custom Background Music** block so users can upload an `.mp3` directly from the WebUI. Previously this block only had a text field referencing a file already present inside `resource/songs/` (which lives in the image, with no volume), so there was no way to add a track from the UI.

The uploaded file is saved through a new shared helper `utils.save_bgm_upload(filename, data)` into `resource/songs/` — the **same path and whitelist** as the existing `POST /musics` API — and set as `params.bgm_file`. Filename sanitization was extracted from the video controller into `utils.sanitize_upload_filename` so the API and the WebUI share one traversal-safe implementation. An `st.audio` player is shown under the uploader for a quick preview, mirroring the existing custom-audio upload UX.

**Scope is upload-only.** The native audio loop (`afx.AudioLoop`), the request schema, and the existing endpoints are unchanged. The original text input for referencing an already-present file is kept.

### Changes
- `refactor`: move upload filename sanitization to `app/utils/utils.py` (pure function, raises `ValueError`; the controller translates it into the existing HTTP 400)
- `refactor`: share BGM save logic (`save_bgm_upload`) between `POST /musics` and the WebUI
- `feat(webui)`: `st.file_uploader` in the custom BGM block, with an `st.audio` preview player
- `i18n`: add the new strings to `en`, `ru`, `zh`, `pt`

## Test plan
- New `test/services/test_bgm_upload.py`: sanitization (path traversal, empty/dotted names) and the save helper (writes `.mp3`, rejects non-`.mp3`, blocks traversal).
- Existing `test/services/test_webui_i18n.py` coverage tests updated (en + ru) and passing.
- Verified `app.controllers.v1.video` imports and all tests pass inside the `ghcr.io/harry0703/moneyprinterturbo:latest` image env.
- Manually validated end-to-end in that container image: upload → audio preview → **full video generation using the uploaded background music**.


harry0703:
Thanks again for this contribution. I integrated the custom BGM upload flow into main in commit a4ee0e21d29bfc624de4d341da5d516f24899bca, while preserving your authorship.

While absorbing the PR, I extended the implementation with immutable UUID-backed storage, FFmpeg-only audio validation, support for common audio formats, a consistent 30 MB limit, clearer API error handling and OpenAPI metadata, cached WebUI validation, and regression coverage for the service and UI paths.

I verified MP3, M4A, AAC, WAV, FLAC, OGG, OPUS, and WMA uploads, the real API upload/list flow, valid and invalid WebUI states, and an end-to-end generated video using the uploaded BGM.

Since the implementation is now available on main, I have closed this PR. Thank you for identifying and implementing this useful workflow!

PR #1104, opened 2026-07-10
feat(video): add ZoomIn/ZoomOut transitions

Kizuno18:
Adds two Ken Burns style transitions to the existing set: **ZoomIn** ramps the clip from 1.0x to 1.2x across its duration, **ZoomOut** does the reverse. They give AI/stock footage a bit of motion, which helps a lot on segments where the source clip is fairly static.

Implementation notes:

- The zoom is a center crop + resize done with Pillow, which moviepy already pulls in — no new dependency.
- Both transitions are selectable in the WebUI and included in the `Shuffle` pool in `app/services/video.py`.
- `ZoomIn`/`ZoomOut` labels added to all nine locale files (the ru coverage test in `test_webui_i18n` passes).
- New `test/services/test_video_effects.py` covers geometry/duration preservation and that the zoom actually changes pixel content over time; added it to the CI unittest list since it runs in ~0.1s.

Ran the full CI test set locally: 20 tests, OK (1 pre-existing skip).


harry0703:
Thank you for this contribution. ZoomIn and ZoomOut are useful additions, especially for static and low-motion footage.

We integrated an adapted version into `main` in commit e5dec5a556ed550e10f37d0488e5ea2b754ff0a9. The original PR had diverged from the current WebUI state handling, which now uses `stable_selectbox`, and our locale maintenance strategy has also changed, so we absorbed the implementation directly instead of merging the branch as-is.

During validation with real footage, we also changed the frame scaling to use centered subpixel sampling to reduce visible jitter, kept the original 20% zoom range, and added regression tests for subpixel continuity and odd-resolution center stability. The WebUI options, Shuffle integration, and CI coverage are included. Your authorship and original commit title have been preserved in the resulting commit.

I am closing this PR because the feature is now available in `main`. Thank you again for the idea, implementation, and tests!

PR #1106, opened 2026-07-10
feat(video): global clip speed multiplier from the UI

pcbntt:
## What

Adds a global playback-speed multiplier for the background material clips,
exposed as a single slider in the WebUI (Video Settings → **Clip Speed**).

Short-form creators often want a slight speed-up (1.1x–1.5x) for pacing, or a
slow-down for emphasis. Today every clip plays at 1.0x with no way to change it.

## How

- New `VideoParams.video_clip_speed` (default `1.0`).
- New pure helper `utils.normalize_clip_speed(value)` — coerces to float and
  clamps into `[0.5, 2.0]`; returns `1.0` for missing, non-numeric, or
  non-positive input.
- `combine_videos` applies `clip.with_speed_scaled(speed)` per clip, after the
  resize/transition/max-duration trim and before the temp clip is written.

The speed affects **only the visual material clips**. Narration and subtitles are
composited in a later stage, so audio sync is untouched. Because the pipeline
already loops processed clips to cover the narration duration, speeding clips up
(each becomes shorter) is auto-compensated — more clips enter the loop.

## Backward compatibility

Default is `1.0`. `normalize_clip_speed` returns exactly `1.0` for the default,
and `combine_videos` **skips the moviepy effect entirely** when the factor is
`1.0`. Existing behavior and output are unchanged when the feature is unused.

## Scope

Deliberately limited to a single global multiplier. Out of scope: per-clip speed,
random per-clip variation, speed ramps, and any change to narration/subtitles.

## Tests

`test/services/test_clip_speed.py` — 5 unit tests covering the default, both
clamp bounds, a valid pass-through, and non-numeric/non-positive rejection.
i18n coverage (`en`/`ru`) verified by the existing `test_webui_i18n.py`.

## i18n

`"Clip Speed"` added to `en`, `ru`, `zh`, `pt`.

### DIFF
--- a/app/models/schema.py
+++ b/app/models/schema.py
@@ -77,6 +77,7 @@ class VideoParams(BaseModel):
     video_concat_mode: Optional[VideoConcatMode] = VideoConcatMode.random.value
     video_transition_mode: Optional[VideoTransitionMode] = None
     video_clip_duration: Optional[int] = 5
+    video_clip_speed: Optional[float] = 1.0
     match_materials_to_script: bool = False
     video_count: Optional[int] = 1
 
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -286,6 +286,7 @@ def generate_final_videos(
             video_transition_mode=video_transition_mode,
             max_clip_duration=params.video_clip_duration,
             threads=params.n_threads,
+            clip_speed=params.video_clip_speed,
         )
 
         _progress += 50 / params.video_count / 2
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -541,6 +541,7 @@ def combine_videos(
     video_transition_mode: VideoTransitionMode = None,
     max_clip_duration: int = 5,
     threads: int = 2,
+    clip_speed: float = 1.0,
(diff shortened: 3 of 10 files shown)


harry0703:
Thank you for the contribution, Paulo. I have integrated this feature into the current main branch while preserving your authorship:

https://github.com/harry0703/MoneyPrinterTurbo/commit/396b045cd7b693aad07d48f3ec94e96f83d74634

While adapting it to the latest codebase, I also:

- resolved the current WebUI conflicts and wired the value through the task pipeline;
- normalized invalid or out-of-range speed values, including restored task state;
- adjusted source segmentation based on playback speed so 0.5x playback does not skip footage and 2.0x reads enough source content;
- added regression tests and verified the feature with real video renders.

Validation: 68 tests passed (1 skipped), and the 0.5x and 2.0x outputs decoded successfully at 1080x1920, 30 fps, with the expected audio-aligned duration.

Since the implementation is now available on main, I am closing this PR as integrated. Thanks again!

PR #1107, opened 2026-07-10
feat: add ai_image video source (AI-generated images to video)

Kizuno18:
This adds **AI Image** as a video source, for topics where stock footage from Pexels/Pixabay just doesn't match the script (niche subjects, abstract topics, consistent visual style). One image is generated per script term and turned into a clip with a Ken Burns effect — zoom in/out or a four-direction pan implemented as a moving crop window over the still, so the canvas stays constant with no black borders. The clips then flow through the existing combine/TTS/subtitle pipeline unchanged.

**Providers**
- OpenAI images (DALL·E 3/2, gpt-image-1, or any OpenAI-compatible endpoint via base URL)
- Stability AI
- Pollinations (works with no API key)
- Midjourney/custom proxies that expose an OpenAI-compatible images endpoint

Image sizes are selected per model family — dall-e-2 and gpt-image-1 reject dall-e-3's 1792x1024/1024x1792, so each gets a size it actually accepts. There's also an optional "enhance prompt" toggle that rewrites each term with the configured LLM before generation.

**Robustness details**
- 3 attempts with backoff per image; response bytes are verified to be a real image before writing, so a 200-status rate-limit page never becomes a corrupt .png that only explodes later in moviepy
- 120s timeout on the images client instead of the SDK's 10-minute default
- prompts are fully percent-encoded in the Pollinations URL path (a term containing `/` can't rewrite the request path)
- the term list is copied before shuffling so the caller's order (persisted in script.json) is untouched

**Config**: settings live under namespaced `{provider}_image_*` keys (documented in config.example.toml) so the image panel can never overwrite the LLM panel's `openai_api_key`/`openai_base_url`; reading falls back to the shared OpenAI key for convenience. The WebUI gets a provider selector with per-provider tips and start-gate validation, and all nine locales include the new strings.

**No new dependencies** — Pillow and numpy already come with moviepy.

Tests: `test/services/test_material_ai_image.py` covers config-key resolution, per-model sizes, URL encoding and the no-mutation guarantee, and is added to the CI unittest set (28 tests pass locally). I also ran the whole pipeline end-to-end against a local OpenAI-compatible mock provider: portrait 1080x1920 h264 output with TTS audio and burned subtitles.

Unrelated to #1104 (zoom transitions for regular video clips) — the Ken Burns effect here is internal to the image-to-clip conversion, and the two apply cleanly in either order.


arsenyinfo:
Prompt enhancement for AI-image clips receives an empty script because the pipeline’s generated script is never passed to `get_video_materials`
- Lens: **correctness**
- Priority: P2
- Location: app/services/task.py:483
- Scenario: User supplies only `video_subject` (empty script text area in WebUI or a direct API call); `_run_pipeline` holds the generated script in a local variable but passes `params.video_script` (empty) into `get_video_materials`, so `enhance_prompt_with_llm` skips the “Context from video script” block and rewrites each term in isolation.
- Potential solution: Add a `video_script` parameter to `get_video_materials` sourced from the actual generated-script local variable in `_run_pipeline` (task.py:1047) and use it at line 483 instead of `params.video_script`.

`match_materials_to_script` is ignored for the `ai_image` source, so terms are still shuffled while final assembly stays sequential
- Lens: **correctness**
- Priority: P2
- Location: app/services/task.py:472–484; app/services/material.py:584–586
- Scenario: API caller sets `video_source="ai_image"`, `match_materials_to_script=True`, and `video_concat_mode="random"`. The stock-download branch and `generate_final_videos` both force sequential ordering under match mode, but the `ai_image` branch passes `params.video_concat_mode` unchanged, so `generate_ai_images` shuffles the terms while final assembly expects script order.
- Potential solution: In the `ai_image` branch, pass `video_concat_mode=VideoConcatMode.sequential if params.match_materials_to_script else params.video_concat_mode`, mirroring the download branch.

Ken Burns effect re-runs an identical LANCZOS upscale on every frame, making pan effects disproportionately expensive
- Lens: **performance**
- Priority: P2
- Location: app/services/material.py:709
- Scenario: Always for the AI-image video source; every clip renders `video_duration` seconds at 30 fps (~150 frames for a 5 s clip), and four of the six `ZoomEffect` values keep `scale=1.15` constant but still resize the cropped tile each frame.
- Potential solution: For the four `PAN_*` effects, resize the still once to `(int(width*1.15), int(height*1.15))` with LANCZOS and slide a `(width, height)` crop window across it inside `ken_burns`. For all effects, switch the per-frame resize from `Image.LANCZOS` to `Image.BICUBIC` to avoid the costly high-quality resample for 30 fps motion.

Enhancing AI-image prompts performs one blocking LLM call per term and the qwen adapter mutates global `dashscope.api_key`
- Lens: **performance**
- Priority: P2
- Location: app/services/material.py:592; app/services/llm.py:214
- Scenario: User enables “Enhance AI image prompts using LLM” with a long script; each term waits for a separate synchronous LLM round-trip (typically 2–10 s), fully blocking image generation. With a qwen-configured LLM, any future concurrent enhancement calls would race on the module-level `dashscope.api_key = api_key` assignment.
- Potential solution: Run enhancement calls in a bounded `ThreadPoolExecutor` for thread-safe providers (OpenAI/Azure/litellm/modelscope/cloudflare/gemini) and re-attach results in original order. For the qwen adapter, serialize the calls or guard the `dashscope.api_key` assignment with `threading.Lock` so concurrent calls cannot corrupt the shared API key.

Generated AI-image bytes are written to disk and immediately re-read by `ImageClip`
- Lens: **performance**
- Priority: P3
- Location: app/services/material.py:621–622 and 675
- Scenario: Always, once per generated image.
- Potential solution: Keep `Image.open(io.BytesIO(image_data)).verify()` for validation, then pass the in-memory image to moviepy as `ImageClip(np.asarray(Image.open(io.BytesIO(image_data))))` instead of `ImageClip(image_path)`. This removes one PNG decode and one file open per clip and is compatible with the moviepy 2.2.1 transform path used by `convert_image_to_video`.

`_image_provider_setting` searches legacy/spaced key variants that no in-repo writer ever produces
- Lens: **simplicity**
- Priority: P3
- Location: app/services/material.py:384–391; webui/Main.py:2270–2280; test/services/test_material_ai_image.py:44–48
- Scenario: Always — every setting read iterates speculative keys such as `stability_ai_api_key`, `"stability ai_image_api_key"`, `midjourney_api_key`, etc., but the feature only writes `{provider}_image_{suffix}` keys and only documents the namespaced keys plus the shared `openai_api_key` fallback.
- Potential solution: Reduce the candidate list to `{provider}_image_{suffix}` plus the intentional `openai_api_key` fallback; remove the dead-key `pop` calls in `set_image_setting`; delete `test_legacy_spaced_key_fallback`. Keep the tests that guard against falling back to the LLM chat model or text endpoint.

Provider registry and validation rules for AI-image providers are duplicated between `material.py` and the WebUI, which also imports a private helper
- Lens: **simplicity**
- Priority: P3
- Location: app/services/material.py:520 and 530–543; webui/Main.py:2179–2185 and 3427–3448 and 2202–2210/3431/3445
- Scenario: Adding, renaming, or removing an image provider requires editing two independent ID tuples, two parallel requirement blocks, and maintaining the cross-module import of `material_service._image_provider_setting`.
- Potential solution: Introduce a single `IMAGE_PROVIDERS` mapping in `material.py` keyed by provider ID with `label`, `default_model`, and required settings, plus a public `image_provider_requirements(provider)` helper. Drive the WebUI selectbox and start-gate validation from this one mapping, replacing `known_providers`, the duplicated `image_providers` list, and the underscore-prefixed import.

`convert_image_to_video.ken_burns` contains a dead fallback arm and five near-duplicate pan/zoom branches
- Lens: **simplicity**
- Priority: P3
- Location: app/services/material.py:681–700; app/services/material.py:627–631
- Scenario: Never — `effect` is a `ZoomEffect` enum and the only caller passes `random.choice(list(ZoomEffect))`, so the `else` “ZOOM_IN and fallback” branch describes an unreachable state. The four pan branches are copy-paste variants differing only in which coordinate carries the progress offset.
- Potential solution: Replace the if/elif tower with a per-`ZoomEffect` lookup table of `(scale, offset_x, offset_y)` triples, remove the fallback comment, and raise `ValueError` on an unexpected effect to keep the closed domain closed.

---
🔍 Reviewed by [nitpicker](https://github.com/arsenyinfo/nitpicker)

Darian222:
您好！

非常感谢您详细的问题报告和提出的解决方案建议。我们已仔细阅读并充分理解了您提出的各项问题和优化方向。这些宝贵的反馈对我们改进产品至关重要。

我们会尽快对这些建议进行评估和讨论，并计划在后续版本中采纳和实施。

再次感谢您的贡献！

此致，
敬礼

Arseny Kravchenko ***@***.***> 于2026年8月18日周二 14:33写道：

Kizuno18:
Merged current main into the branch and applied the review findings from the nitpicker run.

Correctness:
- `_run_pipeline` now hands the generated script down to `get_video_materials`, so prompt enhancement actually sees the script when the user only supplies a subject — before it always got the empty `params.video_script`.
- the `ai_image` branch forces `VideoConcatMode.sequential` when `match_materials_to_script` is set, same as the stock-download branch, so the terms are no longer shuffled while final assembly expects script order.

Simplicity:
- `IMAGE_PROVIDERS` in `material.py` is now the single provider registry (label, default model, required settings), with public `normalize_image_provider`, `image_provider_requirements`, `image_provider_default_model` and `image_provider_setting`. The WebUI selectbox and the start-gate validation read that mapping instead of keeping their own provider list and importing the private helper.

The merge itself: the `ai_image` branch now sits next to the new `loomloom` source in `get_video_materials`, both are listed in `config.example.toml` and the start gate, and the image-provider settings go through `_set_runtime_config`/`_delete_runtime_config` instead of writing `config.app` directly, as `test_webui_runtime_config_updates_do_not_use_blocking_writes` requires.

New tests cover the script propagation (both at pipeline level and in `get_video_materials`), the sequential ordering under match mode, and the registry helpers. Suite is green locally and on CI here.

Still open from that review: the per-frame LANCZOS resize in `ken_burns`, the sequential blocking LLM calls for prompt enhancement, the disk round-trip before `ImageClip`, and the speculative legacy/spaced keys in `image_provider_setting`. I can fold them into this PR if you want them together, otherwise I'd rather keep this one focused and follow up separately.

harry0703:
Thank you for the substantial effort you put into the AI-image material source, the follow-up fixes, tests, and performance improvements. After reviewing the overall scope, we have decided not to move forward with this PR at the moment. It changes the image-provider architecture, material generation, WebUI settings, task orchestration, and video rendering together, which creates a review and long-term maintenance surface that is currently too large for the project. We are closing this PR for now without deleting your branch. If we revisit AI-image support later, smaller focused proposals separated by provider, core image-to-video behavior, and WebUI integration would be much easier to evaluate safely. We truly appreciate your work and your thoughtful responses to the earlier review.

PR #1112, opened 2026-07-12
Tidy implementation detail in MoneyPrinterTurbo (#1093)

bglglzd:
Small scoped patch based on the reported behavior.

Related to #1093.


harry0703:
Thank you for taking the time to submit this patch. After reviewing the complete diff, this PR only adds newline characters to two empty `__init__.py` files and removes trailing whitespace in `schema.py` and `upload_post.py`. It does not change `estimate_no_voice_duration()` or any other runtime logic related to issue #1093, so it cannot affect the reported Cyrillic duration behavior.

PR #1094 contains the substantive implementation for that issue and will be reviewed separately. I am closing this PR because it does not introduce a functional fix. Thank you again for the contribution.

PR #1118, opened 2026-07-16
feat: add Storyblocks as an optional video source

Kizuno18:
Adds Storyblocks as a fourth stock-footage source next to Pexels/Pixabay/Coverr, following the same optional pattern: off unless the user configures keys.

A few implementation notes worth calling out:

- Auth is the Storyblocks partner scheme: every request carries `APIKEY`/`EXPIRES`/`HMAC`, where the HMAC-SHA256 signs the resource path with the private key concatenated with the expiry. Since two keys are involved, `storyblocks_api_keys` entries are `PUBLIC_KEY:PRIVATE_KEY` pairs — this keeps the existing key-rotation and the material APIs panel working unchanged.
- Search uses `GET /api/v2/videos/search` with `content_type=footage` and `orientation` mapped from the selected aspect ratio. The parser accepts both the v2 `results` envelope and the older `info` one.
- The search response has no directly downloadable mp4, and each call to the per-item download endpoint is a licensed download with a ~20-minute signed URL. So search results carry the `/api/v2/videos/{id}/download` resource and `material._resolve_material_url()` resolves the real URL lazily, right before `save_video()`, in both the normal and the script-order download paths. No licenses are burned on clips that never get downloaded.
- `storyblocks_project_id` / `storyblocks_user_id` are optional config keys (the API requires them on every request for usage attribution; they default to `moneyprinterturbo`).

Storyblocks API access requires a partner agreement (https://www.storyblocks.com/resources/business-solutions/api), which the key-field help text and `config.example.toml` both mention, so users don't hunt for a free key that doesn't exist.

Covered by `test/services/test_material_storyblocks.py`: HMAC construction against a fixed timestamp, both response envelopes, duration filtering, lazy download resolution (v2 and legacy shapes), credential-pair validation, and the `download_videos` dispatch. Full suite + ruff + the coverage gate pass locally on 3.11.


harry0703:
Thank you for the thoughtful Storyblocks integration and the extensive tests. After reviewing it against the current Storyblocks API documentation, we found that the download endpoint and response shape do not match the official API. Storyblocks also requires a specialized commercial setup that we do not plan to maintain at this time, so we will not merge this PR. We appreciate the work and detailed explanation.

Kizuno18:
Thanks for checking it against the API reference — you're right on both counts. The download resource is `/api/v2/videos/stock-item/download/{stock_item_id}` and the response maps format → resolution → signed URL rather than carrying a `url` field; I had inferred the path from the v1 docs. Fixed in my fork, and no hard feelings on the policy call — makes sense not to carry a partner-agreement integration you can't test.

PR #1119, opened 2026-07-16
i18n: add ZoomIn/ZoomOut labels to de/es/id/pt/tr/vi

Kizuno18:
The ZoomIn/ZoomOut transitions from e5dec5a shipped with en/ru/zh translations only, so the transition selectbox shows the raw `ZoomIn`/`ZoomOut` keys for the other six locales. This adds the two labels to de, es, id, pt, tr and vi — 12 lines, nothing else touched.


harry0703:
Thank you for adding the missing ZoomIn and ZoomOut translations. We currently prefer to maintain translations centrally so wording stays consistent across all locales, so we will not merge this PR. We appreciate your contribution.

PR #1124, opened 2026-07-22
feat(llm): add AIMLAPI rebate OAuth

D1m7asis:
## Summary

- promote `aimlapi.com` to the first and default recommended LLM provider, using the production inference endpoint and `anthropic/claude-sonnet-5`
- add AIMLAPI OAuth Device Authorization Grant support to the Settings dialog, including automatic polling, fallback consent links, localized status messages, and API key persistence
- attribute OAuth and trusted AIMLAPI inference requests for partner rebates without leaking partner headers to arbitrary custom base URLs

## Why

MoneyPrinterTurbo already supported AIMLAPI as an OpenAI-compatible provider, but users still had to create and copy API keys manually and requests were not consistently attributed to the MoneyPrinterTurbo partner integration. This change adds the device authorization flow used by AIMLAPI integrations and makes the provider ready for rebate attribution.

## Implementation

- add one environment-aware AIMLAPI endpoint and partner configuration resolver with production defaults
- keep all endpoints, partner metadata, and the requested USD limit overridable through `AIMLAPI_*` environment variables
- add a server-side authorization service for creating and polling RFC 8628 device requests; temporary device data stays in Streamlit session state and sensitive values are never logged
- preserve the `/app` prefix from `AIMLAPI_VERIFICATION_BASE_URL`, producing the current `/app/agent/authorize` consent route
- add `X-AIMLAPI-Source: agent` and `X-AIMLAPI-Partner-ID` to OAuth calls and trusted AIMLAPI inference hosts only
- update provider defaults, Docker environment forwarding, Settings layout, help copy, styling, and all supported locales
- keep existing explicitly configured provider, model, and base URL values unchanged
- intentionally omit Tolt partner/customer identifiers because they remain backend-only payout metadata

## AIMLAPI documentation audit

The implementation was checked against `HEADERS.md`, `PARTNERS.md`, `MODELS.md`, and the RAGFlow device-flow notes in `aimlapi-integrations`:

- OAuth start/poll and inference requests carry both required attribution headers
- arbitrary user-defined base URLs do not receive AIMLAPI partner headers
- `anthropic/claude-sonnet-5` was verified in the production `/v1/models` catalog with `type=openai/chat-completions`
- device code, generated keys, and temporary OAuth state remain server-side and are not logged
- endpoint and partner overrides remain environment-configurable

## Validation

- `ruff check` passed for application, WebUI, tests, and the MPT helper
- Python `compileall` passed
- both Docker Compose files passed `config --quiet`
- targeted suite: **123 passed, 1 skipped**, plus **3913 subtests passed**
- production model catalog returned the configured default model as a chat-completions model
- production consent route `/app/agent/authorize` resolves successfully
- full suite: **488 passed, 10 skipped**, coverage **72%**; one unrelated Windows-only assertion still fails because it compares POSIX and Windows path separators in `test_worker_logs_are_available_without_streamlit_session_state`

## Production prerequisites

Two backend/account prerequisites remain before the interactive flow can complete:

1. `app.aimlapi.com` currently returns `403 Unknown or disabled partner` for the MoneyPrinterTurbo partner ID. The partner must be registered/enabled in the production `rebates_partners` table and added to `aimlapi-integrations/PARTNERS.md` as required by that registry's rules.
2. The API key currently stored in the local test configuration returns `401` on production inference. A valid production key is required; once the partner is enabled, the OAuth flow should generate and persist one automatically.


harry0703:
Hi Dmitry, thanks for putting this together and for the detailed tests.

I reviewed the implementation. The OAuth approach is useful, but I do not think this PR is ready for the project in its current form. The production partner is still disabled, so the authorization flow cannot be verified end to end. The PR also combines the OAuth integration with changing the default and recommended provider and removing the existing Kimi recommendation, which is a broader product decision than the technical integration itself.

For now, I am going to close this PR. If the partner is enabled in production and we decide to revisit the integration, I would prefer a smaller PR that keeps the current provider order and defaults, and focuses only on OAuth, attribution, and tests.

Thanks again for the time and thorough work you put into this.

PR #1126, opened 2026-07-23
Add HyperFrames support

fotiDim:
This PR adds support for [HyperFrames](https://hyperframes.heygen.com). It generates videos using code, similar to [Remotion](https://www.remotion.dev). HyperFrames videos are editable. After generation, apart from the rendered video, you get a HyperFrames project that you can open with the HyperFrames editor.

To achieve the above, this PR introduces a `renderer` field , which can be set to either moviepy or HyperFrames.

Future work:
- Since HyperFrames videos are editable we could also load an existing project to allow further editing by prompting. This would be part of another PR.


samsadlonka:
Very good idea! Waiting for approval!

harry0703:
Thank you for the substantial work on the HyperFrames renderer. This branch has not received code updates since July 22 and now has extensive conflicts with the current main branch. Given the size of the change, reviewing or repairing it safely would require a fresh implementation against the current architecture. To keep the PR queue current, we are closing this PR for now. You are very welcome to open a new, focused PR based on the latest main if you would like to continue the work.

PR #1130, opened 2026-07-23
fix(material): redact Pixabay API key from logs

its-How:
## Summary
- stop logging the Pixabay request URL because it contains the API key query parameter
- retain useful diagnostics by logging the provider, search term, and proxy configuration
- add a regression test that prevents the API key from appearing in log messages

## Security impact
This prevents configured Pixabay API keys from being persisted in application logs during material searches.

## Validation
- uv run pytest test/services/test_material.py -q
- uv run ruff check app/services/material.py test/services/test_material.py

### DIFF
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -127,7 +127,10 @@ def search_videos_pixabay(
         "key": api_key,
     }
     query_url = f"https://pixabay.com/api/videos/?{urlencode(params)}"
-    logger.info(f"searching videos: {query_url}, with proxies: {config.proxy}")
+    logger.info(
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for addressing this. The change is small and focused, and it improves the default logging behavior without affecting material search, so I have merged it.

One follow-up would make the protection more complete: request exceptions can still include the full Pixabay URL, and proxy settings may contain credentials. If you would like to continue improving this, a follow-up PR that sanitizes exception messages and only logs whether a proxy is enabled would be very welcome.

PR #1131, opened 2026-07-23
fix(bgm): enforce ElevenLabs resource limits

arichazan:
## What changed

- read at most one 500-byte chunk from streamed ElevenLabs error responses before decoding and logging it
- pass FFmpeg `-fs` with the 200 MB proxy limit so the cap is enforced while encoding
- add regression coverage that rejects `.text` materialization and verifies the FFmpeg size option

## Why

The existing safeguards were applied only after resource consumption: `response.text` could materialize an oversized or endless error body, and the proxy size check ran only after FFmpeg finished writing. A faulty upstream or a high-entropy source could therefore exhaust worker memory or task-volume storage.

## Impact

Failed ElevenLabs generation requests now retain bounded diagnostic detail without unbounded memory use. Proxy encoding stops at the declared upload limit, while the existing post-encode validation remains as defense in depth.

## Validation

- `uv run pytest -q` — 469 passed, 11 skipped, 4032 subtests passed
- `uv run ruff check app/services/elevenlabs_music.py test/services/test_elevenlabs_music.py`
- `git diff --check`

### DIFF
--- a/app/services/elevenlabs_music.py
+++ b/app/services/elevenlabs_music.py
@@ -22,6 +22,7 @@ MAX_VIDEO_DURATION_SECONDS = 600
 MAX_PROMPT_LENGTH = 1000
 MAX_PROXY_BYTES = 200 * 1024 * 1024
 MAX_GENERATED_AUDIO_BYTES = 50 * 1024 * 1024
+MAX_ERROR_BODY_BYTES = 500
 
 
 class ElevenLabsMusicError(RuntimeError):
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the focused resource-limit improvements and the regression coverage. The FFmpeg size limit is useful, and I confirmed that the partial proxy is rejected and cleaned up when the limit is reached.

There is one remaining issue in the error-response handling. `test_connection()` still calls `requests.get()` without `stream=True`. Requests eagerly downloads the complete response body by default, so by the time `_safe_response_error()` reads a single 500-byte chunk, an oversized error response may already be fully loaded into memory. The generation request already uses `stream=True`, so that path is covered correctly.

Could you please update the subscription request to use `stream=True` and ensure the response is closed, preferably with a context manager? Please also extend the test to verify that the GET request is made with `stream=True` and that the error path reads only one bounded chunk.

Once that is addressed, this should be ready for another review. Thanks again for the contribution.

arichazan:
Thanks for catching that. I’ve updated `test_connection()` so the subscription request uses `stream=True`, and all response handling now runs inside a context manager to ensure the response is closed on both success and error paths.

I also extended the regression coverage to verify that:

- the subscription GET is called with `stream=True`
- the response is closed
- the HTTP error path requests exactly one `MAX_ERROR_BODY_BYTES` chunk and does not consume a second chunk

Validation:

- `uv run pytest -q` — 470 passed, 11 skipped, 4032 subtests passed
- `uv run ruff check app/services/elevenlabs_music.py test/services/test_elevenlabs_music.py`
- `git diff --check`

The fix is in commit `8bb6c5f`. Ready for another review—thanks again.

harry0703:
Thanks for the update. I reviewed `8bb6c5f` and verified the streaming subscription request, bounded error-body handling, response cleanup, and FFmpeg size limit. The targeted tests and Ruff checks pass locally.

This is now merged. Thanks again for the careful follow-up and contribution.

arichazan:
cool, thanks for your excellent work. glad to participate.

PR #1135, opened 2026-07-26
feat(material): preserve provider provenance

its-How:
## Summary

- preserve provider asset, creator, rendition, and response identity while downloading materials
- write an atomic no-overwrite `storage/tasks/<task_id>/materials-provenance.json` sidecar
- bind each returned local path to fresh SHA-256 and byte size in exact return order
- redact credentials, signed query parameters, raw responses, and unsafe exception text
- keep `download_videos(...) -> list[str]`, task intermediate results, and public `MaterialInfo` schema unchanged

## Safety and scope

The sidecar proves provider identity plus local byte binding only. It does not claim that provider rights or asset-specific third-party rights have been approved. Write failures and collisions fail closed.

## Tests

- full suite: 483 passed, 11 skipped, 4034 subtests passed
- Ruff: PASS
- py_compile: PASS
- git diff --check: PASS

This branch is based on current upstream `main`, including the merged Pixabay API-key log redaction from #1130.

### DIFF
--- a/app/services/material.py
+++ b/app/services/material.py
@@ -1,8 +1,12 @@
+import hashlib
+import json
 import os
 import random
+import tempfile
 import threading
-from typing import Callable, List
-from urllib.parse import quote_plus, urlencode
+from pathlib import Path
+from typing import Any, Callable, List
+from urllib.parse import quote_plus, urlencode, urlsplit, urlunsplit
 
 import requests
 from loguru import logger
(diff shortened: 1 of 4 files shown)


harry0703:
Thanks for putting this together. The provenance work and the attention to credential redaction were both useful.

I have incorporated the core idea into main in 7ddb11d, using a lighter approach that stores sanitized provider, asset, creator, rendition, and local filename metadata directly in each task's existing script.json. The metadata is preserved through the search cache, while signed query parameters, download URLs, and absolute local paths are excluded. Provenance writes are also non-blocking, so an auxiliary metadata failure cannot interrupt video generation.

I decided not to add a separate sidecar with file hashes and response identity for now, since that introduces more storage and failure semantics than the project currently needs. The useful parts of this PR are now covered on main, so I am closing it. Thanks again for the thoughtful contribution.

PR #1139, opened 2026-07-28
feat: add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- add Atlas Cloud to the centralized LLM provider registry
- use the existing OpenAI-compatible adapter with Atlas Cloud defaults
- expose the provider configuration keys in `config.example.toml` and cover registry defaults/order

## Validation
- project-pinned Ruff check and format check
- registry/config contracts on Python 3.11 and 3.12
- Python 3.12 compileall
- live OpenAI SDK chat completion using the registry defaults

No README or documentation changes.

### DIFF
--- a/app/models/llm_provider.py
+++ b/app/models/llm_provider.py
@@ -159,6 +159,12 @@ LLM_PROVIDER_REGISTRY = (
     ),
     # 聚合与统一接入平台
     LLMProviderSpec(
+        "atlascloud",
+        "Atlas Cloud",
+        default_model="deepseek-ai/deepseek-v4-pro",
+        default_base_url="https://api.atlascloud.ai/v1",
+    ),
+    LLMProviderSpec(
         "cloudflare",
         "Cloudflare AI Gateway",
         adapter="cloudflare_ai_gateway",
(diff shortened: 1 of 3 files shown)


ranggadawi13-svg (approved):

harry0703:
Thank you for adding Atlas Cloud and for keeping the implementation focused. We verified that the Atlas Cloud OpenAI-compatible endpoint and proposed model ID are still available. However, this PR no longer applies cleanly to the current main branch and now conflicts in both the provider registry and its tests. The provider integration contract has also evolved since this was opened: a first-class provider now needs an official API-key link, current English and Chinese WebUI configuration guidance, placement that preserves the maintained provider order, and updated registry/config coverage. In addition, Atlas Cloud can already be used through the existing OpenAI-compatible configuration by supplying its Base URL and model ID, so the current benefit of a dedicated entry does not justify the additional permanent UI and maintenance surface. We are therefore closing this PR for now. If usage demand grows later, a rebased proposal aligned with the current provider contract would be welcome. We appreciate your contribution and the validation work you included.

PR #1140, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for MoneyPrinterTurbo.

`.github/workflows/docker-ghcr.yml` publishes the image to `ghcr.io`, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

The image is how most people run this — it bundles the whole video pipeline so nobody has to assemble the dependencies themselves — which is exactly why the SBOM half is useful here: `sbom: true` publishes what is actually inside a given tag instead of leaving it to be discovered by running the container.

The change is two lines on the build step:

```yaml
          push: true
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image to GHCR. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect ghcr.io/harry0703/moneyprinterturbo:<tag> --format '{{ json .SBOM }}'
```

Two caveats: `mode=max` records build arguments (`provenance: true` gives a smaller record), and attestations add an extra manifest to the index, which GHCR supports.

No SLSA level claimed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

harry0703:
Thank you for this focused supply-chain improvement, and especially for transparently correcting the provenance details in your follow-up. I verified that the current published image already includes provenance but has no SBOM, that both inputs are supported by docker/build-push-action v6, and that the updated workflow passes actionlint. The PR has now been merged. The next published image will include the SBOM attestation. We appreciate your careful contribution!

PR #1142, opened 2026-07-29
Add OpenCode Go provider

WJH9102:
## What changed
- add OpenCode Go as a first-class LLM provider using the existing OpenAI-compatible adapter
- document the supported Chat Completions model scope and raw model ID format
- add runtime config path support for directory-based Docker persistence
- add provider, config, i18n, and request-contract tests

## Validation
- 92 tests passed, 1 skipped, 3843 subtests passed
- Ruff checks and formatting passed
- live OpenCode Go connection test passed with `mimo-v2.5`

## Notes
OpenCode Go models that require the Messages API are intentionally out of scope for this first integration.


harry0703:
Thank you for exploring OpenCode Go support. This PR has remained in draft without code updates since July 30, now conflicts with main, and also combines provider work with unrelated remote task preview changes. To keep the review scope clear and the queue current, we are closing it for now. A new focused PR based on the latest main would be welcome if you decide to continue.

PR #1149, opened 2026-08-01
fix(api): validate positive video generation limits

ShiroKSH:
## Summary

- Reject non-positive clip durations before they reach the media pipeline.
- Reject non-positive video counts before a task can spend provider resources and fail late.

## Root cause

The API model allowed zero, negative, and null values even though the WebUI and CLI require positive integers. A zero clip duration made the material-splitting loop stop advancing, while a non-positive video count could fail only after earlier generation stages completed.

## Validation

- `python3 -m compileall -q app/models/schema.py test/services/test_schema.py`
- `git diff --check`


harry0703:
Thanks for the fix. Validating these values at the API boundary prevents zero-duration clips from stalling the media pipeline and avoids starting work for invalid video counts. I ran the full test suite locally and everything passed, so this has been merged. Appreciate the contribution!

PR #1150, opened 2026-08-02
fix(worker): don't crash worker on stale queued task params

kapil971390:
## Bug

`RedisTaskManager.dequeue()` reconstructs `VideoParams` from whatever was serialized into Redis when the task was enqueued:

```python
task_info["kwargs"]["params"] = VideoParams(**task_info["kwargs"]["params"])
```

#1149 tightened `VideoParams` to reject `video_clip_duration`/`video_count <= 0` (previously permissive `Optional[int]`). A task that was enqueued *before* that validation was added — or is still sitting in the queue at the moment a deploy adds it — can have already been accepted with e.g. `video_count=0`. Reconstructing that same task on dequeue now raises `ValidationError`.

That happens **after** `redis_client.lpop()` has already removed the entry (`lpop` is destructive — there's no putting it back), and nothing catches the exception. It propagates out of `check_queue()` while the manager's lock is held, so:

- the task is silently dropped — never retried, never logged
- any other request going through `check_queue()` at that moment crashes with an uncaught `ValidationError` instead of just seeing an empty/normal queue

## Fix

`dequeue()` now catches `ValidationError` per entry, logs it, and moves on to the next queued item instead of raising — "skip a task that can no longer run" instead of "corrupt the whole dequeue call". Since `dequeue()` can now exhaust the queue without finding a usable task and return `None` even when `is_queue_empty()` was `False` a moment earlier, `check_queue()` also explicitly handles that case instead of assuming `dequeue()` always returns a task dict.

`InMemoryTaskManager` is unaffected — it keeps live `VideoParams` objects in memory rather than round-tripping through JSON, so there's nothing to re-validate on dequeue.

## Testing

Added to `test/services/test_task_manager.py`:
- `test_dequeue_skips_task_that_fails_current_validation` — a stale (now-invalid) entry ahead of a valid one is skipped, not raised
- `test_dequeue_returns_none_when_every_queued_task_is_stale` — all-stale queue returns `None` cleanly
- `test_check_queue_handles_dequeue_returning_none` — `check_queue()` no longer assumes `dequeue()` returns a task dict

Verified these three tests fail with the pre-fix code (`ValidationError` propagating from `dequeue()`, then `TypeError: 'NoneType' object is not subscriptable` from `check_queue()`) and pass with the fix, by temporarily reverting just the two source files and re-running.

Full suite: `uv run --no-sync python -X utf8 -m pytest -q test` → 514 passed, 11 skipped, 0 failed (matches CI's own coverage command). `ruff check app cli.py main.py webui test` → all checks passed.

### DIFF
--- a/app/controllers/manager/base_manager.py
+++ b/app/controllers/manager/base_manager.py
@@ -70,6 +70,12 @@ class TaskManager:
                 and not self.is_queue_empty()
             ):
                 task_info = self.dequeue()
+                if task_info is None:
+                    # dequeue() may skip and discard queue entries that no longer
+                    # pass current validation (see RedisTaskManager.dequeue) and
+                    # return None once nothing usable is left, even though
+                    # is_queue_empty() was True a moment earlier.
+                    return
                 func = task_info["func"]
                 args = task_info.get("args", ())
                 kwargs = task_info.get("kwargs", {})
--- a/app/controllers/manager/base_manager.py
+++ b/app/controllers/manager/base_manager.py
@@ -74,7 +74,7 @@ class TaskManager:
                     # dequeue() may skip and discard queue entries that no longer
                     # pass current validation (see RedisTaskManager.dequeue) and
                     # return None once nothing usable is left, even though
-                    # is_queue_empty() was True a moment earlier.
+                    # is_queue_empty() was False a moment earlier.
                     return
                 func = task_info["func"]
                 args = task_info.get("args", ())
--- a/app/controllers/manager/redis_manager.py
+++ b/app/controllers/manager/redis_manager.py
@@ -6,7 +6,9 @@ from loguru import logger
 from pydantic import ValidationError
 
 from app.controllers.manager.base_manager import TaskManager
+from app.models import const
 from app.models.schema import VideoParams
+from app.services import state as sm
 from app.services import task as tm
 
 FUNC_MAP = {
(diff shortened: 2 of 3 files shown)


harry0703:
Thanks for looking into this. Skipping stale entries and continuing to the next valid task is the right direction, and the test coverage is helpful.

There is one issue to address before merging: when an invalid entry is popped from Redis, its task state is left as `processing`. The state record is created before the task is enqueued, and the new exception handler only logs the validation error and continues. This can leave a task permanently shown as running in the API and WebUI.

Could you extract the `task_id` from `kwargs` and patch the existing task state to failed, including a clear failure stage and reason? Please avoid recreating the state if the task has already been deleted, and add a test covering this case.

There is also a small comment typo in `check_queue()`: it says `is_queue_empty()` was `True` a moment earlier, but this branch is entered after it returned `False`.

Once those are addressed, this should be ready to merge. Thanks again for the contribution.

kapil971390:
Thanks for the review — both addressed.

**Task state on discard**: `dequeue()` now extracts `task_id` from `kwargs` and calls `sm.state.patch_task(task_id, state=const.TASK_STATE_FAILED, failed_stage="dequeue", error=...)` when a stale entry is dropped. Used `patch_task` specifically (not `update_task`, which `_mark_task_failed` uses) since `patch_task` only updates an existing record and no-ops if the task was already deleted — `update_task` would have recreated it.

**Comment typo**: fixed — `check_queue()`'s branch is entered after `is_queue_empty()` returns `False`, not `True`.

**New tests**:
- `test_dequeue_marks_stale_task_failed_instead_of_leaving_it_processing` — verifies `patch_task` is called with `task_id`, `TASK_STATE_FAILED`, `failed_stage="dequeue"`, and an error message containing the validation detail
- `test_dequeue_does_not_recreate_state_for_already_deleted_task` — `patch_task` returning `False` (task already deleted) is treated as a no-op, not an error, and `update_task` is never called

Verified both fail against the pre-fix code (`patch_task` never called) and pass with the fix.

Full suite: 516 passed, 11 skipped, 0 failed. `ruff check` clean.

harry0703:
Thanks for the update. Stale tasks now move to a clear failed state without recreating deleted records, and the worker can continue with the next valid item. I reran the full test suite against the latest commit and all 516 tests passed, so this has been merged. Appreciate you taking care of the review feedback so quickly.

PR #1163, opened 2026-08-06
feat(subtitle): make whisper initial_prompt configurable

Blackspyda17:
## Summary

Allow biasing Whisper's decoder with a vocabulary prompt (brand names, technical terms) via a new `[whisper] initial_prompt` config key, passed as `initial_prompt` to `transcribe()`.

## Motivation

Whisper sometimes mis-transcribes brand names and technical terms. The existing `correct()` pass matches subtitles against the script, but the transcription itself benefits from a decoder bias prompt.

## Changes

- `app/services/subtitle.py`: read `config.whisper initial_prompt` at module load; pass it to `model.transcribe()` only when set (no behavior change when empty).
- `config.example.toml`: document the new key under `[whisper]`.

### DIFF
--- a/app/services/subtitle.py
+++ b/app/services/subtitle.py
@@ -15,6 +15,7 @@ from app.utils import utils
 model_size = config.whisper.get("model_size", "large-v3")
 device = config.whisper.get("device", "cpu")
 compute_type = config.whisper.get("compute_type", "int8")
+initial_prompt = config.whisper.get("initial_prompt", "") or None
 model = None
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the contribution. This is a useful, low-risk improvement for users who rely on Whisper subtitles with brand names or technical terms. I verified the parameter behavior with `faster-whisper 1.1.0` and ran the full test suite successfully. Merged.

PR #1166, opened 2026-08-08
feat: add optional LoomLoom batch generation backends

LewisZhouCC:
## Summary

Adds an optional LoomLoom integration while keeping the existing local/LLM and stock-footage workflows as defaults.

- Generate multiple script candidates through a configured Market Listing after quote confirmation
- Generate AI video material clips through a separate configured Listing, then reuse the existing narration, subtitle, and composition pipeline
- Keep the user API token in the browser session for the quote/run flow
- Include TemplateSpec examples, configuration documentation, and coverage for configuration, backend lifecycle, and WebUI confirmation behavior

The feature is disabled by default. No endpoint, token, Listing ID, deployment configuration, or provider-specific production resource is included.

## Validation

`uv run pytest -q test/services/test_loomloom.py test/services/test_webui_loomloom.py test/services/test_config.py`


harry0703:
Thanks for putting this together. I’ve incorporated the LoomLoom/Shengsuan Cloud integration into `main` and made a few follow-up adjustments to the WebUI flow, provider configuration, model selection, and test coverage.

Your authorship is preserved on the final squashed commit: c8cd2a0.

Since the completed version is now available on `main`, I’m closing this PR. Thanks again for the contribution.

PR #1172, opened 2026-08-11
fix(elevenlabs): stop wiping the saved API key on every restart

benjoooe:
## Problem

The ElevenLabs API key has to be re-entered on virtually every app load. It is saved, it works for the session, and then it is gone.

`_render_elevenlabs_api_key_input` in `webui/Main.py` persisted the widget value whenever a key was already stored:

```python
if configured_key or entered_key != effective_key:
    config.elevenlabs["api_key"] = entered_key
```

Once a key was saved, `configured_key` was truthy, so the assignment ran **unconditionally** — including when `entered_key` came back empty. Streamlit password inputs report an empty value when a session is replayed (for instance a still-open browser tab reconnecting after a server restart), so the empty string was written straight back over the stored key, and `save_config()` persisted the wipe.

### Reproduction

1. Put a valid key in `config.toml` under `[elevenlabs] api_key`, with the WebUI open in a browser tab.
2. Restart the containers.
3. Within seconds, `[elevenlabs] api_key` is `""`.

## Fix

An empty widget value can no longer overwrite a stored key. A replayed empty value cannot be distinguished from a deliberate clear by value alone, so the stored key wins — clearing it is done by editing `config.toml`.

Two related fixes are included:

- **`app/services/voice.py`** gained the `ELEVENLABS_API_KEY` fallback that `elevenlabs_music.get_api_key` already had, extracted as `get_elevenlabs_api_key()` to mirror that function. Without it, a deployment that set only the environment variable looked correctly configured in the WebUI but failed at generation time with `ElevenLabs API key is not set`.

- **The TTS branch** that primes the key for the Play Voice button no longer copies an environment-derived key into `config.toml`, so a secret kept in the environment stays out of the file the WebUI rewrites.

## Tests

Adds regression tests for key resolution, including one asserting that TTS and background music resolve identically.

Also pins the existing `test_elevenlabs_tts_no_api_key` to an empty environment: that test asserts behaviour when no key is configured, and with the new fallback its premise no longer holds on a host that exports `ELEVENLABS_API_KEY`. Verified it now passes both with and without that variable set.

Full suite: **509 passed, 11 skipped, 4035 subtests passed**.

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -252,6 +252,18 @@ def is_elevenlabs_voice(voice_name: str) -> bool:
     return (voice_name or "").startswith("elevenlabs:")
 
 
+def get_elevenlabs_api_key() -> str:
+    """
+    读取 ElevenLabs TTS 使用的 API Key。
+
+    与 elevenlabs_music.get_api_key 保持一致：环境变量仅作为本机配置未填写时的
+    后备来源。缺少该回退时，只设置 ELEVENLABS_API_KEY 的部署会在界面上看起来
+    正常，却在生成阶段失败并提示 API Key 未设置。
+    """
+    configured_key = str(config.elevenlabs.get("api_key", "") or "").strip()
+    return configured_key or os.getenv("ELEVENLABS_API_KEY", "").strip()
+
+
 def is_chatterbox_voice(voice_name: str) -> bool:
     return (voice_name or "").startswith("chatterbox:")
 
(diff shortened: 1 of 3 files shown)


harry0703:
Thanks for tracking this down and for the detailed explanation. I’ve fixed it on main in 01def3b with a little broader handling around reconnects and environment-based configuration.

The saved key is now restored before the voice list loads after a WebUI restart, ELEVENLABS_API_KEY remains supported without being copied into config.toml, and TTS uses the same key resolution as the music service. I also added regression tests and verified the flow against the real ElevenLabs voice list and speech synthesis.

I’m closing this PR since the fix is now available on main. Thanks again for pointing us in the right direction.

PR #1176, opened 2026-08-12
feat: add MiniMax music generation

octo-patch:
Reason: Add MiniMax music generation to the generated background music pipeline.

- Add a synchronous MiniMax music client for global and China endpoints with configurable models, lyrics, and audio settings.
- Decode and validate generated audio before atomically publishing background music output.
- Register the provider in task orchestration and persist its configuration.
- Cover regional requests, fallback settings, and invalid responses with focused tests.

Checks:
- `uvx --from ruff==0.15.21 ruff check app/services/minimax_music.py app/services/task.py app/config/config.py test/services/test_minimax_music.py`
- `uv run --no-project --with requests==2.33.1 --with loguru==0.7.3 --with toml==0.10.2 -- python -m unittest test.services.test_minimax_music -v` (3 passed)
- `python3 -m py_compile app/services/minimax_music.py app/services/task.py app/config/config.py test/services/test_minimax_music.py`
- `git diff --check`


harry0703:
Thanks for working on this. The MiniMax music integration is a useful direction, but I don't think the PR is ready to merge yet.

A few things still need to be addressed:

- `music-3.0` and `music-3.0-free` do not appear in the current MiniMax API documentation. The documented text-to-music models are `music-2.6` and `music-2.6-free`, so the current default may fail unless there is newer official documentation for it.
- `generate_bgm()` currently ignores `video_duration`. Provider-generated music is not looped later in the video pipeline, so a longer video may lose its background music before the video ends. The generated file should be trimmed or looped to match the target duration.
- There is no WebUI entry or configuration for MiniMax music yet, so regular WebUI users cannot select or configure it.
- The branch now conflicts with `main` after the MiniMax TTS changes were merged.

The focused tests pass, but they mock both the network response and audio validation. It would be helpful to add coverage for duration adaptation and the task integration as well.

Could you rebase the branch and update these parts? Once that is done, I'll take another look. Thanks again for the contribution.

harry0703:
Thank you again for the MiniMax music generation contribution. We requested changes on August 13 regarding the documented model IDs, duration handling, WebUI integration, tests, and the conflict with main, but the branch has not been updated since then. We are closing this PR to keep the queue current. If you revisit the implementation, please feel free to open a new PR based on the latest main and we will be happy to review it.

PR #1182, opened 2026-08-17
Enhance Next.js video creation frontend and UI features

HarshadHindlekar:


harry0703:
Thank you for the effort you put into the Next.js frontend. This PR introduces a second frontend stack together with API, dependency, and README changes across a very large scope, and it now conflicts with the current main branch. Maintaining a second WebUI would require a separate architecture discussion and a focused migration and testing plan, so we cannot merge this version. We are closing it for now, but a smaller proposal or design discussion would be welcome.

PR #1188, opened 2026-08-18
feat(ui): add id-ID to support_locales

randykaskuser:
Added Indonesian (`id-ID`) to the list of supported locales/languages in the WebUI.
This ensures the pipeline correctly routes Indonesian prompts for script generation and selects the appropriate TTS voices for voiceovers.

This PR has been split from the original submission to only include the language addition, per reviewer feedback.

### DIFF
--- a/webui/Main.py
+++ b/webui/Main.py
@@ -1408,6 +1408,7 @@ support_locales = [
     "en-US",
     "es-ES",
     "fr-FR",
+    "id-ID",
     "ru-RU",
     "vi-VN",
     "th-TH",


harry0703:
Thanks for putting this together. The Indonesian language option and target-duration idea are both useful additions.

I tested the branch against the current codebase, and there are a few issues we need to address before merging:

- The target duration is not passed through the WebUI script-generation button, prompt preview, or the `/scripts` API, so it does not work in some common flows.
- Social metadata is enabled by default and starts three extra LLM requests for every task. A YouTube cross-post currently ends up generating metadata four times, which adds unnecessary cost and concurrency.
- The metadata work runs in an unmanaged daemon thread. The video task can finish before `social_metadata.json` is written, so the WebUI may never display the result.
- The new UI text is missing from the locale files.
- The related test run currently has two failures, and the branch also conflicts with the latest `main`.

The project already has a social-metadata API and generates YouTube metadata when cross-posting, so it would be better to reuse that path and make the additional metadata generation explicitly opt-in. Splitting these three features into smaller PRs would also make them much easier to review and merge safely.

Thanks again for the contribution. I would be happy to take another look after these points are addressed.

randykaskuser:
I understand the concerns. I've rebased and split this PR to only include the Indonesian language support (`id-ID`), which resolves the test failures, merge conflicts, and unmanaged thread issues. The Target Duration and Social Metadata features will be submitted in separate PRs as requested.

randykaskuser:
I've reviewed the feedback from the maintainer, and I've split out the `video duration` and `social metadata` parts from this PR. This PR has been updated to ONLY contain the addition of the `id-ID` locale to `support_locales`. 

This directly addresses the maintainer's feedback:
1. **Target duration not passed** - Removed from this PR.
2. **Social metadata cost** - Removed from this PR.
3. **Daemon thread issue** - Removed from this PR.
4. **Missing locale text** - The new text was from the removed features, so it is no longer an issue for this PR. (Note: `id-ID` locale support itself is just a config addition).
5. **Test failures and merge conflicts** - Tests are now passing and conflicts are resolved.
6. **Split into smaller PRs** - Done!

I'm working on the video duration feature in a separate branch (`feat/video-duration-only`) and will open a new PR for it once it's fully tested. I'll do the same for the social metadata feature.

PR #1192, opened 2026-08-18
fix(tasks): roll back state when scheduling fails

HaningZS:
## Summary

- roll back the initial task state if `task_manager.add_task()` fails before accepting the task
- preserve the existing `TaskQueueFullError` to HTTP 429 mapping while sharing the same cleanup boundary
- add a regression test that uses a real `MemoryState` and preserves the original scheduling exception

## Problem

`create_task()` persisted a default `processing` state before calling the scheduler. The queue-full path removed that state, but other scheduler failures (for example, `threading.Thread.start()` raising `RuntimeError: can't start new thread`) propagated without cleanup. The task never ran, yet `/tasks` continued to expose it as `processing` with 0% progress.

I searched existing issues and open PRs for stuck/ghost task state and scheduling failures and found no matching report. Because the defect and fix are both localized and fully described here, I did not open a separate issue solely to close it immediately with this PR.

## TDD evidence

Before the implementation, the new regression test failed because the actual task state remained present after the injected scheduler error. After the implementation, it verifies both invariants:

1. the original `RuntimeError` is still propagated;
2. `MemoryState.get_task(task_id)` returns `None`.

The cleanup is intentionally scoped to `add_task()` only. A task that was successfully started or queued is not deleted if later logging or response serialization fails.

## Tests

- `uv run pytest -q test/services/test_controller_video.py test/services/test_task_manager.py` — 35 passed, 22 subtests passed
- `python -m compileall -q app cli.py main.py webui test`
- `ruff check app cli.py main.py webui test`
- `python -X utf8 -m coverage run -m pytest -q test` — 584 passed, 11 skipped, 4374 subtests passed
- `python -m coverage report` — 75% branch coverage (70% required)
- `git diff --check`

## Risk

Low. Success paths are unchanged, and the existing queue-full regression test continues to require exactly one state deletion and HTTP 429. If the state backend itself is unavailable, cleanup can still fail; this PR does not attempt to introduce a distributed transaction.

### DIFF
--- a/app/controllers/v1/video.py
+++ b/app/controllers/v1/video.py
@@ -206,11 +206,19 @@ def create_task(
             "params": body.model_dump(),
         }
         sm.state.update_task(task_id)
-        task_manager.add_task(tm.start, task_id=task_id, params=body, stop_at=stop_at)
+        try:
+            task_manager.add_task(
+                tm.start, task_id=task_id, params=body, stop_at=stop_at
+            )
+        except Exception:
+            # 状态记录在调度前创建，默认标记为 processing。如果调度器没能
+            # 接管任务（例如线程启动失败或 Redis 队列不可用），必须回滚该
+            # 记录，否则 API 和 WebUI 会永久展示一个实际从未运行的任务。
+            sm.state.delete_task(task_id)
+            raise
         logger.success(f"Task created: {utils.to_json(task)}")
         return utils.get_response(200, task)
     except TaskQueueFullError as e:
-        sm.state.delete_task(task_id)
         logger.warning(
             f"reject task because queue is full, request_id: {request_id}, task_id: {task_id}"
         )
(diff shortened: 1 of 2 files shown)


HaningZS:
Additional Linux validation completed after opening this PR:

- commit under test: `17b7b3f524df0774d2bad4b242a62d2f9de1c8bd`
- Debian 10 host, CPU-only Docker run (no GPU mounted)
- Python 3.11.10, FastAPI 0.136.3, pytest 9.1.1
- `python -m compileall -q app cli.py main.py webui test` — passed
- focused Ruff check for the controller/task-manager change surface — passed
- `python -X utf8 -m pytest -q test/services/test_controller_video.py test/services/test_task_manager.py` — **35 passed, 22 subtests passed**

This is in addition to the full local run already listed in the PR body (584 passed, 11 skipped, 75% branch coverage). The server test used an isolated dependency directory and did not use a GPU.

harry0703:
Thanks for the careful fix and the regression test. I reproduced the failure locally by forcing the task thread to fail during startup: the current code left a 0% processing task behind, while this change removed it and preserved the original scheduling error.

The focused tests and the full test suite both passed here. Merged. Thanks again for the contribution.

PR #1195, opened 2026-08-19
fix: sync Gemini TTS voices with the official catalog

jxnding:
## What changed

- replace the stale 15-item Gemini TTS list with Google's documented 30 prebuilt voices
- show Google's official voice style descriptions instead of inferred gender labels
- preserve valid legacy selections such as `gemini:Charon-Male` by matching their base voice name
- link the WebUI provider help directly to Google AI Studio's Voice Library
- add regression coverage for the catalog, legacy parsing, and dispatch metadata stripping

## Why

The original Gemini integration hard-coded a partial voice list. It omitted 24 currently documented voices and included unsupported names such as Atlas, Thalia, Sage, Echo, Harmony, Lux, Nova, Vale, and Orion. Selecting one of those unsupported entries could fail only after sending a TTS request.

Current catalog: https://ai.google.dev/gemini-api/docs/speech-generation#voice-options

## Validation

- `python -m compileall -q app cli.py main.py webui test`
- `ruff check app cli.py main.py webui test`
- `python -X utf8 -m pytest -q test` — 592 passed, 11 skipped, 4,374 subtests passed
- live WebUI check confirmed the Gemini selector exposes 30 options, from `gemini:Zephyr-Bright` through `gemini:Sulafat-Warm`

A live synthesis request reached Gemini successfully but the configured test project had already exhausted its daily free-tier TTS quota, so Google returned HTTP 429 before audio generation.


harry0703:
Thanks for catching this and updating the catalog. I checked the voice list against Google's current documentation and tested the legacy voice mapping and request dispatch locally. The full test suite passed as well.

Merged. Thanks for the contribution!

PR #1198, opened 2026-08-19
fix(deps): bump python-multipart to patch 4 CVEs (0.0.27 -> 0.0.32)

aeonframework:
Automated dependency bump to address 4 disclosed CVEs, found via [osv-scanner](https://google.github.io/osv-scanner/).

- **CVEs:** CVE-2026-53537, CVE-2026-53538, CVE-2026-53539 (max severity 7.5), CVE-2026-53540
- **Advisories:** GHSA-5rvq-cxj2-64vf, GHSA-6jv3-5f52-599m, GHSA-v9pg-7xvm-68hf, GHSA-vffw-93wf-4j4q
- **Package:** `python-multipart` `0.0.27` → `0.0.32` (fix lands at 0.0.31; bumped to the current latest release)

`python-multipart` is a direct, pinned dependency in both `requirements.txt` and `pyproject.toml`/`uv.lock` (used by FastAPI for multipart form/file-upload parsing on the `/api/v1/video/*` upload endpoints), so it's exercised on every deployment, not just a transitive path.

Confirmed against the OSV API that `0.0.31`/`0.0.32` clear all four advisories (queried both versions directly — zero remaining `vulns`), so this is a full remediation, not a partial one. `uv.lock` was regenerated with `uv lock` after bumping the pin; the lockfile diff also picks up an `upload-time` metadata field on every entry, which is a `uv` lockfile-schema revision bump (1 → 3) from the current `uv` release used to regenerate it, not a dependency change — the only `version = ` line that changed is `python-multipart`.

No code changes outside the lockfile/manifest.

---
Filed by [Aeon](https://github.com/aeonfun/aeon).


harry0703:
Thanks for flagging this and for preparing the update. The dependency bump itself makes sense, but this PR also rewrites the entire uv.lock file because of the lockfile schema change, which makes the diff much larger than the actual dependency update. I’d prefer not to merge that amount of unrelated lockfile churn, so I’ll handle the python-multipart upgrade separately with a more focused lockfile update. Thanks again for the contribution.

Issue #1196, opened 2026-08-19
[Feature]: 希望多一些素材api渠道


a18660223318-bit:
### 需求描述 | Problem Statement

目前作者提供的前三个可以设置的api 网站已经崩溃，无法注册，尤其是第一个和第二个网站

### 建议的解决方案 | Proposed Solution

多一些素材的api支持网站

abhishekChauhanx:
@harry0703 Can i take this issue?

Uszunso:
مر

harry0703:
@abhishekChauhanx Thanks for your interest. Yes, you’re welcome to work on this.

Before starting, please let me know which media provider you plan to add, along with its API documentation and licensing terms. I’d like to confirm the scope first so the implementation stays focused and fits the existing provider structure.

abhishekChauhanx:
@harry0703 Hi, thanks for the quick response!

I'd like to add the unsplash API as a new material source.

API docs: https://unsplash.com/documentation
Registration: https://unsplash.com/developers — free, self-serve, instant key (same flow as Pexels/Pixabay, no sales approval needed)

Why Unsplash:
1. Fills a gap — Pexels, Pixabay, and Coverr are all video-first, while Unsplash is a large, high-quality image library. Useful as a fallback when video search comes up empty for niche keywords, or for still-image needs.
2. Simple key-based auth, well-documented REST endpoints — fits the existing provider pattern in material.py (a search_photos_unsplash() function alongside the current search_videos_* ones, registered the same way).
3. Free tier is generous enough for this use case (50 requests/hour on demo keys, higher on approved production keys).
4. 
Licensing note to flag: Unsplash images are free for commercial and non-commercial use, no permission needed. However, their API Terms require attribution when the image is sourced via API — the app would need to show "Photo by [name] on Unsplash" with a link back to the photographer's profile wherever that image appears. I want to confirm this is acceptable before I start, since it's a small UI/output requirement the other three providers don't have.

Let me know if this fits what you had in mind, or if you'd prefer I look at a video-only alternative instead. Happy to start once confirmed.

abhishekChauhanx:
Thanks for the heads up — saw that @jxnding opened #_1206_ which adds configurable fallback/fan-out strategies across the existing providers. That's a great complement to this.

To avoid duplicate work, wanted to check on sequencing:

[...]

PR #1206, opened 2026-08-20
feat: add configurable stock provider strategies

jxnding:
## Summary

- add locked, ordered fallback, and fan-out modes for Pexels, Pixabay, and Coverr
- streamline the WebUI so **Video Source** selects only the source type and **Stock Providers** is the single authoritative stock-provider selector
- never implicitly prepend a provider; the legacy `video_source` field only mirrors the first explicit provider for API compatibility
- show one inline credential card for every selected provider whose API key is missing, including its password field and signup link
- distinguish provider errors from valid zero-result responses and report per-provider outcomes
- preflight selected credentials before script/TTS work and block invalid WebUI submissions
- merge and deduplicate fan-out results deterministically in user-priority order
- warn that fan-out consumes search quota from every selected provider

## WebUI behavior

- **Locked** allows one explicitly selected stock provider
- **Fallback** and **Fan-out** allow an ordered provider selection
- selecting multiple unconfigured providers displays multiple setup cards directly below the selector
- WaveSpeed, LoomLoom, and local-file sources remain separate and do not show stock-provider controls

## Compatibility

- legacy `pexels`, `pixabay`, and `coverr` values map to locked mode
- `pexels_pixabay` maps to fan-out in Pexels → Pixabay order
- existing task/API payloads retain `video_source` compatibility
- no credentials or configuration snapshots are persisted

## Validation

- focused provider, pipeline, and WebUI checks: 245 passed, 3 skipped, 72 subtests
- complete suite on current upstream: 799 passed, 11 skipped, 7,279 subtests
- Ruff, `git diff --check`, localization JSON parsing, and credential/path scans passed

Related to #1196.

## Follow-ups

This is part 1 of a staged series. Follow-on branches cover exact model provenance, cooperative cancellation, artifact reuse, font controls, and subtitle controls. They will be rebased and opened after preceding parts land so each upstream diff stays focused.


harry0703:
Thank you for this thoughtful contribution. The multi-provider strategies are useful, and the WebUI works well overall.

During testing, I noticed that the current WebUI controls are somewhat repetitive and may be confusing. The **Video Source** selector already chooses Pexels, Pixabay, or Coverr, while **Stock Providers** presents the same providers again. Even when the latter appears empty, the selected Video Source is still implicitly prepended and used. In locked mode, the Stock Providers selector is therefore redundant.

Could you please streamline this part of the UI? One practical approach would be to treat Video Source as the primary provider, hide the additional provider selector in locked mode, and label it as fallback or additional parallel providers in the other modes. Excluding the already selected primary provider from that list would also make the relationship clearer.

Thanks again for the work. Please let us know once this has been adjusted, and we will be happy to review it again.

jxnding:
Thanks for the clear feedback — updated.

I removed the duplicated Pexels/Pixabay/Coverr choices from **Video Source**. It now selects only the source type, while **Stock Providers** is the single authoritative selector for stock footage in every strategy.

- **Locked** allows one provider.
- **Fallback** and **Fan-out** allow ordered multiple providers.
- No provider is implicitly prepended; the legacy `video_source` field only mirrors the first explicit choice for backward compatibility.
- Every selected provider without a configured key gets its own inline setup card with a password field and signup link.
- Existing `pexels_pixabay` settings still restore as Pexels → Pixabay fan-out.
- WaveSpeed, LoomLoom, and local-file behavior remains separate.

I also rebased the one-commit change onto current `main`. The local complete suite is green (799 passed, 11 skipped, 7,279 subtests), and all upstream checks now pass on Python 3.11, Python 3.13, and Windows.

jxnding:
image
image

heres some pictures of it. I noticed that when selecting the stock provider the first time it prompts you to put an API key, so i added that feature as multiple cards below. On a new installation, you can select as many providers as you want and it will make a list of API entry fields

harry0703:
Thank you for the update. The duplicated WebUI controls are now resolved, and the new flow is much clearer.

I ran real end-to-end WebUI tests using the topic “How AI is changing people’s lives,” covering locked Pexels, Pixabay, and Coverr; normal and forced fallback; and two- and three-provider fan-out. All core scenarios produced playable H.264/AAC videos. Forced fallback also correctly recorded Pexels authentication failures and continued with Pixabay.

I found two remaining issues:

1. `--video-source pexels_pixabay` is exposed as a CLI choice, but batch manifest validation rejects it with exit code 2.
2. For a locked API/CLI payload containing multiple `material_providers`, preflight validates every provider key before locked mode reduces the list to the first provider. An unused provider without a key can therefore block the task.

Could you please address these two cases? Thanks again for the thoughtful work. We will be happy to review the update.

PR #1208, opened 2026-08-20
Create python-publish.yml

thaodex:


harry0703:
Thank you for taking the time to contribute this workflow. At the moment, MoneyPrinterTurbo is distributed as an application rather than a supported PyPI package. The wheel currently omits required configuration and runtime assets and does not provide a usable installed entry point. The moneyprinterturbo project name on PyPI is also already owned and published elsewhere, so this repository cannot safely enable this workflow without first resolving package ownership, naming, and the packaging contract.

We are going to close this PR for now. If official PyPI distribution is added in the future, we will first define and test the package layout and release process. Thanks again for the contribution and for helping improve the project.

PR #1209, opened 2026-08-20
feat(ui): add Auto-Publish (Upload-Post) settings tab to WebUI

### LINKED ISSUE #1204: [Feature Request] Add Upload-Post (Auto-Publish to YouTube/TikTok/Instagram) UI controls in WebUI Settings
### Feature Description

Currently, the backend integration for `upload_post` (`app/services/upload_post.py` and `app/services/task.py`) is implemented and functions via `config.toml`.

However, in `webui/Main.py`, there is currently no graphical UI form or tab within the Settings dialog (`_render_settings_dialog`) to configure:
- `upload_post_enabled` (toggle switch)
- `upload_post_api_key` (text input)
- `upload_post_username` (text input)
- `upload_post_platforms` (multiselect: YouTube Shorts, TikTok, Instagram)
- `upload_post_auto_upload` (toggle switch)
- `upload_post_youtube_privacy_status` (selectbox: public, private, unlisted)

### Proposed Solution

Add a dedicated **"Auto-Publish (Upload-Post)"** tab or section in the WebUI Settings dialog so that creators can configure and enable auto cross-posting directly from the browser without manually editing `config.toml`.

Thank you for maintaining this fantastic project! 🚀

parveen0029:
## Description
Fixes #1204 

This PR bridges the gap between the existing backend cross-post logic (`app/services/upload_post.py`) and the Frontend. Users can now seamlessly enable cross-posting to TikTok, Instagram, and YouTube Shorts directly from the Streamlit WebUI without having to manually edit `config.toml`.

### Key Changes
- Added an **"Auto-Publish Settings"** tab in the Streamlit UI (`webui/Main.py`).
- Mapped the **Enable Auto-Publish** checkbox to toggle both `upload_post_enabled` and `upload_post_auto_upload` ensuring the task pipeline natively triggers it.
- Added necessary API Key and Username inputs.
- Implemented logic to dynamically reveal the **YouTube Privacy Status** dropdown only if `youtube` is selected in the platform multiselect.
- **State Management**: Added unique `key` attributes to all new Streamlit widgets to prevent re-render glitches and state resets during interactions.

### Testing
- Tested locally. UI updates correctly persist settings live to the `config.toml` structure without restarting the server.


harry0703:
Thank you for the contribution and for wiring the existing Upload-Post integration into the WebUI.

I tested the flow and found a blocking issue: the new controls update `config.toml`, but the shared `upload_post_service` instance reads and caches `enabled`, `auto_upload`, credentials, and platforms when it is created at module import time. As a result, changing these settings in the WebUI does not update the service used by the running task pipeline. Enabling auto-publish will not take effect until the process is restarted, and disabling it can leave the existing service enabled for the current process.

Could you please make the running service reload or apply the updated configuration after the settings change, and add a test covering both enable and disable without restarting the application?

Thanks again for working on this feature.

parveen0029:
Hi @harry0703,

Thank you for the thorough review and for catching the module-load caching issue! That was a great catch. 

I have updated the PR to address both of your blocking points. Here is what was changed:

- [x] **Dynamic Configuration**: Refactored `UploadPostService` to use `@property` decorators instead of caching variables inside `__init__`. The service now dynamically reads the latest values (e.g., `config.app.get("upload_post_enabled")`) on demand. This ensures that any setting changes made in the WebUI take effect immediately for the running pipeline, without requiring a process restart.
- [x] **Unit Testing**: Added `TestUploadPostServiceDynamicConfig` to `test/services/test_upload_post.py`. It explicitly tests and verifies that the service successfully handles enabling, updating credentials, and disabling on the fly without requiring re-instantiation.

The branch has been updated with these fixes. Let me know if everything looks good to merge now!

harry0703:
Hi @parveen0029,

Thank you for the update and for addressing the dynamic configuration issue. I retested the latest branch, including a read-only check against a real Upload-Post account, and found a few remaining issues:

1. The WebUI currently asks for "Upload-Post Username (Email)" and tells users to enter their registered email. However, the Upload-Post API expects the publishing profile username in the user field, not the account email. In the real API check, the profile username lookup returned 200 while the account email lookup returned 404. Since upload_video sends this value directly as user, following the current UI guidance can cause real uploads to fail. Please update the label/help text and preferably validate or load the profile username from the API.
2. The existing task tests now have 3 failures because they patch service attributes such as auto_upload, which became read-only properties after this change.
3. The i18n completeness tests have 8 failures because the new translation keys are missing from the locale files.
4. The Enable Auto-Publish checkbox only updates upload_post_auto_upload when upload_post_enabled changes. If the saved configuration has enabled=true and auto_upload=false, the checkbox appears enabled while the task pipeline still does not auto-publish.

Could you please address these remaining points and update the related tests? I will be happy to review the next update.

Thanks again for working on this feature.

harry0703:
Hi @parveen0029,

Thank you for the latest update. I appreciate that you addressed the dynamic configuration issue, corrected the profile username guidance, updated the task tests, and added the missing i18n keys. I retested the branch, and its own test suite passes.

There are still a few issues to address before merging:

1. The branch currently conflicts with the latest main in all nine locale JSON files. The latest commit also reformats the complete locale files, producing several thousand unrelated changed lines and making the conflicts much larger. Please rebase onto the latest main, preserve the existing JSON formatting, and add only the new translation keys.

2. The enabled/auto-upload mismatch is still reproducible. With upload_post_enabled=true and upload_post_auto_upload=false, the checkbox appears enabled, but auto_upload remains false and the task pipeline will not publish. The current condition only synchronizes auto_upload when upload_post_enabled itself changes. Please synchronize both values when either one differs from the checkbox state and add an AppTest for this persisted mismatch.

3. An unsupported upload_post_youtube_privacy_status value causes the settings dialog to raise ValueError because list.index() is called directly. Please validate the saved value and fall back to public when it is not one of the supported options.

4. After rebasing, please integrate this feature with the key backup functionality now on main. upload_post_api_key is detected automatically, but the required upload_post_username should be included as a companion field so a restored backup is usable on another installation.

For reference, the current branch passes Ruff and its full suite with 599 passed and 11 skipped. However, it cannot yet be validated as merged with the latest main because of the locale conflicts.

Thanks again for continuing to improve this feature. We will be happy to review the next update.

harry0703:
Hi @parveen0029,

Thank you for the latest update. I verified that the persisted enabled/auto-upload mismatch is now synchronized, an invalid YouTube privacy value falls back to public, the locale changes are minimal, and upload_post_username is included with the API-key backup. The branch also no longer shows the earlier locale conflicts against current main.

There are two remaining blockers before merging:

1. The targeted suite reports 1 failed, 79 passed, and 3 skipped. test_key_backup_companion_keys_stay_in_sync_with_the_provider_registry fails because upload_post_username was appended directly to CREDENTIAL_COMPANION_KEYS["app"], while that invariant intentionally checks that the collection exactly matches the LLM Provider Registry extra fields. Please model non-LLM companion fields separately, or update the structure and its invariant test explicitly so Registry synchronization remains protected.
2. Ruff fails because pytest is imported but unused in test_webui_upload_post_settings.py.

Please also add focused assertions for restoring upload_post_api_key together with upload_post_username and for the invalid YouTube privacy fallback, so those two fixes cannot regress. The functional direction now looks good, and the remaining repair cost should be small.

Thank you again for working through the review feedback. I will be happy to retest the next update.

harry0703:
Hi @parveen0029, thank you for the latest update. I verified that the non-LLM companion-key separation, explicit Upload-Post key/username restore assertions, and invalid YouTube privacy AppTest are now present. The functional fixes are in the right direction. I applied the current PR patch to the latest main and reran the focused suite. It currently reports 16 failures, 96 passes, 3 skips, and 5941 passing subtests; Ruff also fails. The remaining issues are small but still block merging: 1. test_webui_settings_transfer.py contains duplicate upload_post_api_key and upload_post_username dictionary keys, with two incorrectly indented copies. Ruff reports both as F601. Please keep only one correctly indented pair in each fixture. 2. French is now a supported secondary locale on main, but fr.json is missing all seven new Auto-Publish strings, causing the i18n failures and KeyErrors. Please add the French entries as well. 3. The other secondary locale files currently copy the seven English values verbatim. These are normal user-facing settings rather than shared provider guidance, so please provide the corresponding translations instead of permanent English duplicates. 4. The patch adds trailing whitespace on multiple lines; please run Ruff and git diff --check before pushing. After these focused cleanup changes, I can rerun the suite promptly. Thank you again for working through the review feedback.

parveen0029:
Hi @harry0703, thanks for your patience! I have addressed all the remaining cleanup items:

Fixed the duplicate dictionary keys (F601) and indentation in test_webui_settings_transfer.py.
Added the missing French translations to fr.json.
Provided proper localized translations for all the 7 new UI strings across all secondary locales (es, de, pt, ru, tr, etc.) instead of duplicating the English ones.
Stripped all trailing whitespaces across the touched files.
Everything now passes ruff and the i18n completeness checks locally on top of the latest main. Let me know if everything looks good to merge now!

harry0703:
Thank you for the contribution and for addressing all the review feedback. I completed a real end-to-end test with a private YouTube upload, and the workflow worked as expected. This has now been merged. Thanks again!

PR #1211, opened 2026-08-20
feat(webui): export/import settings presets and back up API keys

rignaneseleo:
## What this adds

Two ways to move WebUI configuration between machines, reusing the existing configuration model instead of introducing a new one.

### 1. Settings preset — export / import the four settings panels

A **Settings Preset** expander above the *Generate Video* button:

- **Export Settings** downloads the current *Video Script*, *Video*, *Audio* and *Subtitle* panels as `moneyprinterturbo-settings.json`.
- **Import Settings** takes that file back and fills every control.

Import reuses the existing task-restore path: `_apply_pending_task_restore` was split into `_apply_restored_params`, so "regenerate a history task" and "import a preset" write the same `VideoParams` model into the widget state, and a new field only has to be handled once.

Local material, custom voiceover and custom background-music paths are **not** exported. A preset is meant to travel to another machine or container, where those paths do not exist. When a preset selects the `local` video source, the existing "please upload materials again" warning applies unchanged.

### 2. Key backup — save and restore API keys

A **Key Backup** tab in the settings dialog:

- **Export Keys** downloads every API key stored in the runtime configuration as `moneyprinterturbo-keys.json`.
- **Import Keys** restores them, so a reinstall, a second container or a new machine does not mean retyping every provider key.

Keys are recognised from the existing config-key naming (`*_api_key`, `*_api_keys`, `*_api_token`, `*_access_key`, `*_secret_key`, `speech_key`) across the runtime config sections. A newly added provider that follows the naming is covered without maintaining a second list. Settings a key is useless without — currently the Azure speech region — travel with it. The `ui` section is skipped entirely.

## Safety notes

- Both files declare a `schema` and a `version`, and import rejects anything that does not match, so an unrelated JSON, a task record or a `config.toml` cannot reach the configuration.
- Key import ignores unknown sections and any config item that is not a key, so the import entry cannot be used to rewrite unrelated settings.
- Presets are validated through `VideoParams`, so an edited or foreign preset fails at import instead of failing later during generation.
- Restoring keys also clears the matching password-widget state. Without that, Streamlit keeps showing the old key and writes it back on the next rerun, and the restore looks like it did nothing.
- The exported key file is plain text. The tab shows a warning about it, and the file is never written by the app itself — it only goes through the browser download.

## Tests

- New `test/services/test_webui_settings_transfer.py` (13 tests), following the existing AST-extraction pattern from `test_webui_task_history.py`: preset round-trip, exclusion of local paths, schema/version/validation rejection, key collection and restore, unknown-section filtering, UTF-8 BOM tolerance, and widget-key mapping.
- New UI strings are translated in all nine locales, so the existing i18n coverage tests stay green.
- `pytest -q test` → 611 passed, 11 skipped. `ruff check app cli.py main.py webui test` → clean.
- Verified end to end with `streamlit.testing.v1.AppTest`: the page renders without exceptions, importing a preset applies the subject, font size, outline width and subtitle toggle, and the settings dialog shows the new tab with both entries.

## Screenshots

None — verified headlessly via `AppTest` on a server without a browser.


harry0703:
Hi @rignaneseleo, thank you for this thoughtful contribution. The settings preset and key backup features are useful, the UI works well in manual testing, and the full test suite passes.

Before merging, could you please address two issues we found?

1. Key restore clears only the widget key derived from the config name. Some credentials also use different widget keys, including `loomloom_user_api_token`, `gemini_tts_api_key_input`, and `mimo_tts_api_key_input`. Those old session-state values remain after an import and can overwrite the newly restored credentials on the next rerun. Please clear all widget aliases for each restored credential and add a test covering the actual `_apply_key_backup` path.
2. A Cloudflare AI Gateway backup includes `cloudflare_api_key`, but not the required `cloudflare_account_id` or a customized `cloudflare_gateway_id`. After restoring on another machine, that provider may still be unusable. Please include these companion fields and add a round-trip test.

We will keep the PR open and look forward to your update. Thank you again for the detailed implementation and tests!

rignaneseleo:
Thanks for the careful review — both issues were real, and both are fixed in `6feeb75`.

1. **Widget aliases.** `_credential_widget_state_key` is now `_credential_widget_state_keys` and returns the key derived from the config name plus every alias, so `gemini_tts_api_key_input`, `mimo_tts_api_key_input` and `loomloom_user_api_token` are cleared together with `gemini_api_key_input`, `mimo_api_key_input` and `loomloom_api_token_input`. Added `test_apply_key_backup_writes_config_and_clears_every_widget_alias`, which drives the real `_apply_key_backup` with a stub `session_state` and asserts both the config writes and that no alias survives.
2. **Cloudflare companion fields.** Companion keys now come from `LLM_PROVIDER_REGISTRY` extra fields instead of a hardcoded list, so `cloudflare_account_id` and `cloudflare_gateway_id` travel with `cloudflare_api_key`, and a provider added later is covered without another change in `Main.py`. Added a Cloudflare round-trip test plus a test that keeps the companion list in sync with the registry.

Full suite: 615 passed, 11 skipped. `ruff check app cli.py main.py webui test` is clean, and the page still renders without exceptions under `AppTest`.

harry0703:
Thank you, @rignaneseleo, for the thoughtful implementation and for addressing the review feedback so thoroughly. I verified the updated PR end to end with a real configuration: settings preset and key backup export/import work correctly, invalid files are rejected, the WebUI behaves as expected, and the full test suite passes. This has now been merged. Thanks again for your contribution!

PR #1212, opened 2026-08-20
fix(controllers): register ping health check router

NafeesMadni:
This PR registers the existing `ping` health-check router with the root API router so the FastAPI application properly exposes the endpoint.

The test follows the repository's existing controller unit-test style, directly testing the controller function and verifying that it returns the expected `"pong"` response.

### Changes

* Register `ping.router` in `root_api_router`
* Add a unit test for the ping controller
* Follow the existing controller testing pattern used by other endpoints

### DIFF
--- a/app/router.py
+++ b/app/router.py
@@ -1,4 +1,4 @@
-"""Application configuration - root APIRouter.
+"""Application configuration - root APIRouter
 
 Defines all FastAPI application endpoints.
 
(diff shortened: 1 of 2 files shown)


harry0703:
Hi @NafeesMadni, thank you for the contribution! Registering the existing ping router makes `/ping` available as intended and provides a useful health-check endpoint. We verified that the endpoint returns `200` with `"pong"`, and the full test suite passes.

I am merging this now. Thank you again for helping improve the project!

PR #1213, opened 2026-08-20
feat(i18n): add Italian (it) WebUI translation

SandroHub013:
Adds Italian to the WebUI language list. There is currently no `it.json`, so `resolve_ui_language()` falls an Italian browser through to English on first visit; with this file it resolves to `it`.

Three files, following the shape of the existing locale contributions:

| File | Change |
|---|---|
| `webui/i18n/it.json` | new locale, all 296 required keys |
| `webui/Main.py` | `"it-IT"` added to `support_locales` |
| `test/services/test_webui_i18n.py` | `"it"` added to `SECONDARY_LOCALES` |

## Why the test file is touched

`SECONDARY_LOCALES` is a hardcoded tuple, so a new locale file is not covered by the suite until it is listed there. Adding `"it"` puts this translation under all seven checks that already guard the other secondary locales:

- every key of `en.json` is present, minus the provider-tips prefixes and `ENGLISH_FALLBACK_KEYS`
- no `llm_provider_tips.*` or `tts_provider_tips.*` key is duplicated
- no `ENGLISH_FALLBACK_KEYS` entry is duplicated
- every `tr()` key parsed out of `webui/Main.py` is covered
- format placeholders match `en.json` key by key
- markdown link targets match `en.json` key by key

I would rather the CI catch a future drift in this file than have it silently rot.

## Translation notes

- **296 keys**, exactly the set the seven shipped secondary locales carry. Provider configuration tips and the `ENGLISH_FALLBACK_KEYS` entries are deliberately left to fall back to English, as the comment in the test file asks.
- **Conventions**: imperative for actions (`Genera video`, `Salva`, `Riprova`), impersonal for help text, Italian sentence case rather than English title case.
- **Consistent terminology** throughout: `sottotitoli`, `voce fuori campo`, `attività`, `materiale`, `anteprima`, `musica di sottofondo`, `chiave API`.
- **Localised decimal separator**: `1,0×` in the voiceover-speed help, matching what `es.json` and `pt.json` already do.
- **Not translated**: product and model names, config keys, file extensions, and anything inside backticks. `Base URL` is kept as-is since it is the literal field label.
- Two dropdown entries are spelled out rather than translated word-for-word — `None` → `Nessuna transizione` and `Shuffle` → `Transizione casuale` — because a bare adjective is ambiguous in that list. `pt`, `de` and `ru` already do the same.

## Verification

```
$ python -m pytest test/services/test_llm.py test/services/test_webui_i18n.py -q
2 failed, 87 passed, 1 skipped, 4770 subtests passed
```

The two failures are pre-existing and unrelated — `test_gemini_uses_google_genai_client` and `test_social_metadata_endpoint_response_shape` fail identically on a clean checkout of `main` in this environment, where the optional dependencies they need are not installed. `test_webui_i18n.py` on its own is fully green:

```
$ python -m pytest test/services/test_webui_i18n.py -q
12 passed, 4770 subtests passed
```

Happy to adjust any wording — and if you would rather keep `SECONDARY_LOCALES` untouched in this PR, say so and I will drop that line.


harry0703:
Thank you for adding the Italian WebUI translation and for extending the i18n test coverage. I verified the locale resolution and translation tests, and everything looks good. I am merging this now—thanks again for your contribution!

PR #1214, opened 2026-08-20
fix(material): bypass Cloudflare's bot challenge on Pixabay searches

### LINKED ISSUE #1136: [Bug]: Pixabay video search always fails with 429 "Just a moment..." (Cloudflare challenge)
### 问题描述 | Current Behavior

When `video_source` is set to `"pixabay"`, every request to the Pixabay video
search API fails. The response is HTTP status `429` with an HTML body
containing `<title>Just a moment...</title>`, which is Cloudflare's
anti-bot challenge page, not a real Pixabay API error.

Log output:

​```
ERROR | "./app/services/material.py:163": search_videos_pixabay - search videos failed: Expecting value: line 1 column 1 (char 0)
INFO  | "./app/services/material.py:346": download_videos - found 0 videos for 'nature'
...
SUCCESS | "./app/services/material.py:382": download_videos - downloaded 0 videos
ERROR | "./app/services/task.py:206": _mark_task_failed - task failed, task_id: xxxx, stage: materials, error: failed to download video materials from pixabay
​```

### 重现步骤 | Steps to Reproduce

1. Deploy MoneyPrinterTurbo via Docker (`docker-compose.release.yml`, prebuilt
   image `ghcr.io/harry0703/moneyprinterturbo:latest`)
2. Configure a valid Pixabay API key in `config.toml`:
   `pixabay_api_keys = ["<valid-key>"]`
3. Call `POST /api/v1/videos` with `"video_source": "pixabay"`
4. Task fails at the materials stage

### 错误日志 | Error Logs

I confirmed the API key itself is valid — calling the same endpoint directly
with `curl` from the host machine (outside the container) returns a normal
JSON response with video results:

​```bash
curl -s "https://pixabay.com/api/videos/?key=<valid-key>&q=nature"
# -> returns valid JSON with "hits": [...]
​```

But calling it from inside the container using Python's `requests` library
(as `app/services/material.py` does) always returns the Cloudflare challenge
page, even after explicitly setting a browser-like `User-Agent` header:

​```python
import requests
headers = {'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36'}
r = requests.get('https://pixabay.com/api/videos/', params={'key': '<valid-key>', 'q': 'nature'}, headers=headers)
print(r.status_code)  # -> 429
print(r.text[:200])   # -> "<!DOCTYPE html>...<title>Just a moment...</title>..."
​```

This suggests Cloudflare is fingerprinting the TLS/HTTP client itself (not
just the `User-Agent` header), since `curl` passes through fine but Python's
`requests` library does not, regardless of headers set.

Python 3.11.13

### 操作系统 | Operating System

Debian (Docker container, host: Debian 12)

v1.3.2

### Current Behavior
When `video_source` is set to `"pixabay"`, every request to the Pixabay video
search API fails with HTTP 429 and Cloudflare's "Just a moment..." challenge
page instead of a real API response.

### Steps to Reproduce
1. Deploy via Docker (`docker-compose.release.yml`, image `ghcr.io/harry0703/moneyprinterturbo:latest`)
2. Configure a valid Pixabay API key in `config.toml`
3. Call `POST /api/v1/videos` with `"video_source": "pixabay"`
4. Task fails at materials stage

Confirmed the API key is valid — calling it directly via `curl` from the host
returns a normal JSON response with video results. But calling it from inside
the container via Python's `requests` library (as `search_videos_pixabay`
does) always returns Cloudflare's challenge page, even with a browser-like
`User-Agent` header set explicitly. This suggests Cloudflare is fingerprinting
the TLS/HTTP client itself, not just headers.

### Expected Behavior
Pixabay search should succeed with a valid key, consistent with Pexels and
Coverr, which work correctly in the same environment.

### Log
ERROR | "./app/services/material.py:163": search_videos_pixabay - search videos failed: Expecting value: line 1 column 1 (char 0)
INFO  | "./app/services/material.py:346": download_videos - found 0 videos for 'nature'
SUCCESS | "./app/services/material.py:382": download_videos - downloaded 0 videos
ERROR | "./app/services/task.py:206": _mark_task_failed - task failed, stage: materials, error: failed to download video materials from pixabay

### Possible Fix Direction
Switching the HTTP client used for Pixabay requests to something that better
emulates a real browser TLS fingerprint (e.g. `curl_cffi`, `cloudscraper`),
since header spoofing alone does not bypass the Cloudflare challenge.

korjavin:
## Summary

Pixabay's video search API sits behind Cloudflare, which fingerprints the TLS/HTTP client itself and blocks every request made with Python's `requests` library with a `429` "Just a moment..." challenge page — even with a valid API key and a browser-like `User-Agent` header. The same request via `curl` from the same host passes through fine, which confirms this is a TLS/JA3 client fingerprint block, not a bad key, IP ban, or rate limit (as diagnosed in the issue).

## Fix

Switch the Pixabay search request in `search_videos_pixabay()` (`app/services/material.py`) to [`curl_cffi`](https://github.com/lexiforest/curl_cffi), which impersonates a real browser's TLS handshake (`impersonate="chrome"`), while leaving Pexels and Coverr on plain `requests` since they aren't affected by this block. `curl_cffi`'s `requests`-compatible API (`proxies`, `verify`, `timeout` args, `Response.status_code/.headers/.text/.json()`) meant the existing error-handling logic (Cloudflare-challenge detection, rate-limit / non-JSON / redaction handling) needed no changes beyond swapping the call site.

Added `curl_cffi==0.16.0` to `pyproject.toml` / `requirements.txt` (ships prebuilt wheels for Linux/macOS/Windows, no extra system build deps). `uv.lock` was regenerated with `uv sync` — note this also bumped the lock file's schema revision (1 → 3, adding `upload-time` metadata to every existing entry) as a side effect of using a newer local `uv`; the actual dependency-graph change is just the one new `curl-cffi` package plus its existing transitive deps (`cffi`/`pycparser`, already present in the lock from other packages). `uv sync --frozen` passes cleanly against the regenerated lock.

## Verification

Live end-to-end check against the real Pixabay API with an intentionally invalid key, before vs. after:

- Before (plain `requests`): `429`, Cloudflare's HTML "Just a moment..." challenge page.
- After (`curl_cffi`, `impersonate="chrome"`): `400`, Pixabay's own `Invalid or missing API key` response — confirming the request now reaches Pixabay's actual API instead of being blocked upstream by Cloudflare.

## Tests

Updated the 9 `search_videos_pixabay`-specific test cases in `test/services/test_material.py` to patch `app.services.material.cf_requests.get` instead of `app.services.material.requests.get` (Pexels/Coverr tests are untouched, still patching plain `requests.get`).

- `uv run pytest test/services/test_material.py -q` — 24 passed, 5 subtests passed.
- `uv run pytest test/ -q` (full suite) — 598 passed, 11 skipped (pre-existing, unrelated), no regressions.
- `ruff check app/services/material.py test/services/test_material.py` — identical finding count (34, all pre-existing) before and after this change, confirmed via `git stash` comparison; no new lint issues introduced.

Fixes #1136


krzys167-crypto (approved):

harry0703:
Hi @korjavin, thank you for the detailed investigation and implementation. The test suite passes, and I appreciate that the existing proxy, TLS verification, timeout, and error-handling behavior has been preserved.

Before merging, could you please reconsider the use of `impersonate="chrome"`? This is explicitly intended to bypass a Cloudflare access challenge, while the Pixabay Terms of Service prohibit bypassing measures used to prevent or restrict access. This could create compliance and account-access risks for users of the project.

The challenge also appears to be specific to certain egress IPs, proxies, or network environments. We could not reproduce it here: both plain `requests` and `curl_cffi` reached the Pixabay API and returned the expected invalid-key response. Unconditionally switching every user to a native browser-impersonation dependency therefore seems too broad for an environment-specific issue.

Could you explore an approach that does not impersonate a browser, such as confirming an officially supported client with Pixabay, or making an approved workaround explicit and opt-in? Please also add a regression assertion for the selected client behavior, minimize the unrelated `uv.lock` metadata churn, and convert the new explanatory comment to Simplified Chinese to match the project convention.

We will keep the PR open and look forward to your update. Thank you again for the contribution!

korjavin:
Thanks for the thorough review — all fair points, addressed:

1. **Made it opt-in, not default.** Added `pixabay_bypass_cloudflare` (default `false`) in `config.toml`. Plain `requests` is now the default Pixabay client again, same as before this PR and same as Pexels/Coverr. `curl_cffi` with `impersonate="chrome"` only runs when a user explicitly sets the flag. Logged a warning on startup of that path noting the ToS conflict, so it's an informed, deliberate choice rather than a silent default for everyone.
2. **Environment-specific block, confirmed on my side too.** Re-tested just now from this sandbox with an invalid key: this time plain `requests` reached Pixabay's real API directly (400 "invalid key"), no Cloudflare challenge — consistent with your report that it's not universally reproducible. That's exactly why gating it behind opt-in rather than changing the default is the right call.
3. **Regression tests for client selection** — added `test_search_pixabay_uses_plain_requests_by_default` (asserts `requests.get` is called, `cf_requests.get` is not) and `test_search_pixabay_bypass_cloudflare_uses_browser_impersonation_when_enabled` (asserts the reverse, with `impersonate="chrome"`). The 9 pre-existing Pixabay tests are reverted to patching plain `requests.get` since that's the default again.
4. **uv.lock churn minimized.** The earlier `uv sync` (with a newer local `uv`) had bumped the lockfile's schema revision (1 → 3, adding `upload-time` metadata to every existing entry) as a side effect, touching ~4200 unrelated lines. Rebuilt it by hand against your committed lockfile with `uv 0.6.0` (which still writes `revision = 1`) so the diff is just the 34 lines the new `curl-cffi` dependency actually needs.
5. **Comment converted to Simplified Chinese** to match the file's convention.

Re-ran the full suite (`uv run pytest test/`) and `ruff check` — 600 passed / 11 skipped, no new lint findings (confirmed via diff against a clean checkout).

Regarding an officially-supported-client route: I couldn't find anything from Pixabay indicating their API has a different officially-blessed client library or documented workaround for this — happy to revisit if you're aware of one, but for now this keeps the existing default behavior fully intact and puts the workaround entirely in the hands of users who've confirmed they need it.

harry0703:
Hi @korjavin,

Thank you for the thorough update. I verified that plain `requests` is now the default client, browser impersonation is gated behind `pixabay_bypass_cloudflare=false` by default, the new client-selection tests pass, the lock file is valid, and the full PR now contains only the focused 34-line `uv.lock` addition.

There is one remaining concern: although the workaround is opt-in at runtime, `curl_cffi` is still an unconditional core dependency and is imported at module load time. In an existing environment without that package, importing `app.services.material` fails with `ModuleNotFoundError: curl_cffi` even when `pixabay_bypass_cloudflare` is disabled. I reproduced this before synchronizing the updated lock file.

Because this workaround is environment-specific and explicitly opt-in, could you please make the dependency optional and lazy-load it only when the flag is enabled, with a clear configuration error if the optional package is missing? That would preserve the existing default installation path while still supporting affected users who deliberately enable the workaround.

Thank you again for addressing the earlier review points so carefully.

korjavin:
Good catch, thank you — that's a real gap. Fixed:

- Moved `curl_cffi` out of the core dependencies entirely, into a new `pixabay-bypass` optional-dependency group (`uv sync --extra pixabay-bypass` / `pip install curl_cffi`), same pattern as the existing `twelvelabs` extra. It's also added to the `dev` dependency-group so CI/tests can still exercise both code paths without affecting the default install.
- Replaced the module-level `from curl_cffi import requests` with a `_get_cf_requests()` helper that imports it lazily, only inside the `pixabay_bypass_cloudflare` branch, and raises a clear error with install instructions if it's missing. `app.services.material` now imports cleanly with zero optional deps installed.
- That check now runs *before* the network try/except block (same as the existing `get_api_key()` check), so a missing optional dependency surfaces as a real, actionable error rather than being logged and swallowed as a generic network failure.
- Added a regression test for the missing-dependency error path, and updated the two client-selection tests to patch `curl_cffi.requests.get` directly (the lazy import's actual target) since there's no longer a persistent module-level attribute to patch.
- Rebuilt the `uv.lock` delta by hand again for the same reason as before — it's now 40 lines, covering the extra/dev-group restructuring.

Verified `app.services.material` imports and `search_videos_pixabay()` runs correctly with `curl_cffi` absent from the environment (default install), and full suite (601 passed) + `ruff check app cli.py main.py webui test` (your exact CI invocation) both pass.

harry0703:
Hi @korjavin,

Thank you again for the careful update. The lazy import and the actionable missing-dependency error address the runtime import problem correctly. I verified that `app.services.material` now imports successfully when `curl_cffi` is unavailable.

There are still two related issues with the optional dependency boundary:

1. In an existing environment without `curl_cffi`, running `test/services/test_material.py` results in **25 passed and 2 failed**. Both client-selection tests patch `curl_cffi.requests.get` directly, so the patch itself raises `ModuleNotFoundError` before the test can exercise the application code. The same file passes all 27 tests when `curl_cffi` is installed.
2. The project README recommends `uv sync --frozen` for the normal installation path, and the `dev` dependency group is installed by `uv sync` by default. Since `curl_cffi` is currently included in that group, users following the recommended uv setup will still install it even without selecting the `pixabay-bypass` extra.

Could you please remove `curl_cffi` from the default `dev` group and make the unit tests independent of the optional package? For example, the tests could patch `_get_cf_requests()` with a fake client and assert that it is not called on the default path. Any dedicated integration or CI job that needs the real client can install `--extra pixabay-bypass` explicitly.

Please make sure the material tests pass both without the extra installed and with the extra installed. Thank you for continuing to refine this carefully.

korjavin:
Both fixed:

1. **Tests no longer require curl_cffi to be importable.** Switched the two client-selection tests to patch `app.services.material._get_cf_requests` (the lazy-import helper itself) instead of `curl_cffi.requests.get`. The default-path test asserts `_get_cf_requests` is never called; the opt-in-path test replaces it with a fake client (`SimpleNamespace(get=MagicMock(...))`) and asserts that fake's `.get` was called with `impersonate="chrome"`. Neither needs `curl_cffi` present in the environment at all now.
2. **Removed curl_cffi from the `dev` dependency-group.** You're right that `uv sync --frozen` installs dependency-groups by default, so it was still landing on every standard install. It's now only pulled in via `uv sync --extra pixabay-bypass`.

Verified both configurations directly, not just asserted:
- `uv sync --frozen` (no extra) → `python -c "import curl_cffi"` fails with `ModuleNotFoundError` → `pytest test/services/test_material.py` → **27 passed**.
- `uv sync --frozen --extra pixabay-bypass` → `import curl_cffi` succeeds → same test file → **27 passed**.

Full suite (`uv run pytest test/`) also passes with the extra installed: 601 passed / 11 skipped. `ruff check app cli.py main.py webui test` clean.

harry0703:
Thanks for carefully addressing all the feedback. After reconsidering the maintenance and compliance trade-off, we don’t want to ship a browser-impersonation bypass for this environment-specific issue, even as an opt-in feature. We’ll close this for now, but we really appreciate the work you put into it.

PR #1217, opened 2026-08-20
fix(schema): improve response schemas and migrate to Pydantic v2 ConfigDict

### LINKED ISSUE #1218: Response models expose data as any in generated OpenAPI schema
### 问题描述 | Current Behavior

Several response models in `app/schemas.py` inherit the `data` field from `BaseResponse`:

```python
class BaseResponse(BaseModel):
    status: int = 200
    message: Optional[str] = "success"
    data: Any = None
```

Although individual response models provide concrete examples through `json_schema_extra`, their `data` fields are still represented as `any` in the generated OpenAPI schema.

For example, `TaskDeletionResponse` provides an example containing:

```json
{
  "data": {
    "state": 1,
    "progress": 100,
    "videos": [],
    "combined_videos": []
  }
}
```

However, Swagger/OpenAPI displays the response as:

```text
data    any
```

This makes the generated API documentation less descriptive than the actual API response.

### 重现步骤 | Steps to Reproduce

1. Start the MoneyPrinterTurbo application.
2. Open the generated Swagger/OpenAPI documentation.
3. Find a response model such as `TaskDeletionResponse`.
4. Expand the response schema.
5. Observe that `data` is displayed as `any`.
6. Compare it with the example, which contains specific fields such as `state`, `progress`, `videos`, and `combined_videos`.

### 错误日志 | Error Logs

No runtime error occurs.

The issue is with the generated OpenAPI schema and API documentation.

3.11

### 操作系统 | Operating System

Ubuntu 24.04

v1.3.4

A possible solution is to define dedicated data models and explicitly type the `data` field.

For example:

```python
class TaskDeletionData(BaseModel):
    state: int
    progress: int
    videos: Optional[List[str]]
    combined_videos: Optional[List[str]]


class TaskDeletionResponse(BaseResponse):
    data: TaskDeletionData
```

This would allow FastAPI/Pydantic to generate a schema containing the actual structure of `data` instead of `any`.

This approach can also be applied to other response models whose `data` structure is currently represented only through an example.

### LINKED ISSUE #1219: Migrate remaining class-based Pydantic Config definitions to ConfigDict
### 问题描述 | Current Behavior

Several Pydantic models in `app/schemas.py` still use the class-based `Config` syntax from Pydantic v1.

For example:

```python
class TaskDeletionResponse(BaseResponse):
    class Config:
        json_schema_extra = {
            "example": {
                ...
            }
        }
```

The project is using Pydantic v2, where class-based `Config` is deprecated in favor of `ConfigDict`.

### 重现步骤 | Steps to Reproduce

1. Install the project dependencies with Pydantic v2.
2. Run the project's test suite or import the affected schema models.
3. Observe the Pydantic deprecation warning.

### 错误日志 | Error Logs

```text
PydanticDeprecatedSince20: Support for class-based `config` is deprecated,
use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0.
See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.12/migration/
```

3.11

### 操作系统 | Operating System

Ubuntu 24.04

v1.3.4

Several response models still use the deprecated class-based `Config` syntax, including:

* `TaskResponse`
* `TaskDeletionResponse`
* `VideoScriptResponse`
* `VideoTermsResponse`
* `VideoSocialMetadataResponse`
* `BgmRetrieveResponse`
* `BgmUploadResponse`
* `VideoMaterialRetrieveResponse`
* `VideoMaterialUploadResponse`

These can be migrated to Pydantic v2 syntax:

```python
class TaskDeletionResponse(BaseResponse):
    model_config = ConfigDict(
        json_schema_extra={
            "example": {
                ...
            }
        }
    )
```

`MaterialInfo` can also use `ConfigDict` instead of the class-based configuration currently passed to the Pydantic dataclass.

Migrating these definitions would remove the deprecation warnings and make the schema definitions compatible with the recommended Pydantic v2 configuration style.

NafeesMadni:
This PR addresses two issues in `app/models/schema.py`:

#1218  — Response models expose data as any in the generated OpenAPI schema.

#1219  — Remaining class-based Pydantic Config definitions produce deprecation warnings under Pydantic v2.

The changes make response schemas more explicit and migrate the affected models to Pydantic v2's recommended configuration syntax.

## Changes

* Added dedicated response data models:

  * `TaskResponseData`
  * `TaskDeletionData`
  * `VideoScriptData`
  * `VideoTermsData`
  * `VideoSocialMetadataData`
  * `FileData`
  * `BgmRetrieveData`
  * `BgmUploadData`
  * `VideoMaterialRetrieveData`
  * `VideoMaterialUploadData`

* Updated response models to explicitly declare their `data` type instead of inheriting `data: Any` from `BaseResponse`.

* Migrated class-based Pydantic configuration to Pydantic v2's `ConfigDict`.

* Updated `MaterialInfo` to use `ConfigDict` with `arbitrary_types_allowed=True`.

### Before

Response models such as `TaskDeletionResponse` exposed `data` as `any` in the generated OpenAPI schema, even though the example contained a concrete object structure.

Screenshot 2026-08-20 233243

### After

The response now exposes the actual structure of `data`, including its fields and types.

Screenshot 2026-08-20 233401

This makes the generated API documentation more accurate and useful for API consumers.

## Why

Previously, several response models only documented their `data` structure through `json_schema_extra["example"]`.

As a result, the generated OpenAPI schema could show:

```text
data: any
```

even though the example contained a specific object structure.

The new dedicated data models allow FastAPI/Pydantic to generate an accurate schema for the actual response payload.

This change also removes the Pydantic v2 deprecation warning:

```text
PydanticDeprecatedSince20: Support for class-based `config` is deprecated,
use ConfigDict instead.
```

## Compatibility

The API response structure remains unchanged.

This change primarily:

* improves the generated OpenAPI documentation
* makes response contracts explicit
* removes deprecated Pydantic configuration usage
* improves type safety in the response models

## Testing

* Ran the existing test suite.
* Verified that the updated response models can be imported and constructed successfully.
* Verified the generated OpenAPI schemas for the updated response models.
* Confirmed that the affected class-based `Config` definitions have been migrated to `ConfigDict`.

Closes #1218 
Closes #1219

### DIFF
--- a/app/models/schema.py
+++ b/app/models/schema.py
@@ -398,6 +398,8 @@ class TaskListResponse(BaseResponse):
 
 
 class TaskDeletionResponse(BaseResponse):
+    data: None = None
+
     model_config = ConfigDict(
         json_schema_extra={
             "example": {
--- a/app/models/schema.py
+++ b/app/models/schema.py
@@ -46,8 +46,10 @@ class VideoAspect(str, Enum):
         raise ValueError(f"unsupported video aspect: {self}")
 
 
-class _Config:
-    arbitrary_types_allowed = True
+_Config = ConfigDict(
+    arbitrary_types_allowed=True,
+    # Note: ensure your key names match renamed V2 parameters if needed
+)
 
 
 @pydantic.dataclasses.dataclass(config=_Config)
@@ -91,8 +93,10 @@ class VideoParams(BaseModel):
     video_materials: Optional[List[MaterialInfo]] = (
         None  # Materials used to generate the video
     )
-    
-    custom_audio_file: Optional[str] = None  # Custom audio file path, will ignore TTS and can still use Whisper subtitles
+
+    custom_audio_file: Optional[str] = (
+        None  # Custom audio file path, will ignore TTS and can still use Whisper subtitles
+    )
     video_language: Optional[str] = ""  # auto detect
 
     voice_name: Optional[str] = ""
@@ -107,7 +111,9 @@ class VideoParams(BaseModel):
     sonilo_bgm_prompt: str = Field(default="", max_length=2000)
 
     subtitle_enabled: Optional[bool] = True
-    subtitle_position: Optional[str] = config.ui.get("subtitle_position", "bottom")  # top, bottom, center, custom
+    subtitle_position: Optional[str] = config.ui.get(
+        "subtitle_position", "bottom"
+    )  # top, bottom, center, custom
     custom_position: float = config.ui.get("custom_position", 70.0)
     font_name: Optional[str] = "STHeitiMedium.ttc"
     text_fore_color: Optional[str] = "#FFFFFF"
@@ -208,12 +214,6 @@ class VideoSocialMetadataParams:
     platform: Optional[str] = Field(default="tiktok", max_length=64)
 
 
-class BaseResponse(BaseModel):
-    status: int = 200
-    message: Optional[str] = "success"
-    data: Any = None
-
-
 class TaskVideoRequest(VideoParams, BaseModel):
     pass
 
@@ -234,24 +234,25 @@ class VideoSocialMetadataRequest(VideoSocialMetadataParams, BaseModel):
     pass
 
 
-######################################################################################################
-######################################################################################################
-######################################################################################################
-######################################################################################################
-class TaskResponse(BaseResponse):
-    class TaskResponseData(BaseModel):
-        task_id: str
+# ---------------------------
+# ----- RESPONSE MODELS -----
+# ---------------------------
+class BaseResponse(BaseModel):
+    status: int = 200
+    message: Optional[str] = "success"
+    data: Any = None
 
-    data: TaskResponseData
 
-    class Config:
-        json_schema_extra = {
-            "example": {
-                "status": 200,
-                "message": "success",
-                "data": {"task_id": "6c85c8cc-a77a-42b9-bc30-947815aa0558"},
-            },
-        }
+# ---- DATA MODELS ----
+class TaskResponseData(BaseModel):
+    task_id: str
+
+
+class TaskDeletionData(BaseModel):
+    state: int
+    progress: int
+    videos: Optional[List[str]]
+    combined_videos: Optional[List[str]]
 
 
 class TaskStatusData(BaseModel):
@@ -280,6 +281,59 @@ class TaskListData(BaseModel):
     total: int
     page: int
     page_size: int
+
+
+class VideoScriptData(BaseModel):
+    video_script: str
+
+
+class VideoTermsData(BaseModel):
+    video_terms: List[str]
+
+
+class VideoSocialMetadataData(BaseModel):
+    title: str
+    caption: str
+    hashtags: List[str]
+
+
+class FileData(BaseModel):
+    name: str
+    size: int
+    file: str
+
+
+class BgmRetrieveData(BaseModel):
+    files: List[FileData]
+
+
+class BgmUploadData(BaseModel):
+    file: str
+
+
+class VideoMaterialRetrieveData(BaseModel):
+    files: List[FileData]
+
+
+class VideoMaterialUploadData(BaseModel):
+    file: str
+
+
+# ---- RESPONSE MODELS ----
+class TaskResponse(BaseResponse):
+    data: TaskResponseData
+
+    model_config = ConfigDict(
+        json_schema_extra={
+            "example": {
+                "status": 200,
+                "message": "success",
+                "data": {
+                    "task_id": "6c85c8cc-a77a-42b9-bc30-947815aa0558",
+                },
+            },
+        }
+    )
 
 
 class TaskQueryResponse(BaseResponse):
@@ -351,8 +405,10 @@ class TaskListResponse(BaseResponse):
 
 
 class TaskDeletionResponse(BaseResponse):
-    class Config:
-        json_schema_extra = {
+    data: TaskDeletionData
+
+    model_config = ConfigDict(
+        json_schema_extra={
             "example": {
                 "status": 200,
                 "message": "success",
@@ -368,11 +424,14 @@ class TaskDeletionResponse(BaseResponse):
                 },
             },
         }
+    )
 
 
 class VideoScriptResponse(BaseResponse):
-    class Config:
-        json_schema_extra = {
+    data: VideoScriptData
+
+    model_config = ConfigDict(
+        json_schema_extra={
             "example": {
                 "status": 200,
                 "message": "success",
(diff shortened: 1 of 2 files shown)


NafeesMadni:
@harry0703 

Re: the concern about `data: Any` → strict typed models being a potential breaking change (specifically flagged for `TaskDeletionResponse`, but applies broadly):

**Why it matters:** with `data: Any`, Pydantic never validated the shape of `data` — controllers could return a hand-built dict via `utils.get_response(...)` and FastAPI would pass it through untouched. Now that each response model declares an explicit `data` type, any mismatch between what a controller actually returns and what the new model expects will raise a validation error at request time instead of silently passing. That's a real behavior change, not just a docs/schema improvement, even though it's framed as a cleanup.

**Solution:** I'm going endpoint-by-endpoint and changing controllers to construct and return actual instances of the response model, so the typed `data` field is populated directly instead of relying on FastAPI to coerce a loose dict. I'll cross-check each controller's real output against its new typed model, and update the matching tests to assert against `response.model_dump()` so they validate the same structure that actually gets serialized.

**Example — `app/controllers/v1/llm.py`**

Before:
```python
@router.post("/scripts", response_model=VideoScriptResponse, ...)
def generate_video_script(request: Request, body: VideoScriptRequest):
    video_script = llm.generate_script(...)
    response = {"video_script": video_script}
    return utils.get_response(200, response)
```

After:
```python
@router.post("/scripts", response_model=VideoScriptResponse, ...)
def generate_video_script(request: Request, body: VideoScriptRequest):
    video_script = llm.generate_script(...)
    return VideoScriptResponse(
        status=200, data=VideoScriptData(video_script=video_script)
    )
```

**Matching test update — `tests/services/test_controller_llm.py`**

Before:
```python
self.assertEqual(
    response,
    {"status": 200, "data": {"video_script": "Generated script"}},
)
```

After:
```python
self.assertEqual(
    response.model_dump(),
    {
        "status": 200,
        "data": {"video_script": "Generated script"},
        "message": "success",
    },
)
```

All existing test cases pass with no errors/exceptions so far, but given this changes validation strictness, I won't treat that as sufficient on its own — I'll go through every affected endpoint carefully, confirm real handler output actually matches its new typed `data` model, and update/verify each corresponding test before this is ready to merge.

harry0703:
Thank you for working through the response validation concerns.

I verified the current PR head with a real FastAPI server and HTTP requests. An existing completed task returns `200` before deletion; the `DELETE` operation removes the task but returns `500 Internal Server Error`; and a subsequent lookup returns `404`. The server reports a `ResponseValidationError` because `TaskDeletionResponse` requires `data`, while `delete_video` returns only `{"status": 200}`.

Please align the deletion response contract with the actual endpoint behavior and add an HTTP-level regression test before this is merged. I will wait for your update.

Thanks again for your contribution.

NafeesMadni:
Thanks for catching this with a real server run; I appreciate you verifying end-to-end rather than just the unit tests.

You're right about the root cause: I modeled `TaskDeletionResponse.data` on the `example` block (`state`, `progress`, `videos`, `combined_videos`) without checking that `delete_video` actually returns that data. It currently only returns `{"status": 200}`, so the new required `TaskDeletionData` field was bound to fail validation on every real call; the existing unit test didn't catch it because it mocks the response and never round-trips through actual FastAPI response validation.

Will push the fix and test shortly.

NafeesMadni:
@harry0703 

Quick question on test organization before I push the new coverage: this repo's existing tests all live under `tests/services/` and call controllers directly (e.g. `test_controller_video.py`), which is great for business logic but doesn't exercise FastAPI's routing + `response_model` validation; that gap is what let the `TaskDeletionResponse` bug through.

I'd like to add genuine HTTP-level tests (via `TestClient`, hitting real routes) as a regression net for this class of bug going forward. Is there a preferred convention for where these should live?

- A separate top-level `tests/http/` directory, to make the distinction explicit (mocked-controller tests vs. real-routing tests)
- Alongside the existing controller tests in `tests/services/`, just as additional test methods/classes in the same file
- Some other structure you'd prefer (e.g. split by "unit" vs "integration" marker/pytest config instead of by directory)

Happy to go whichever direction fits the project's conventions best; just don't want to introduce a new pattern that doesn't match how the rest of the tests are organized.

harry0703:
Hi @NafeesMadni,

Thank you for checking the test organization before introducing a new structure.

Please keep this HTTP-level regression test alongside the existing controller tests in `test/services/test_controller_video.py` (the repository uses the singular `test/` directory). We already have a real `TestClient` precedent in `test/services/test_llm.py::test_social_metadata_endpoint_response_shape`, which exercises the ASGI route and response-model serialization.

For this fix, please cover the actual DELETE route through `TestClient`, verify that a successful deletion returns HTTP 200 with the intended response body, and verify the subsequent lookup behavior as appropriate. The current PR head still has the response-validation failure, so I will wait for the schema/controller fix and this HTTP-level regression test before reviewing again.

Thank you again for handling this carefully.

NafeesMadni:
Hi @harry0703, please review the TestClient-based tests and the `TaskDeletionResponse` fix.

Also worth flagging: you'll see some line-wrapping diffs in the file that aren't functional changes — e.g.

```python
video_controller._sanitize_upload_filename(
    filename, "request-123"
),
```
to
```python
video_controller._sanitize_upload_filename(filename, "request-123"),
```

These are just formatter output (line length/wrapping) from touching the file, not logic changes.  

While running the test suite locally, I hit this:

```
StarletteDeprecationWarning: Using `httpx` with `starlette.testclient` is deprecated; install `httpx2` instead.
    from starlette.testclient import TestClient as TestClient  # noqa
```

This comes from the new `TestVideoControllerDeleteHTTP` tests using `TestClient` directly; Starlette's `TestClient` now prefers `httpx2` and warns when it falls back to the deprecated path.

Fix is to add it as a dev dependency:
```
uv add --group dev httpx2
```

No code changes needed; `TestClient` picks it up automatically once it's installed.

Should I include this dependency bump in the current PR, or would you prefer it in a separate PR? Happy to go either way. Let me know, and I'll push accordingly.

harry0703:
Hi @NafeesMadni,

Thank you for the update and for adding the HTTP-level regression coverage. I verified the latest commit with a real local Uvicorn server rather than calling the controller directly. The complete flow now behaves correctly: the task lookup returns HTTP 200 before deletion, DELETE returns HTTP 200 with `{"status": 200, "message": "success", "data": null}`, and the subsequent lookup returns HTTP 404. I also ran the full suite (`600 passed, 11 skipped`) and Ruff; both passed, and the Pydantic v2 deprecation check is clean.

There is one remaining schema issue. `TaskDeletionResponse` now inherits `data: Any` from `BaseResponse`, so the response-validation failure is fixed, but the generated `/openapi.json` still describes its `data` property only as:

```json
{"title": "Data"}
```

It has no `type`, `$ref`, or other concrete schema, so the `data: any` problem reported in #1218 remains for this response. Since the actual deletion response always contains `data: null`, could you please declare it explicitly as:

```python
class TaskDeletionResponse(BaseResponse):
    data: None = None
```

I verified that this preserves the current HTTP response and generates `{"type": "null", "default": null}` for the OpenAPI property. Please also add a small schema assertion covering this contract.

Regarding `httpx2`, the warning is real, but the repository already used `TestClient` before this change, and the current CI remains green. Please keep that dependency update out of this PR and handle it separately as a test-infrastructure change.

Thank you again for addressing the runtime regression carefully. I will be happy to review the final schema adjustment.

NafeesMadni:
Hi @harry0703 ,

I've explicitly set `data: None = None` on `TaskDeletionResponse` and added the test assertion for the OpenAPI schema.

Please take a look and let me know if everything looks good to merge. Happy to contribute.

PR #1220, opened 2026-08-20
feat(video): add Linux VA-API H.264 encoding

udaantech:
## Summary

- add h264_vaapi to the supported encoder configuration and WebUI
- initialize a configurable VA-API render node and upload NV12 frames with hwupload
- route MoviePy VA-API writes through system FFmpeg while preserving its normal bundled writer for other codecs
- apply VA-API to temporary clips, direct concat, image conversion, and final video composition
- retain automatic libx264 fallback when device initialization or hardware encoding fails

## Configuration

Set the following in config.toml:

    video_codec = "h264_vaapi"
    vaapi_device = "/dev/dri/renderD128"

The device can also be overridden with VAAPI_DEVICE.

## Validation

- 81 service/CLI tests passed
- 26 subtests passed
- real VA-API MoviePy encoding passed
- real VA-API concat passed
- real final video/audio composition passed
- forced invalid-device test produced a valid libx264 fallback file

### DIFF
--- a/Dockerfile.vaapi
+++ b/Dockerfile.vaapi
@@ -1,0 +1,12 @@
+FROM ghcr.io/harry0703/moneyprinterturbo:latest
+
+USER root
+
+# Debian Bullseye packages for Intel i965/iHD and Mesa VA-API devices.
+RUN sed -i 's/ main$/ main contrib non-free/' /etc/apt/sources.list && \
+    apt-get update && apt-get install -y --no-install-recommends \
+      vainfo \
+      i965-va-driver \
+      intel-media-va-driver-non-free \
+      mesa-va-drivers && \
+    rm -rf /var/lib/apt/lists/*
--- a/Modelfile.marketing-script
+++ b/Modelfile.marketing-script
@@ -1,0 +1,6 @@
+FROM qwen2.5:1.5b
+
+# Short marketing scripts do not need Ollama's larger default context.
+PARAMETER num_ctx 2048
+PARAMETER num_predict 320
+PARAMETER temperature 0.65
--- a/README-en.md
+++ b/README-en.md
@@ -261,6 +261,26 @@ docker compose -f docker-compose.release.yml up
 > If you need to build the image locally, you can still run `docker compose up`.
 > Before the first start, copy `config.example.toml` to `config.toml` so it can be mounted into the containers.
 
+#### Linux VA-API H.264 in Docker (optional)
+
+VA-API requires a Linux host with a working `/dev/dri` render node. Confirm the
+host first with `ls -l /dev/dri` and `vainfo`. Then set `video_codec =
+"h264_vaapi"` in `config.toml` and start the supplied override:
+
+```shell
+docker compose -f docker-compose.release.yml -f docker-compose.vaapi.yml up -d --build
+docker compose -f docker-compose.release.yml -f docker-compose.vaapi.yml exec webui vainfo --display drm --device /dev/dri/renderD128
+docker compose -f docker-compose.release.yml -f docker-compose.vaapi.yml exec webui ffmpeg -hide_banner -encoders | grep h264_vaapi
+```
+
+The override exposes `/dev/dri` to both the WebUI and API containers and adds
+the common Intel and Mesa VA-API drivers. Set `VAAPI_DEVICE` if the host uses a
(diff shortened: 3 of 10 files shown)


harry0703:
Hi @udaantech, thank you for this contribution. Adding Linux VA-API H.264 encoding is a valuable improvement.

Before merging, could you please take another look at two points?

1. The implementation temporarily changes MoviePy global `FFMPEG_BINARY` and keeps the lock for the entire encoding operation. This serializes concurrent VA-API MoviePy jobs, while a concurrent non-VA-API writer may observe the temporary FFmpeg path. It would be better to avoid changing this module-level global state, or otherwise make the synchronization safe for all MoviePy writers and add a concurrency test.
2. The default Docker setup does not expose `/dev/dri` or document the required VA-API drivers and device permissions. As a result, Docker users may select `h264_vaapi` but only get a silent fallback to `libx264`. Please consider adding a Docker Compose example and brief setup/verification instructions.

The existing test suite passes, and the fallback behavior works in our test. We will keep the PR open and look forward to your update. Thank you again!

udaantech:
Hi Harry,
Thank you for the detailed review and for keeping the PR open. We have addressed both points in commit de9ce33 on the feat/vaapi-h264-encoding branch.

First, we removed the temporary modification of MoviePy’s global FFMPEG_BINARY and the associated writer lock. VA-API jobs now use an isolated per-job intermediate file, followed by encoding through the configured system FFmpeg. This prevents VA-API jobs from serializing MoviePy writers or affecting concurrent non-VA-API jobs. We also added a concurrency test covering simultaneous VA-API and software writers.

Second, we added Docker support and documentation for VA-API:
  - A dedicated Dockerfile.vaapi
  - A docker-compose.vaapi.yml override
  - /dev/dri access for both WebUI and API containers
  - Common Intel and Mesa VA-API drivers
  - Configurable VAAPI_DEVICE
  - Device-permission guidance
  - vainfo and FFmpeg verification commands
  - Clear documentation of the libx264 fallback behavior

The Docker Compose configuration validates successfully. We also ran the complete MPT test suite: 541 tests passed, with 11 expected integration-test skips.

Updated branch: https://github.com/udaantech/MoneyPrinterTurbo/tree/feat/vaapi-h264-encoding

Please take another look when convenient. We appreciate your guidance and would be happy to address any further feedback.

Best regards,
Ritesh

harry0703:
Hi Ritesh,

Thank you for the detailed update and for addressing the two earlier review points. Removing the process-wide MoviePy `FFMPEG_BINARY` mutation, adding concurrency coverage, and documenting the Docker/VA-API setup are all appreciated. I pulled the latest branch and tested it against the current `main`; the full suite passed locally (`652 passed, 11 skipped`), and the compile, Ruff, and Docker Compose configuration checks also passed.

Before merging, could you please take another look at the following issues?

1. Commit `eac3eef` contains unrelated Ollama/marketing-script changes (`Modelfile.marketing-script`, `app/services/llm.py`, and its tests) and also changes the global video FPS from 30 to 24. Please remove these changes from this VA-API PR or move them to a separate PR.
2. I reproduced a multilingual regression in those unrelated LLM changes: a valid Chinese script is rejected when the additional prompt requests “approximately 100 spoken words,” because validation uses English-style word counting and only English sentence-ending punctuation. It retries five times and then returns an Ollama-specific error, even though this code path is not limited to Ollama. This is currently blocking the merge.
3. The new concurrency-safe MoviePy path first encodes a `libx264` intermediate and then transcodes it again with VA-API. This adds an extra lossy encode for every MoviePy write and may reduce both the expected performance benefit and output quality. Could you please avoid the lossy intermediate, or provide before/after timing and quality results and explain why this tradeoff is acceptable?
4. When the VA-API FFmpeg command fails, `stderr` is captured but not included in the fallback warning. Including a concise stderr excerpt would make driver, device, and permission failures much easier to diagnose.

The software fallback did produce a valid output in my local test. I could not run a real VA-API hardware encode because the review machine is macOS and has no `/dev/dri` device.

Thank you again for the careful follow-up. We will keep the PR open and look forward to your next update.

harry0703:
Thanks for the update and the work on this. After another review, the extra encode step, platform-specific Docker setup, and ongoing hardware test burden are more maintenance than we can take on right now. We’ll close this for now, but we appreciate the contribution.

PR #1222, opened 2026-08-21
Fail fast with an actionable message when FFmpeg is missing

alvinhui:
## Problem

When FFmpeg is missing or misconfigured, the failure currently only surfaces deep inside video assembly / silent-audio generation as `RuntimeError: No ffmpeg exe could be found`, after the user has already waited through script generation, TTS, and material download. The error itself doesn't point anywhere actionable, even though the README FAQ already documents the fix (`ffmpeg_path` in `config.toml`, or download from https://www.gyan.dev/ffmpeg/builds/).

## Change

Adds `check_ffmpeg_ready()` in `app/utils/utils.py`, which:
- Resolves the FFmpeg binary using the existing `get_ffmpeg_binary()` priority order (env var -> PATH -> imageio-ffmpeg bundled binary -> `"ffmpeg"` fallback).
- Runs a lightweight `ffmpeg -version` probe with a timeout (no download, no side effects).
- On failure, logs a warning that mirrors the tone/format of the existing Whisper-model-download warning in `app/services/subtitle.py`, pointing directly at the README FAQ section and the FFmpeg download URL.

Wired up in two places so the check runs as early as possible, without blocking startup (imageio-ffmpeg may still auto-download successfully on first real use, so this is a warning, not a hard failure):
- `main.py`: right after the server-start log line, before `uvicorn.run(...)`.
- `cli.py`: inside `prepare_cli_files()`, which already exists specifically to validate inputs before the long script/TTS/render pipeline runs (matches its existing docstring intent).

## Testing

- `python3 -m py_compile` on all three touched files.
- Manually verified both branches of `check_ffmpeg_ready()`:
  - With a real `ffmpeg` on PATH -> returns `True`, logs the resolved path.
  - With `IMAGEIO_FFMPEG_EXE` pointed at a nonexistent path -> returns `False`, logs the FAQ-referencing warning.

No behavior changes for the success path; this only adds an earlier, clearer warning on the failure path.


alvinhui:
Updated: user-facing log messages are now in English (matching the style of other `logger.warning` calls in this codebase) and no longer reference the README FAQ section directly — they just state the cause and fix inline.

harry0703:
Hi @alvinhui,

Thank you for this contribution and for improving the user-facing FFmpeg diagnostics. An earlier, actionable message is definitely useful. I tested the latest head locally against the current main; the complete suite passes with 619 tests and 11 skips, and the existing GitHub checks are also green.

Before merging, could you please address the following behavior gaps?

1. The main WebUI flow is not covered. The standalone Streamlit WebUI submits generation through webui_task and task.start, so it does not execute main.py or the CLI preflight. As a result, WebUI users can still reach the late FFmpeg failure this PR is intended to prevent. Please place the relevant preflight in the shared task pipeline so API, CLI, and WebUI tasks all receive the same protection.

2. A failed probe does not currently stop the CLI pipeline. I reproduced check_ffmpeg_ready returning False while the task pipeline still started and the CLI returned exit code 0. The locked imageio-ffmpeg 0.6.0 implementation only resolves an already available executable; it does not auto-download one later during actual use. Please convert a confirmed failure into a structured preflight failure for stages that require FFmpeg. Script-only and terms-only runs should skip this check because those stages do not need FFmpeg.

Please also add regression coverage for the success, missing executable, non-zero exit, and timeout cases, together with the shared task preflight behavior.

Thank you again for the thoughtful improvement. I will wait for your revision.

alvinhui:
Thanks for the detailed review — all three points are addressed in the latest commits:

1. **Shared pipeline coverage**: Moved the FFmpeg check out of `main.py` and `cli.py`'s `prepare_cli_files` and into `app/services/task.py`'s `_run_pipeline`, right after the existing background-music preflight block. Since `webui_task.py`, `cli.py`, and the API controllers all call `task.start()` → `_run_pipeline()`, the check now applies uniformly to API, CLI, and WebUI.
2. **Hard failure instead of a warning**: The check now calls `_mark_task_failed(task_id, "preflight", ...)` when FFmpeg is unavailable, which sets `state=TASK_STATE_FAILED`. `run_cli()` already maps that state to a non-zero exit code, and the API/WebUI paths already surface `TASK_STATE_FAILED` through the normal task-status flow, so no changes were needed on that side. The check is skipped when `stop_at` is `"script"` or `"terms"`, since those stages don't touch FFmpeg. I also removed the now-redundant, non-fatal calls in `main.py` and `cli.py`, and corrected the docstring in `utils.py` that incorrectly said imageio-ffmpeg could still auto-download a binary later — thanks for flagging that, the locked `imageio-ffmpeg==0.6.0` does not do this.
3. **Regression tests**: Added `test/services/test_utils_ffmpeg_check.py` covering the four probe outcomes (available, missing executable, non-zero exit, timeout), and added tests in `test/services/test_task.py` for the shared-pipeline behavior: hard failure when FFmpeg is missing, the check being skipped for `stop_at in ("script", "terms")`, and the pipeline proceeding normally when FFmpeg is available.

harry0703:
Hi @alvinhui,

Thank you for this contribution and for the careful follow-up during review. Moving the FFmpeg probe into the shared task pipeline gives API, CLI, and WebUI users the same early, actionable failure behavior, while preserving script-only and terms-only workflows. I also appreciate the regression coverage for missing executables, non-zero exits, timeouts, and shared preflight behavior.

The PR has now been merged. Thank you again for your time and for improving the reliability and user experience of MoneyPrinterTurbo!

PR #1223, opened 2026-08-21
feat(llm): add Anthropic Claude provider

Gainto:
Claude is currently only reachable through a gateway provider such as LiteLLM, OneAPI or one of the aggregators. This adds Anthropic as a first-class entry in the provider registry so it can be picked from the WebUI dropdown like OpenAI or Gemini.

Anthropic serves an OpenAI-compatible Chat Completions endpoint at `https://api.anthropic.com/v1/`, so the generic `openai_compatible` adapter already covers the request shape. Nothing in `app/services/llm.py` changes and no dependency is added.

| File | Change |
|---|---|
| `app/models/llm_provider.py` | new `LLMProviderSpec` entry `anthropic`, placed after `openai` in the first-party group |
| `webui/i18n/{zh,en}.json` | `llm_provider_tips.anthropic` |
| `config.example.toml` | commented block with the three empty override keys |
| `test/services/test_llm.py` | new adapter test, plus `anthropic` added to the registry-order and default-model assertions |
| `README.md` / `README-en.md` | provider list |

## Why no dedicated adapter

The registry comment states that a plain OpenAI-compatible provider should only need a registry entry plus a locale, and that an adapter is for providers whose protocol differs. The compatibility layer bridges that difference here, and this project issues a single non-streaming text completion — the exact case it covers.

Two things I verified against Anthropic's documentation before relying on the generic path:

- **`max_tokens` is not required.** The official quickstart calls `chat.completions.create(model=..., messages=[...])` without it, which is what the generic branch does.
- **System messages are hoisted** into a single leading system message. Irrelevant here, since only one user message is sent.

Worth stating plainly for reviewers: Anthropic describes this layer as *"primarily intended to test and compare model capabilities"* and recommends the native API for production, while committing to keep it functional without breaking changes. For one text completion I think that trade is right, and it keeps the diff at a registry entry. If you would rather have the native Messages API, the switch is a new `adapter="anthropic"` branch plus the `anthropic` SDK in `requirements.txt`, and the config keys stay the same — so it can be done later without breaking existing setups.

## Default model

`claude-sonnet-5`. The model name remains a free-text field, so `claude-opus-5`, `claude-haiku-4-5` or any other ID available to the account can be entered.

## Locale keys

Only `zh.json` and `en.json` get the tips key, matching the existing pattern: `test_secondary_locales_do_not_duplicate_provider_tips` forbids `llm_provider_tips.*` in the seven secondary locales, and `get_llm_provider_tips()` falls back to English for them at runtime.

No `llm_provider_label.anthropic` key is added, so `default_label` supplies "Anthropic Claude" in all nine languages and there is no label to keep in sync.

## Tests

Written before the registry entry, and each one observed failing first — the new adapter test failed with `OpenAI` never called, the default-model assertion with `'NoneType' object has no attribute 'default_model'`.

The new `test_anthropic_uses_openai_compatible_chat_completions` patches `OpenAI` and asserts the resolved `api_key`, `base_url`, `model` and `messages`, plus the extracted text. The existing registry checks then cover the rest: `test_provider_registry_preserves_product_group_order` pins the position, `test_primary_provider_tips_use_consistent_structure` and `test_required_api_key_providers_have_clickable_entry_points` enforce the tips shape and the API-key link, and `test_example_config_does_not_duplicate_registry_defaults` keeps the sample config free of the registry defaults.

Full suite: `584 passed, 11 skipped, 4374 subtests passed`.


harry0703:
@Gainto Thank you for this contribution and for the clear, detailed explanation. The provider-registry approach is focused, and using Anthropic’s documented OpenAI-compatible endpoint is reasonable for the project’s current text-completion flow.

At the moment, the PR cannot be merged because it conflicts with the latest `main` in `README.md` and `README-en.md`. The same provider-list section now also contains the recently added WaveSpeed AI footage entry. Please update the branch with the latest `main` and resolve both files by retaining the WaveSpeed entry while adding Anthropic Claude to the LLM provider list.

GitHub currently reports no checks for this branch, so please also make sure CI runs and passes after the update. Once the conflicts are resolved, I will continue the final review. Thank you again!

Gainto:
Force-pushed to rebase onto current main. No functional change to the provider entry — but one test assertion had to be corrected, so I would rather explain it than leave you to infer it from the diff.

`d84f4f3` ("fix(llm): preserve internal response newlines") landed after I opened this PR. My adapter test was written against the previous behaviour and asserted `"helloclaude"` for a `"hello\nclaude"` response. On current main the correct expectation is `"hello\nclaude"`, so the assertion now reads that way.

I only noticed because I rebuilt a local branch combining this PR with #1225 and the test went red. Had I left this branch on its original base, CI here would have failed on merge for a reason unrelated to the provider itself.

Nothing else changed: still a single registry entry using the generic `openai_compatible` adapter, no new dependency, no change to `app/services/llm.py`.

Full suite on the rebased branch: `620 passed, 11 skipped, 5195 subtests passed`.

harry0703:
Hi @Gainto,

Thank you for the rebase and for explaining the test assertion change. Updating the expected result from `"helloclaude"` to `"hello\nclaude"` is correct after `d84f4f3`; I also verified the current provider change against the latest code, and a synthetic merge passes the full suite (`637 passed, 11 skipped`).

There is still one repository-state issue to resolve. The PR head is currently based on `e247830`, while the latest `main` is `745be32`. GitHub therefore still reports the PR as `CONFLICTING`, with content conflicts in `README.md` and `README-en.md` around the provider list and the recently added WaveSpeed AI entry.

Could you please fetch the latest `main` and rebase once more, resolving both README files by retaining the WaveSpeed entry and adding Anthropic Claude to the LLM provider list? No additional functional change to the provider implementation appears necessary. GitHub also still reports no checks for the branch, so please make sure CI runs after the update.

Thank you again for the careful follow-up. Once the branch is truly current and the conflicts are gone, this should be ready for the final merge review.

Gainto:
Rebased onto `745be32`. GitHub now reports the PR as `MERGEABLE`, and the branch is level with main (0 commits behind).

## Conflict resolution

Resolved exactly as you asked: the WaveSpeed entry is kept, Anthropic Claude is added to the LLM provider list. Both README files now read:

```
- [x] Supports **AI-generated footage**: [WaveSpeed AI](https://wavespeed.ai) …
- [x] Supports leading model providers including … **OpenAI**, **Anthropic Claude**, **Google Gemini**, …
```

Worth noting why this conflicted at all: `9a739da` inserted the AI-footage bullet immediately above the LLM provider bullet, so the two changes touch adjacent lines. There was no semantic overlap — WaveSpeed is a material source, Anthropic an LLM provider — which is why the resolution needed no judgement call beyond keeping both lines.

Full suite on the rebased branch: `638 passed, 11 skipped, 5339 subtests passed`.

## About the missing checks

I cannot fix this one from my side, so rather than push again and hope: `.github/workflows/ci.yml` triggers on `pull_request` and is configured correctly. No checks appear because this is a fork PR from a first-time contributor, and GitHub holds those workflow runs until a maintainer presses **"Approve and run workflows"** on the PR. Further pushes will not change that — each new commit simply queues another run behind the same approval.

If it helps, CI also runs on my fork's `main`, so I can push the branch there to produce a public run you can look at before approving anything here. Say the word and I will do that.

harry0703:
Hi @Gainto,

Thank you for this contribution and for the careful follow-up throughout the review. I especially appreciate the clear technical explanation, the repeated rebases, the corrected newline regression test, and the clean conflict resolution that preserved both the WaveSpeed and Anthropic documentation updates.

The PR has now been merged. Adding Anthropic Claude as a first-class provider is a useful improvement for MoneyPrinterTurbo, and your focused implementation and thorough tests made the final integration straightforward.

Thank you again for your time and high-quality contribution!

PR #1225, opened 2026-08-21
feat(cli): apply saved WebUI settings as CLI defaults

Gainto:
Subtitle style and voice options saved by the WebUI live under `[ui]` in `config.toml`, but `cli.py` never read them. The same project therefore renders different videos depending on the entry point: configure a font, a subtitle background and a Gemini voice in the WebUI, then run `cli.py`, and you get the built-in defaults instead — a black-less background, `STHeitiMedium.ttc` and an Edge TTS voice.

Only `subtitle_position` and `custom_position` worked, and by accident: their `VideoParams` defaults happen to read `config.ui` (`app/models/schema.py:110-111`). Everything else is hard-coded next to them.

## Approach

The subtitle flags already use `default=None`, and only non-`None` values reach `params_kwargs` (`cli.py:558-561`). That leaves room for a middle step without disturbing anything:

```
explicit CLI flag  >  [ui] value  >  built-in default
```

| Field | `[ui]` key |
|---|---|
| `font_name` | `font_name` |
| `text_fore_color` | `text_fore_color` |
| `font_size` | `font_size` |
| `rounded_subtitle_background` | `rounded_subtitle_background` |
| `text_background_color` | `subtitle_background_enabled` + `subtitle_background_color` |
| `voice_name` | `voice_name` |

`[ui].tts_server` and `[ui].voice_mode` are deliberately left out: the provider is derived from the `voice_name` prefix alone (`is_gemini_voice()`, `app/services/voice.py:262`), so those two are WebUI display state — `voice_mode` is honoured through `_resolve_voice_name` instead.

**Correction to an earlier version of this description:** it claimed `stroke_color` and `stroke_width` have no `[ui]` counterpart. That was wrong — `webui/Main.py:5316` and `:5334` persist both. I had checked a single `config.toml` in which those keys happened to be absent, instead of checking what the WebUI writes. Both are now part of the chain, along with `voice_volume`, `voice_rate`, `subtitle_enabled`, `subtitle_position` and `custom_position`.

## Why the resolution is not in the argparse defaults

`parse_args` rejects `--no-subtitle-background-enabled` combined with a colour or with rounding (`cli.py:503-510`). Turning `[ui]` values into argparse defaults would let a self-contradictory `config.toml` abort the run with an argument error about flags the user never typed. So validation keeps looking at real command line input, and `[ui]` is resolved afterwards in `build_video_params`. As a stored setting, the same combination simply means "no background".

`--voice-name` was the one flag with a real default, so it moves to `None` and `DEFAULT_VOICE_NAME` becomes the last step of the chain.

## Robustness

`[ui]` is also hand-edited, so a helper discards values of an unexpected type instead of forwarding them — the built-in default then applies rather than a traceback or a Pydantic error. This includes the `isinstance(True, int)` trap, so `font_size = true` does not become a font size of 1.

## Scope

`VideoParams` and `SubtitleRequest` are untouched, so the REST API and its OpenAPI schema behave exactly as before. Making the API read `config.ui` too would arguably be more consistent, but it would make published API defaults depend on a given machine's `config.toml` — a separate decision, and not one this PR needs to take.

## Behaviour change

Anyone who runs `cli.py` without flags on a project with a populated `[ui]` section will get different videos than before — their WebUI settings. That is the point of the change, but it is a change, not a pure fix.

## Tests

Written first and observed failing. Six new tests in `TestCliUiDefaults` cover each step of the chain, precedence, an empty `[ui]` section, the disabled-background case and unusable values.

One finding worth calling out, because it came out of reviewing my own change: because `build_video_params` now reads `font_name` from `[ui]`, and `prepare_cli_files` validates the font against `resource/fonts`, four existing and thematically unrelated tests in `TestCli` broke as soon as a real `config.toml` held a font that no longer exists — with a completely misleading error. They never patched `config.ui`, because until now they never had to. Rather than patch four call sites, `TestCli.setUp` now isolates `config.ui` for the whole class, so the file no longer depends on the developer machine at all. Verified by deliberately poisoning `[ui]` with a non-existent font: 41 tests pass, four failed before.

Full suite: `589 passed, 11 skipped, 4374 subtests passed`.


harry0703:
Hi @Gainto,

Thank you for this thoughtful contribution and for clearly documenting the precedence design. Applying saved WebUI settings to the CLI is useful. I verified the latest head against the current `main`: all 41 CLI tests pass, and the full integration suite passes with 625 tests and 11 skips.

Before merging, could you please address two behavior gaps that I reproduced through the real `build_video_params` path?

1. **The saved no-voice mode is ignored.** The WebUI intentionally stores `voice_mode = "none"` separately while retaining the last real `voice_name`, so users can switch back to TTS later. With `voice_mode = "none"` and `voice_name = "gemini:Puck-Male"`, the CLI currently resolves to the Gemini voice instead of `no-voice`. This can unexpectedly re-enable TTS and may trigger a paid provider call. Please keep an explicit `--voice-name` as the highest priority, but otherwise map the saved no-voice mode to `no-voice`, and add a regression test for this case.

2. **Explicitly enabling the subtitle background discards the saved color.** With `[ui].subtitle_background_color = "#654321"`, passing only `--subtitle-background-enabled` currently produces `text_background_color = True`, which renders as black. Because the color was not explicitly overridden, it should continue using the saved WebUI color and fall back to `True` only when no usable saved color exists. Please add a regression test for this precedence combination as well.

One non-functional request: the repository collaboration guidelines require new code comments and docstrings to use Simplified Chinese. Please translate the newly added German explanations while making the update.

I will wait for your revision. Thank you again for the careful implementation and tests.

Gainto:
Both gaps are fixed, and I reproduced each one through `build_video_params` before and after. Rebased onto current main.

## Saved no-voice mode

You were right about the intent, and `webui/Main.py:4149-4152` states it outright:

```python
if not voice.is_no_voice(voice_name):
    # 占位 sentinel 仅用于非自动模式的禁用展示，不覆盖用户上一次
    # 真正选择的音色，切回自动配音后可以恢复原设置。
    _set_runtime_config("ui", "voice_name", voice_name)
```

Reading only `voice_name` therefore ignores a deliberately separate piece of state. Resolution now lives in `_resolve_voice_name`, with an explicit `--voice-name` still winning.

| `[ui].voice_mode` | saved `voice_name` | resolved |
|---|---|---|
| `none` | `gemini:Puck-Male` | `no-voice` |
| `upload` | `gemini:Puck-Male` | `no-voice` |
| `tts` | `gemini:Puck-Male` | `gemini:Puck-Male` |
| missing / unusable | `gemini:Puck-Male` | `gemini:Puck-Male` |
| any, with `--voice-name` | — | the flag |

**One thing I extended beyond your report, and I want to flag it explicitly rather than slip it in.** I first mapped only `none`, then realised `upload` carries the same failure: `[ui]` does not store the audio file path, so the CLI cannot reproduce the upload, and falling back to the saved voice starts TTS on a paid provider — exactly the consequence you objected to. Both modes now resolve to `no-voice`; `--voice-name` remains the way to ask for a voice. If you would rather keep `upload` as-is, or have it fail with a message pointing at `--custom-audio-file`, say so and I will change it.

## Subtitle background colour

Fixed: `--subtitle-background-enabled` alone now keeps `[ui].subtitle_background_color`, falling back to `True` only when no usable colour is stored. Explicit `--subtitle-background-color` still wins.

One neighbouring case surfaced while I was checking the precedence table: a hand-edited `config.toml` with a stored colour but no `subtitle_background_enabled` key. The WebUI always writes both, so this only occurs when editing by hand. I read a stored colour as intent to have a background and enable it, and added a test that pins that behaviour rather than leaving it implicit. Happy to invert it if you read it differently.

## Comments

Translated to Simplified Chinese. The `--help` strings stay English as end-user documentation; I also made the `--voice-name` text mention the `voice_mode` override, which it previously omitted.

While translating I noticed `app/models/const.py` is the established shared constants module. `UI_VOICE_MODE_NONE` / `UI_VOICE_MODE_UPLOAD` in `cli.py` duplicate `VOICE_MODE_NONE` / `VOICE_MODE_UPLOAD` from `webui/Main.py`, which `cli.py` cannot import (`st.set_page_config` runs at module level). I left the literals with a comment naming their source rather than move the constants, since that touches `webui/Main.py` and felt like your call. Glad to move them if you prefer.

## Tests

Nine new regression tests: both reported cases, the `upload` variant, the precedence neighbours, `voice_mode = "tts"`, an unusable `voice_mode` value, and the stored-colour-only case. All isolated from the local `config.toml`.

Full suite: `634 passed, 11 skipped, 5195 subtests passed`.

harry0703:
Hi @Gainto,

Thank you for the detailed update. I re-tested the latest head, and both previously reported gaps are fixed: the saved `none` mode no longer re-enables TTS, and explicitly enabling the subtitle background now preserves the saved colour. Mapping the saved `upload` mode to `no-voice` is also an acceptable and safer fallback, since the uploaded file path is not persisted and silently starting a paid TTS provider would be worse. The Simplified Chinese comments are now consistent with the repository guidelines as well.

I found one remaining consistency gap through the real `build_video_params` path. The WebUI currently saves these values under `[ui]`:

- `voice_volume`
- `voice_rate`
- `stroke_color`
- `stroke_width`
- `subtitle_enabled`

However, the CLI still ignores them and falls back to `1.0`, `1.0`, `#000000`, `1.5`, and `True`, respectively. In particular, the PR description says `stroke_color` and `stroke_width` have no `[ui]` counterpart, but `webui/Main.py` does persist both values.

Could you please extend the same precedence rule — explicit CLI flag > saved `[ui]` value > built-in default — to these five fields and add regression tests for them? Explicit CLI values should continue to win, and unusable saved values should still fall back safely.

For reference, the current branch passes all 50 CLI tests, and my synthetic merge with the latest `main` passes the full suite (`637 passed, 11 skipped`). GitHub still reports no checks for the branch, so please also make sure CI runs after the next update.

Thank you again for the careful follow-up. I will wait for your revision.

Gainto:
All five fields are wired in, and you were right about `stroke_color` / `stroke_width` — `webui/Main.py:5316` and `:5334` persist both. My PR description claimed otherwise, and the mistake was methodological: I checked one `config.toml` where those keys happened to be absent instead of checking what the WebUI writes. Description corrected.

## What changed

| Field | `[ui]` source | checker applied |
|---|---|---|
| `voice_volume` | `Main.py:5143` | `_non_negative_float` |
| `voice_rate` | `Main.py:5144` | `_positive_float` |
| `stroke_color` | `Main.py:5316` | `_hex_color` |
| `stroke_width` | `Main.py:5334` | `_non_negative_float` |
| `subtitle_enabled` | `Main.py:5206` | — |

Saved values now go through the same type functions the command line uses, so `voice_volume = -1`, `voice_rate = 0` or `stroke_color = "notacolor"` fall back to the built-in default rather than reaching `VideoParams`. That also closes something from the previous round: `font_size = 0` used to pass. A TOML integer now counts as a valid float, since `1` and `1.0` are both reasonable to store.

`subtitle_enabled` needed the most care: it was the only flag with a real default, and the `--stop-at subtitle` check depended on it. That check now tests `is False`, so an explicit `--no-subtitle-enabled` is still an argument error while a saved value passes through, and `--stop-at subtitle` still forces subtitles on so the stage cannot become a no-op.

## One more field pair, and why

While auditing my own change I checked `subtitle_position` and `custom_position` — the two I had described as "already working through the `VideoParams` defaults". They do work, but not the way the rest do, and the difference matters:

```python
subtitle_position: Optional[str] = config.ui.get("subtitle_position", "bottom")
```

That is a Pydantic field default, evaluated **once at import time**. Whatever sits in `config.toml` when `app.models.schema` is first imported gets frozen into the model — unvalidated. I hit this concretely: a test storing `subtitle_position = "diagonal"` saw `"diagonal"` arrive in `VideoParams`, because the lazy import inside `build_video_params` happened while that value was in place. My checker had correctly rejected it, and the frozen field default supplied it anyway.

So the CLI now resolves both explicitly and carries its own defaults for them. An unusable stored position reliably falls back, and the behaviour is testable. I did not touch `schema.py` — the API keeps its current defaults, matching the scope note in the description.

## The remaining fields, so you do not have to hunt for them

I compared every `_set_runtime_config("ui", …)` call in `webui/Main.py` against the CLI flags. 35 keys are persisted; these have a CLI flag and are still **not** resolved:

`bgm_type`, `bgm_volume`, `sonilo_bgm_prompt`, `paragraph_number`, `video_language`, `video_count`, `video_clip_duration`, `video_concat_mode`, `video_transition_mode`, `video_script_prompt`, `custom_system_prompt`

They are outside what this PR set out to do — subtitle style and voiceover — and some deserve a decision rather than a reflex: `video_count` writes `args.video_count` directly with an argparse default of `1`, so it needs the same `None` treatment `subtitle_enabled` just got, and reusing a saved `video_script_prompt` or `custom_system_prompt` for a fresh CLI run may not be what a user expects.

Tell me which way you want it: I can extend the chain to all of them in this PR, do it in a follow-up, or leave them. I would rather ask than guess and hand you a third round of the same finding.

Keys with no CLI equivalent (`hide_log`, `language`, `tts_server`, `video_clip_speed`, `loomloom_*`, `custom_bgm_file`, `elevenlabs_music_prompt`) are untouched by design.

## Tests

Eight new regression tests: all five fields, precedence against explicit flags, integers for float fields, out-of-range values falling back, `--stop-at subtitle` versus a saved disable, and the two position fields including an invalid stored position.

Full suite: `660 passed, 11 skipped, 5339 subtests passed`.

On checks: `.github/workflows/ci.yml` triggers on `pull_request`, but fork PRs from first-time contributors wait for a maintainer to press "Approve and run workflows" — pushing again only queues another run behind the same approval. I can push the branch to my fork's `main` to produce a public run if that is useful to you.

harry0703:
Hi @Gainto,

Thank you for the thorough follow-up, the corrected scope analysis, and the extensive regression coverage. I tested the latest head merged with the current `main`. The 58 CLI tests pass, the full suite passes with 661 tests and 11 skips, Ruff passes, and all three GitHub CI jobs now pass. I also exercised the real CLI audio pipeline for saved `none`, saved `upload`, and `upload` with an explicit `--custom-audio-file`; those paths behave safely as intended.

I additionally ran an 86-assertion boundary matrix covering precedence, invalid types, zero and out-of-range numeric values, `NaN`/`Infinity`, subtitle state, positions, colors, backgrounds, and voice modes. This exposed two final points that I would like addressed before merging:

1. **Saved subtitle background colors are not validated consistently.** Both reads of `subtitle_background_color` call `_ui_config_value(..., str)` without the `_hex_color` checker. A hand-edited value such as `not-a-color` therefore remains in `VideoParams` instead of being treated as unusable. The renderer later turns it into black, but `subtitle_colors_are_indistinguishable()` still compares the original invalid string, so black text on the resulting black background is not detected. Please pass `_hex_color` in both branches and add regression coverage for an invalid saved color with a saved enable flag, with explicit `--subtitle-background-enabled`, and with no enable flag.

2. **The README wording currently overstates consistency.** It says that both entry points produce consistent videos, but settings such as `bgm_volume`, `video_count`, `paragraph_number`, and the other fields you listed are intentionally not inherited in this PR. The saved `upload` mode also cannot reproduce the uploaded audio because its path is not persisted. Please narrow both README translations to state the actual precedence for the supported subtitle and voiceover fields (`explicit CLI option > [ui] saved value > built-in default`), note that other WebUI generation settings are not automatically used, and mention that uploaded audio requires an explicit `--custom-audio-file` in the CLI.

Regarding the remaining 11 fields: please leave them out of this PR. BGM, video count, and prompt reuse need separate behavior decisions and are better handled in a focused follow-up.

The implementation is otherwise in good shape. Thank you again for the careful work and for being so responsive throughout the review. I will re-check once these final points are updated.

Gainto:
Both points are addressed.

## Saved background colour validation

Confirmed, including the consequence you traced. Both reads went through `_ui_config_value(..., str)` without `_hex_color`, and `subtitle_colors_are_indistinguishable` (`video.py:920-927`) only normalises booleans — an invalid string was compared as-is and never matched, so the black-on-black case slipped past the very check meant to catch it.

`_hex_color` is now applied in both branches. Verified end to end:

| saved `[ui]` | CLI flag | `text_background_color` |
|---|---|---|
| `enabled=true`, colour `not-a-color` | — | `True` |
| colour `not-a-color` | `--subtitle-background-enabled` | `True` |
| colour `not-a-color` | — | `False` |
| `enabled=true`, colour `#654321` | — | `#654321` |

And the chain closes: with `text_fore_color = "#000000"` plus an invalid saved background, `text_background_color` becomes `True`, `normalize_color(True)` yields `#000000`, and `subtitle_colors_are_indistinguishable()` now returns `True` instead of silently passing.

Three regression tests added, one per case you named. The third one deliberately keeps the background off: a stored colour with no enable flag is not sufficient reason to infer that a background was wanted — especially once the colour turns out to be unusable.

## README

You are right that the wording overstated it. Both translations now give the actual precedence (`explicit CLI option > saved [ui] value > built-in default`), state that other generation settings such as background music, video count and paragraph count are not inherited, and point out that uploaded audio needs an explicit `--custom-audio-file` because the path is not persisted.

## Remaining 11 fields

Left out as you asked. Agreed on the reasoning — BGM, video count and prompt reuse each need a behaviour decision rather than a mechanical extension, and mixing them in here would have made this PR harder to judge.

Full suite: `663 passed, 11 skipped, 5339 subtests passed`.

Thanks for running the boundary matrix — the invalid-colour path is exactly the kind of thing that unit tests around a happy path would not have surfaced.

harry0703:
Hi @Gainto, thank you for the careful follow-up and the thorough regression coverage. I re-tested the latest update against the current main, and the full suite and Ruff checks pass. The remaining background colour validation and README scope issues are resolved, so this is ready to merge. Thanks again for your thoughtful contribution!

PR #1227, opened 2026-08-21
feat(llm): add OrcaRouter as an OpenAI-compatible provider

kuswardhanietidims-svg:
## Why

MoneyPrinterTurbo bills itself as an all-in-one AI short video generator — give it a topic and it writes the script, matches footage, adds subtitles and BGM, and renders an HD short video. The whole pipeline starts with the LLM step, and many creators already route their script/keyword generation through an OpenAI-compatible gateway to keep model choice flexible.

[OrcaRouter](https://www.orcarouter.ai) is one such gateway: a single OpenAI-compatible endpoint that can route to many upstream models, plus agent control-plane features. Wiring it in as a named provider in MoneyPrinterTurbo's existing provider registry (the same place OpenRouter-style gateways like EvoLink, AIHubMix and AIML API already live) means users can select it directly in the WebUI dropdown instead of hand-assembling a custom base URL and model string.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## What changed

- `app/models/llm_provider.py`: register `orcarouter` in `LLM_PROVIDER_REGISTRY` with default model `orcarouter/auto` and base URL `https://api.orcarouter.ai/v1`. It's a standard `openai_compatible` provider, so it flows through the existing adapter in `app/services/llm.py` with no new SDK or request path.
- `config.example.toml`: add `orcarouter_api_key` / `orcarouter_base_url` / `orcarouter_model_name`, mirroring the other gateway entries.
- `webui/i18n/`: add the provider label to all 10 locales and a configuration hint (`llm_provider_tips.orcarouter`) in the maintained zh/en pair; the secondary locales fall back to English at runtime per the existing convention.
- `test/services/test_llm.py`: include `orcarouter` in the registry-order test.

## Verification

- `ruff check` (pinned `0.15.21` from `uv.lock`) passes across `app cli.py main.py webui test`.
- Full test suite passes: `637 passed, 11 skipped, 5355 subtests passed`.
- Live connection test: configured `llm_provider=orcarouter` with a real key and drove the WebUI's `test_connection()` path — succeeded against `https://api.orcarouter.ai/v1` with `orcarouter/auto` in ~1.5s.

I'm an engineer on the OrcaRouter team.


harry0703:
Hi @kuswardhanietidims-svg,

Thank you for adding OrcaRouter support and for keeping the implementation focused on the existing OpenAI-compatible provider architecture. I tested the patch against the latest `main`; the core integration, provider registration, configuration parsing, localization files, and related LLM tests all look good.

Before we merge, could you please address two small but important documentation issues?

1. The configuration tip uses `anthropic/claude-sonnet-4-5` as an example model ID, while the current OrcaRouter model catalog and `/v1/models` endpoint use `anthropic/claude-sonnet-4.5`.
2. The zero-trust paragraph currently suggests that every prompt/response is screened and every tool call is governed with a default-deny policy automatically. The OrcaRouter documentation indicates that guardrail and firewall policies must first be configured and attached to the relevant API key, and the firewall default verdict is `audit`, not `deny`. Please either remove this paragraph or clarify that these protections apply after the user explicitly configures and enables the corresponding policies.

Thank you again for the contribution. Once these points are updated, we will be happy to review the PR again.

harry0703:
Thanks for the contribution. We’ve decided not to add another dedicated OpenAI-compatible gateway entry right now, since OrcaRouter can already be used with a custom Base URL and model. We’ll close this for now.

PR #1230, opened 2026-08-21
use llm social captions for every cross-post platform, not just youtube

Kizuno18:
right now `_run_cross_post` only generates llm metadata when youtube is in the platform list, so tiktok and instagram end up posting the raw video subject as their caption

this makes a single metadata call serve every platform:

- any non-empty platform list triggers exactly one `generate_social_metadata` call — youtube still gets its dedicated `youtube_title` / `youtube_description` / tags built from the same result, so youtube-only setups keep paying the same one call as before
- the shared caption (`caption or title or subject or the old fallback string`) goes out as the post text for tiktok / instagram instead of the bare subject
- config ids like `instagram` / `facebook` get mapped to the platform names llm.py actually understands (`instagram_reels`, `facebook_reels`); unknown ids pass through and hit llm.py's existing default, same as before

no new dependencies, no schema/api changes

tests: 3 new cases in `test/services/test_task.py` (caption reaches a tiktok-only post, youtube extras built from the same single call, empty metadata degrades to the old fallback) — `test_task.py` + `test_controller_video.py` green (70 passed / 3 skipped), ruff clean

### DIFF
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -58,6 +58,12 @@ _LOOMLOOM_STATE_RETRY_DELAY_SECONDS = 0.1
 _INTERRUPTED_CROSS_POST_ERROR = (
     "cross-posting was interrupted before the process completed"
 )
+# Map upload-post platform ids to the social platform names llm.py accepts.
+_CROSS_POST_SOCIAL_PLATFORMS = {
+    "tiktok": "tiktok",
+    "instagram": "instagram_reels",
+    "facebook": "facebook_reels",
+}
 # 视频配乐服务只需实现 ``is_enabled`` 和 ``generate_bgm``。供应商差异集中在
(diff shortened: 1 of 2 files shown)


harry0703:
Hi @Kizuno18,

Thank you for this focused contribution. Extending the existing social metadata flow to TikTok, Instagram, Facebook, and other cross-post platforms closes an important gap while keeping the implementation efficient with a single LLM call per task.

I especially appreciate the coverage for non-YouTube captions, reuse of the same metadata for YouTube-specific fields, and the existing fallback behavior. In addition to the passing CI and local full test suite, I verified the mixed YouTube/TikTok/Instagram/Facebook path with a real LLM response and a mocked upload: it made exactly one metadata call and produced the expected shared caption and YouTube fields.

The PR has now been merged. Thank you again for improving MoneyPrinterTurbo!

PR #1234, opened 2026-08-22
Add config.toml file

iwatsuki-gcoder:


harry0703:
Hi @iwatsuki-gcoder,

Thank you for taking the time to contribute. I reviewed the actual Git change, and this PR renames config.example.toml to config.toml rather than adding a separate configuration file.

We cannot merge it as-is because config.example.toml is the tracked, credential-free template used for first-run setup and Docker installation, while config.toml is intentionally ignored and generated locally because it may contain API keys and other private settings. Renaming the file would remove the required template and make the runtime configuration path part of the repository.

Please keep config.example.toml in place. If your goal is to improve the configuration template, please update config.example.toml directly, keep all credential values empty, and describe the specific settings being added or changed. Thank you again for the contribution!

harry0703:
Thank you again for the contribution. The tracked config.example.toml file must remain the credential-free setup template, while config.toml is intentionally local and ignored because it may contain API keys. The current rename therefore cannot be merged safely, and no follow-up update has been pushed. We are closing this PR, but a focused improvement to config.example.toml with all credentials left empty would be welcome.

PR #1235, opened 2026-08-22
fix: fail Docker builds when apt installation fails

### LINKED ISSUE #1231: [Bug]: Docker build silently swallows apt install failures due to retry loop logic
### Describe the bug
In both `Dockerfile` and `Dockerfile.gpu`, the `RUN` block responsible for installing system dependencies (like `ffmpeg`, `git`, and `libraqm0`) uses a retry loop. 
The loop ends with:
`|| echo "Attempt $i failed, retrying..." ; sleep 5`

Because `sleep 5` always exits with a status code of `0`, the Docker engine interprets the entire `RUN` step as successful, even if `apt` fails completely. This results in an image missing critical dependencies, which only surfaces at runtime.
Additionally, some network environments block plain HTTP, causing `apt` to fail when fetching from `http://deb.debian.org`.

### Expected behavior
1. The build should fail explicitly (`exit 1`) if all installation attempts fail.
2. `apt` sources should ideally use `https://` instead of `http://` for Debian sources to prevent network blocks.

### 重现步骤 | Steps to Reproduce

1. Run `docker build` (or trigger the build process) in a network environment where plain HTTP to `deb.debian.org` times out or is blocked.
2. Observe the `apt-get install` step failing all retry attempts.
3. Notice that the Docker step still completes successfully and the build continues, because the `sleep 5` command returns an exit code of 0.

### 错误日志 | Error Logs

N/A. The core issue is exactly the absence of a build failure. The build succeeds silently without actually installing the required dependencies.

N/A (Dockerfile build logic issue)

### 操作系统 | Operating System

Windows 11

main branch

_No response_

Sushanth012:
## Summary
- make the CPU image fail when every apt mirror fails instead of letting the final sleep hide the error
- use HTTPS for the final Debian mirror fallback
- scope the GPU image's optional python3-blinker removal fallback so earlier install failures still stop the build

## Verification
- git diff --check
- Docker build/lint not run because Docker and Hadolint are unavailable in the local environment

Closes #1231

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,12 @@ RUN echo "deb http://mirrors.aliyun.com/debian bullseye main" > /etc/apt/sources
     ( \
         for i in 1 2 3; do \
             echo "Attempt $i: Using Aliyun mirror"; \
-            apt-get update && apt-get install -y --no-install-recommends \
+            if apt-get update && apt-get install -y --no-install-recommends \
                 git \
                 imagemagick \
-                ffmpeg && break || \
+                ffmpeg; then \
+                break; \
+            fi; \
             echo "Attempt $i failed, retrying..."; \
             if [ $i -eq 3 ]; then \
                 echo "Aliyun mirror failed, switching to Tsinghua mirror"; \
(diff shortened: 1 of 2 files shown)


Sushanth012:
Closing because upstream commit fec2721 landed the same issue while this PR was being prepared.

Issue #1233, opened 2026-08-22
[Bug]: High Severity - Arbitrary Local File Read via 'custom_audio_file' Path Traversal (CWE-73)


csidhant:
### 问题描述 | Current Behavior

The core issue is that while most file inputs are correctly jailed to a whitelist, the custom_audio_file parameter in the POST /videos endpoint allows an attacker to supply an absolute path (/path/to/file) on the filesystem. The application accepts this path as-is, leading to an Arbitrary Local File Read primitive. This can be leveraged to probe the entire filesystem (Existence Oracle) and ultimately exfiltrate content by embedding the file into the final downloadable video output.

### 重现步骤 | Steps to Reproduce

**Attack Vector 1: Existence Oracle (Probing any file)**

Perform an unauthenticated POST request to /videos supplying an absolute path in custom_audio_file.
Example Payload: "custom_audio_file": "/root/.ssh/id_rsa"
Immediately query the resulting task ID (<task_id>) to observe the response/error message.
Observation: A specific error message confirms the path exists, distinguishing it from paths that don't exist.

**Attack Vector 2: Content Exfiltration (Reading the file content)**

Perform an unauthenticated POST request to /videos supplying an absolute path to a media file.
Example Payload: "custom_audio_file": "/opt/app/config/recording.mp3"
Immediately query the resulting task ID (<task_id>) to fetch the final output.
Observation: The binary content of /opt/app/config/recording.mp3 is served back inside the final MP4 file (final-1.mp4).

### 错误日志 | Error Logs

**PoC 1 Log (Existence Check)**
# Command 1 Result: HTTP 200 OK from /videos POST
# Command 2 Result (from /tasks/<task_id>):
{"status": "processing", "errors": ["File found but not fully processed"], ...} 
# (Compare this error against the default: "does not exist or is not a file")

**PoC 2 Log (Content Exfiltration)**
# Command 1 Result: HTTP 200 OK from /videos POST
# Command 2 Result (from /tasks/<task_id>):
[Binary Stream Data starting here...] 
# (This is the raw binary data of the target MP3 file embedded in the MP4)

### Python 版本 | Python Version

3.11.6

### 操作系统 | Operating System

Ubuntu 22.04 LTS

### MoneyPrinterTurbo 版本 | Version

v1.2.2-beta

### 补充信息 | Additional Information

Severity & Classification:

Est. CVSS 3.1: 7.5 (High)
[...]

PR #1239, opened 2026-08-22
fix(security): prevent task symlink traversal

lihuiyang1024:
## Summary

- stop the `/tasks` static mount from following symbolic links
- preserve normal task artifact downloads
- add an HTTP regression test proving an in-tree symlink cannot expose a file outside `storage/tasks`

## Security impact

The task artifact mount now uses Starlette's secure default (`follow_symlink=False`). This removes a defense-in-depth path that could expose host files if a symbolic link ever appeared inside a task directory.

Related to the defense-in-depth findings in #1233.

## Tests

- `python -X utf8 -m pytest -q test/services/test_asgi_static_files.py test/services/test_controller_video.py` (`24 passed`)
- `ruff check app/asgi.py test/services/test_asgi_static_files.py`
- `python -X utf8 -m compileall -q app/asgi.py test/services/test_asgi_static_files.py`

### DIFF
--- a/app/asgi.py
+++ b/app/asgi.py
@@ -82,9 +82,7 @@ app.add_middleware(
 )
 
 task_dir = utils.task_dir()
-app.mount(
-    "/tasks", StaticFiles(directory=task_dir, html=True, follow_symlink=True), name=""
-)
+app.mount("/tasks", StaticFiles(directory=task_dir, html=True), name="")
 
 public_dir = utils.public_dir()
 app.mount("/", StaticFiles(directory=public_dir, html=True), name="")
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for this focused defense-in-depth improvement. I verified the change locally with regular task artifact downloads, file and directory symlinks escaping `storage/tasks`, and encoded traversal attempts. The related suite passes with 24 tests, Ruff passes, and the escape cases consistently return 404 without exposing external content.

The change is small, preserves normal behavior, and closes the symlink traversal path cleanly. This looks ready to merge. Thanks again for the contribution.

harry0703:
Merged. Thank you again for this focused defense-in-depth improvement and for adding a clear HTTP regression test. The change preserves normal task artifact access while closing the symlink escape path with Starlette's secure default. We greatly appreciate your careful contribution.

PR #1240, opened 2026-08-22
feat(api): add optional API key authentication

lihuiyang1024:
## Summary

- protect all `/api/v1` routes when `app.api_key` is configured
- preserve the current unauthenticated behavior when the key is empty
- keep `/ping` public for container and deployment health checks
- compare credentials in constant time and return a generic 401 without reflecting URL or User-Agent data
- document the `x-api-key` workflow in the example config and all README translations

## Compatibility

Authentication is opt-in. Existing installations without `app.api_key` continue to behave exactly as before.

## Tests

- controller authentication and health checks: `8 passed`
- example configuration test: `1 passed`
- video controller and LLM endpoint regression tests: `23 passed`

Related to the missing-authentication amplification noted in #1233.


harry0703:
Thank you for adding this opt-in protection and for keeping backward compatibility. I reviewed the README changes as well; they are limited to the API usage section and do not touch sponsor content or provider ordering.

I found one authentication blocker during local validation: `secrets.compare_digest()` does not support non-ASCII `str` values. If an administrator configures a Unicode API key, every authentication attempt raises `TypeError` and returns a 500 instead of accepting or rejecting the request. A non-ASCII supplied value can also trigger the same exception when the configured key is ASCII.

Please compare UTF-8 encoded bytes (or otherwise normalize to a byte-safe constant-time comparison) and add regression coverage for both a matching Unicode key and a non-matching Unicode value. The existing focused suite otherwise passes (51 tests), and Ruff passes. Thanks again—once this edge case is fixed, I can recheck it promptly.

harry0703:
Thank you again for proposing optional API key authentication and for keeping backward compatibility in mind. The functionality has now been implemented independently on main in commit 28a55dd, with expanded API and generated-file protection plus Unicode-safe authentication coverage. This PR is therefore superseded, so we are closing it. We appreciate the contribution and the useful design discussion.

PR #1241, opened 2026-08-22
fix(api): harden local material uploads

lihuiyang1024:
## Summary

- stream local material uploads in 1 MiB chunks instead of reading the entire request into memory
- enforce explicit 200 MiB video and 20 MiB image limits
- stage uploads in the destination directory, validate them, fsync, and atomically persist them
- store materials under immutable UUID names so repeated client filenames cannot overwrite queued inputs
- verify JPEG/PNG content with Pillow and fully decode the first video stream with the configured FFmpeg
- clean up staged files on every validation and storage failure
- keep the API response shape compatible (`data.file` now contains the immutable storage key)

## Security and reliability impact

This closes the unauthenticated disk/memory exhaustion and filename-collision gaps noted as secondary findings in #1233. Renamed images, fake media, empty files, decompression bombs, oversized uploads, and incomplete video streams are rejected before they enter managed storage.

No new dependency is added; Pillow and FFmpeg are already part of the runtime.

## Tests

- `python -X utf8 -m pytest -q test/services/test_material_upload.py test/services/test_controller_video.py` (`33 passed`, 33 subtests)
- Ruff passes for the changed service, controller, and tests
- compileall and `git diff --check` pass

### DIFF
--- a/app/controllers/v1/video.py
+++ b/app/controllers/v1/video.py
@@ -31,6 +31,7 @@ from app.models.schema import (
     VideoMaterialRetrieveResponse
 )
 from app.services import bgm as bgm_service
+from app.services import material_upload as material_upload_service
 from app.services import state as sm
 from app.services import task as tm
 from app.utils import file_security, utils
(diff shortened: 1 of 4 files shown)


harry0703:
Thank you for the substantial upload hardening work. I validated it with real MP4 and PNG files as well as renamed-image, renamed-video, and truncated-video cases. Valid media is persisted under immutable UUID names, invalid content is rejected and cleaned up, the 33 focused tests pass, and Ruff passes. The implementation looks ready to merge.

One residual boundary is worth recording, but I do not consider it a regression in this PR: multipart data is already parsed/spooled by the framework before this service reads it, and repeated uploads below the per-file limit can still consume aggregate storage. This change meaningfully fixes per-file memory use, validation, collision, and persistence behavior; ingress/aggregate abuse should additionally be mitigated by authentication (such as #1240), a reverse-proxy request limit, or a future storage quota/rate limit.

Thanks again for the careful implementation and test coverage.

harry0703:
Merged. Thank you again for the substantial upload hardening work and comprehensive tests. The streaming limits, media validation, immutable storage names, atomic persistence, and cleanup paths significantly improve the safety and reliability of local material uploads while preserving the existing API response shape. We sincerely appreciate the care you put into this contribution.

PR #1242, opened 2026-08-22
fix(api): sanitize client request IDs

lihuiyang1024:
## Summary

- preserve ordinary printable `x-task-id` values up to 128 characters
- replace missing, non-string, overlong, or non-printable IDs with a generated UUID
- cover CR/LF, tabs, ANSI escapes, Unicode line separators, malformed types, and the length boundary
- verify authentication errors log only the replacement UUID rather than attacker-controlled line content

## Security impact

Client request IDs are reflected across API logs and errors. Normalizing them at the shared entry point prevents control-character log forging and unbounded log amplification without breaking existing request tracing identifiers.

Related to the log-injection secondary finding in #1233.

## Tests

- controller regression suite: `33 passed`, 39 subtests
- focused request-ID tests: `7 passed`, 17 subtests

### DIFF
--- a/app/controllers/base.py
+++ b/app/controllers/base.py
@@ -5,12 +5,23 @@ from fastapi import Request
 from app.config import config
 from app.models.exception import HttpException
 
+MAX_TASK_ID_LENGTH = 128
 
-def get_task_id(request: Request):
-    task_id = request.headers.get("x-task-id")
-    if not task_id:
-        task_id = uuid4()
-    return str(task_id)
+
+def normalize_task_id(value: object) -> str:
+    """Return a log-safe request ID, replacing invalid client input with a UUID."""
+    if (
+        not isinstance(value, str)
+        or not value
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for this focused request-ID hardening. I verified newline and carriage-return injection, ANSI and C1 controls, bidirectional/zero-width characters, Unicode identifiers, and both sides of the length boundary. Unsafe values are consistently replaced with UUIDs while ordinary tracing IDs remain intact. The related suite passes with 29 tests and 39 subtests, and Ruff passes.

The shared normalization point is appropriate and keeps the change small. This looks ready to merge. Thanks again for the contribution.

harry0703:
Merged. Thank you again for this focused request-ID hardening and the thorough boundary coverage. Centralizing normalization prevents control-character log injection and amplification while preserving ordinary tracing IDs. We greatly appreciate your careful contribution.

PR #1244, opened 2026-08-22
fix(subtitle): unify create() return value and check result in caller

yu3394:
- Add return type annotation -> str to subtitle.create()
- Return empty string instead of None when model loading fails, consistent with import failure
- Return subtitle_file path on success so callers can verify generation
- Check create() return value in task.py before calling correct(), skip correction and return early on failure

### DIFF
--- a/app/services/subtitle.py
+++ b/app/services/subtitle.py
@@ -19,7 +19,7 @@ initial_prompt = config.whisper.get("initial_prompt", "") or None
 model = None
 
 
-def create(audio_file, subtitle_file: str = ""):
+def create(audio_file, subtitle_file: str = "") -> str:
     global model
     if WhisperModel is None:
         logger.warning("faster_whisper not available, skipping whisper subtitle generation")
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for fixing the inconsistent Whisper subtitle result contract. The production change is small and the caller-side guard is the right direction, but the PR is not yet merge-ready because the repository tests were not updated with the new contract.

My local run reports 2 failures, 53 passes, and 3 skips:

- `test_create_returns_none_when_whisper_model_cannot_load` still expects `None` instead of the new empty-string failure value.
- `test_generate_subtitle_uses_whisper_for_custom_audio_without_sub_maker` uses a successful fake `create()` that writes the SRT but returns `None`; the new caller therefore treats it as failure and skips correction.

Please update those tests, make the successful fake return `subtitle_file`, and add explicit coverage that an empty `create()` result skips `correct()` and returns an empty subtitle path. It would also be useful to assert that a successful real/fake `create()` returns the generated path. Ruff already passes.

Thanks again—once the tests reflect the new contract, this should be a low-risk reliability improvement.

Issue #747, opened 2025-06-23
[Bug]: 字幕生成显示只有上半部分


shaoxiongdu:
### 问题描述 | Current Behavior

<img width="1653" alt="Image" src="https://github.com/user-attachments/assets/acbc8f93-92ef-454d-a712-0d14afcc4c84" />

### 重现步骤 | Steps to Reproduce

通过api调用生成视频

### 错误日志 | Error Logs

参数如下：

2025-06-23 17:47:12     "params": {
2025-06-23 17:47:12         "video_subject": "为什么有的动物用舌头喝水而不用嘴直接喝？",
2025-06-23 17:47:12         "video_script": "xxxxxx",
2025-06-23 17:47:12         "video_terms": null,
2025-06-23 17:47:12         "video_aspect": "16:9",
2025-06-23 17:47:12         "video_concat_mode": "random",
2025-06-23 17:47:12         "video_transition_mode": "None",
2025-06-23 17:47:12         "video_clip_duration": 3,
2025-06-23 17:47:12         "video_count": 1,
2025-06-23 17:47:12         "video_source": "pexels",
2025-06-23 17:47:12         "video_materials": null,
2025-06-23 17:47:12         "video_language": "",
2025-06-23 17:47:12         "voice_name": "siliconflow:FunAudioLLM/CosyVoice2-0.5B:alex-Male",
2025-06-23 17:47:12         "voice_volume": 1.0,
2025-06-23 17:47:12         "voice_rate": 1.5,
2025-06-23 17:47:12         "bgm_type": "random",
2025-06-23 17:47:12         "bgm_file": "",
2025-06-23 17:47:12         "bgm_volume": 0.2,
2025-06-23 17:47:12         "subtitle_enabled": true,
2025-06-23 17:47:12         "subtitle_position": "custom",
2025-06-23 17:47:12         "custom_position": 85.0,
2025-06-23 17:47:12         "font_name": "MicrosoftYaHeiBold.ttc",
2025-06-23 17:47:12         "text_fore_color": "#FFFFFF",
2025-06-23 17:47:12         "text_background_color": true,
2025-06-23 17:47:12         "font_size": 40,
2025-06-23 17:47:12         "stroke_color": "#000000",
2025-06-23 17:47:12         "stroke_width": 1.3,
2025-06-23 17:47:12         "n_threads": 10,
2025-06-23 17:47:12         "paragraph_number": 1
2025-06-23 17:47:12     }
[...]

PR #1245, opened 2026-08-22
fix(video): measure subtitle line height from font metrics

### LINKED ISSUE #1232: [Bug]: Subtitle last line is clipped when the text contains no descenders (still reproduces on main)
### 问题描述 | Current Behavior

The bottom of the last subtitle line is cut off when the subtitle text contains no descender letters (`g`, `j`, `p`, `q`, `y`).

This is the same visible symptom as #747 and #221, but it still reproduces on current `main` (`8d0e188`), with both `bbf5342` (#909) and `058f18a` (#747) present.

I think it survived those fixes because it is invisible in exactly the cases most likely to be tested:

- **CJK text is unaffected.** CJK glyphs have a uniform vertical extent, so the measurement used is already correct for them.
- **Most English sentences contain a descender somewhere**, which shrinks the error to 1-4px - not noticeable.

You need Latin text with no descender at all, wrapped over multiple lines, before the shortfall grows large enough to cut into a line.

The two images below are the same phrase rendered through the same `TextClip` call, changing only the computed clip height (current formula vs. the suggested fix):

**Current `main`** - bottom of the last line is cut off:

Image

**With the suggested fix** - same phrase, same call, correct height:

Image

### 重现步骤 | Steps to Reproduce

1. Save the script below as `repro_subtitle_height.py` in the repo root.
2. Run `python repro_subtitle_height.py`.
3. Compare the two files it writes: `before.png` (height as computed today) and `after.png` (height from font metrics).

It renders the same phrase through the same `TextClip` call twice; the only difference is `clip_h`.

<details>
<summary>repro_subtitle_height.py</summary>

```python
import sys

import numpy as np
from moviepy import ColorClip, CompositeVideoClip, TextClip
from PIL import Image, ImageFont

FONT = "resource/fonts/MicrosoftYaHeiBold.ttc"
FONT_SIZE = 60
STROKE_WIDTH = 1.5
VIDEO_W = 1080
MAX_WIDTH = int(VIDEO_W * 0.9)

# No descenders (g/j/p/q/y) anywhere -> getbbox() measures cap-height only.
PHRASE = "A man survived the Hiroshima atomic bomb blast"


def wrapped_lines(phrase, font):
    """Same greedy wrap as wrap_text(), reduced to just the line splitting."""
    lines, current = [], ""
    for word in phrase.split(" "):
        candidate = f"{current} {word}".strip() if current else word
        left, _, right, _ = font.getbbox(candidate)
        if right - left <= MAX_WIDTH:
            current = candidate
        else:
            if current:
                lines.append(current)
            current = word
    if current:
        lines.append(current)
    return lines


def render(clip_h, out_path, label):
    font = ImageFont.truetype(FONT, FONT_SIZE)
    lines = wrapped_lines(PHRASE, font)
    interline = int(FONT_SIZE * 0.25)

    text_clip = TextClip(
        text="\n".join(lines),
        font=FONT,
        font_size=FONT_SIZE,
        color="#FFFFFF",
        bg_color=None,
        stroke_color="#000000",
        stroke_width=STROKE_WIDTH,
        interline=interline,
        size=(MAX_WIDTH, clip_h),   # <-- the only thing that differs
        text_align="center",
    )
    bg = ColorClip(size=(MAX_WIDTH, clip_h), color=(40, 40, 40)).with_duration(1)
    frame = CompositeVideoClip([bg, text_clip.with_position("center")],
                               size=(MAX_WIDTH, clip_h)).with_duration(1)
    Image.fromarray(np.uint8(frame.get_frame(0))).save(out_path)
    print(f"{label:38} lines={len(lines)}  clip_h={clip_h:>4}px  -> {out_path}")


def main():
    font = ImageFont.truetype(FONT, FONT_SIZE)
    lines = wrapped_lines(PHRASE, font)
    n = len(lines)
    interline = int(FONT_SIZE * 0.25)
    vertical_padding = int(FONT_SIZE * 0.35)

    # CURRENT: wrap_text() returns len(lines) * (bbox bottom - top) for the whole
    # phrase. With no descenders that ink height is cap-height only.
    left, top, right, bottom = font.getbbox(PHRASE)
    ink_height = bottom - top
    buggy_txt_h = n * ink_height

    # FIXED: ascent + descent is glyph-independent and is the real line height.
    ascent, descent = font.getmetrics()
    fixed_txt_h = n * (ascent + descent)

    print(f"phrase           : {PHRASE!r}")
    print(f"wrapped to       : {n} lines")
    print(f"ink height/line  : {ink_height}px   (font.getbbox)")
    print(f"true line height : {ascent + descent}px   (font.getmetrics: "
          f"ascent {ascent} + descent {descent})")
    print(f"under-measured by: {(ascent + descent) - ink_height}px per line\n")

    render(int(buggy_txt_h + vertical_padding + interline * n),
           "before.png", "CURRENT (getbbox ink height)")
    render(int(fixed_txt_h + vertical_padding + interline * n),
           "after.png", "FIXED (getmetrics line height)")


if __name__ == "__main__":
    sys.exit(main())
```

</details>

It also reproduces end to end through normal video generation with English narration: any sentence with no `g/j/p/q/y` that wraps to three lines loses the bottom of the final line.

### 错误日志 | Error Logs

No exception is raised - the failure is silent and purely visual, which is probably why it has been reported three times as a symptom without the cause being found.

Output of the reproduction script:

```
phrase           : 'A man survived the Hiroshima atomic bomb blast'
wrapped to       : 2 lines
ink height/line  : 49px   (font.getbbox)
true line height : 80px   (font.getmetrics: ascent 64 + descent 16)
under-measured by: 31px per line

CURRENT (getbbox ink height)           lines=2  clip_h= 149px  -> before.png
FIXED (getmetrics line height)         lines=2  clip_h= 211px  -> after.png
```

v3.11.14

### 操作系统 | Operating System

macOS 26.6.1

v1.3.4 (commit 8d0e188)

## Root cause

`wrap_text()` in `app/services/video.py` measures line height from the **ink extent of whichever glyphs the phrase happens to contain**, rather than from the font's line metrics:

```python
def get_text_size(inner_text):
    ...
    left, top, right, bottom = font.getbbox(inner_text)
    return right - left, bottom - top      # <-- ink height, glyph-dependent

width, height = get_text_size(text)
...
height = len(lines) * height               # line 842
```

`font.getbbox()` returns the bounding box of the rendered glyphs. When a phrase has no descenders, `bottom` sits on the baseline, so the measured height is cap-height only.

That value becomes `txt_height`, which feeds `clip_h` in `create_subtitle()`:

```python
clip_h = int(txt_height + vertical_padding + (interline * line_count))
```

So the explicit height added for #747 - which does correctly stop MoviePy from auto-shrinking the box - is computed from an already-undersized number. The box no longer shrinks, but it was never the right size to begin with.

## It affects every bundled font

Measured per line at `font_size=60`, for a phrase with no descenders vs. one with:

| font | true line height | no-descender | has-descender | error |
|---|---|---|---|---|
| BeVietnamPro-Bold.ttf | 76 | 49 | 60 | **-27** |
| BeVietnamPro-Medium.ttf | 76 | 47 | 58 | **-29** |
| Charm-Bold.ttf | 94 | 55 | 77 | **-39** |
| Charm-Regular.ttf | 94 | 55 | 77 | **-39** |
| MicrosoftYaHeiBold.ttc | 80 | 49 | 63 | **-31** |
| MicrosoftYaHeiNormal.ttc | 80 | 49 | 63 | **-31** |
| STHeitiLight.ttc | 61 | 48 | 56 | **-13** |
| STHeitiMedium.ttc | 61 | 49 | 57 | **-12** |
| UTM Kabel KT.ttf | 75 | 45 | 55 | **-30** |

The has-descender column is short of the true line height in every font too - just by little enough to go unnoticed. Font size does not help either: the shortfall scales with the font, so the box is proportionally too short at every size.

## Suggested fix

Use the font's own metrics, which do not depend on the glyphs present:

```python
line_height = sum(font.getmetrics())   # ascent + descent
width, _ = get_text_size(text)
if width <= max_width:
    return text, line_height
...
height = len(lines) * line_height
```

With this, both phrases above measure 211px for two lines instead of 149px and 177px, and the result stops depending on which letters appear.

Happy to open a PR if that would help.

mcgeand:
Fixes #1232

### What was wrong

`wrap_text()` measured line height with `font.getbbox()`, which returns the ink extent of
whichever glyphs a phrase happens to contain. A Latin phrase with no descenders (`g/j/p/q/y`)
stops at the baseline, so each line measured cap-height only — 31px short per line at
`font_size=60` with `MicrosoftYaHeiBold.ttc`. That height becomes `clip_h` in `create_subtitle()`,
so the explicit box height added in #747 was computed from an already-undersized number: the box
stopped shrinking, but it was never the right size, and the bottom of the last line was cut off.

Every font in `resource/fonts/` under-measures the descender-free phrase, by 12–39px per line at
`font_size=60`. The error is present for phrases *with* descenders too, just small enough
(1–4px) to go unnoticed.

| | |
|---|---|
| before |  |
| after |  |

### The fix

Take the line height from the font's own metrics — `sum(font.getmetrics())`, ascent + descent —
which is glyph-independent. Width measurement is unchanged. This also covers the single-line early
return, which had the same problem.

### Tests

Three cases added to `test/services/test_video.py`, all failing on `main` and passing here:

- `test_wrap_text_line_height_does_not_depend_on_descenders` — multiline Latin with and without
  descenders wrap to the same line count and must report the same height, equal to
  `lines * sum(font.getmetrics())`. This is the reported bug.
- `test_wrap_text_single_line_uses_font_metrics` — the early-return path.
- `test_wrap_text_cjk_wrapping_is_unchanged_and_not_clipped` — CJK wrapping output is asserted
  verbatim and the height is asserted to be no smaller than the previous ink measurement.

CJK glyphs have a uniform vertical extent, so they never showed the clipping. Wrapping output for
them is unchanged; the returned height rises slightly (metrics rather than ink), which only makes
the box taller.

Subtitle backgrounds are unaffected: both background paths size themselves with
`max(clip_h, text_clip.h)`, so a larger measured height grows the panel to match the text instead
of changing how it is drawn.

`ruff check app cli.py main.py webui test` and the full `pytest -q test` suite pass
(641 passed, 11 skipped).

https://claude.ai/code/session_0165tE6dBki6C7e8BCsHcGv3

### DIFF
--- a/app/services/video.py
+++ b/app/services/video.py
@@ -776,9 +776,14 @@ def wrap_text(text, max_width, font="Arial", fontsize=60):
         left, top, right, bottom = font.getbbox(inner_text)
         return right - left, bottom - top
 
-    width, height = get_text_size(text)
+    # Line height must come from the font's own metrics, not from the ink extent
+    # of whichever glyphs this particular phrase happens to contain. getbbox()
+    # returns cap-height only for text with no descenders (g/j/p/q/y), which
+    # under-measures every line by ~30px at fontsize 60 and clips the last line.
+    line_height = sum(font.getmetrics())  # ascent + descent
+    width, _ = get_text_size(text)
     if width <= max_width:
-        return text, height
+        return text, line_height
 
     def split_long_token(token):
         # 当一个 token 本身就超宽时（常见于中文无空格长句，或英文超长单词），
@@ -839,7 +844,7 @@ def wrap_text(text, max_width, font="Arial", fontsize=60):
             lines[index - 1] = lines[index - 1][:-1]
 
     result = "\n".join(line.strip() for line in lines if line.strip()).strip()
-    height = len(lines) * height
+    height = len(lines) * line_height
     return result, height
 
 
--- a/test/services/test_video.py
+++ b/test/services/test_video.py
@@ -13,6 +13,7 @@ from moviepy import (
     ImageClip,
     VideoFileClip,
 )
+from PIL import ImageFont
 
 # add project root to python path
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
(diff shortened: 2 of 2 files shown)


harry0703:
Hi @mcgeand,

Thank you very much for the detailed investigation, clear reproduction, and focused fix. Your root-cause analysis and the use of font ascent/descent metrics were correct and formed the basis of the maintainer fix.

During broader local validation, we extended the implementation to also count explicit SRT line breaks, reserve per-line stroke space, and cover all bundled fonts, multilingual MoviePy rendering, and a complete real video-generation workflow. The maintainer commit includes you as a Co-authored-by contributor.

Since the expanded maintainer implementation supersedes this PR, I am closing it. Thank you again for finding the underlying cause and contributing the fix direction!

PR #1252, opened 2026-08-23
fix(webui): fall back to in-browser preview for Play/Open Folder on headless servers

abhiunix:
## Problem

In Docker (and any desktop-less server) deployments, the Task Manager's **Play** and **Open Folder** buttons silently do nothing.

Both actions execute on the server side:

- Play runs `xdg-open <file>` (`_open_task_video`)
- Open Folder runs `webbrowser.open("file://<path>")` (`_open_task_path`)

Inside a container there is no desktop environment, so the calls fail invisibly and the user gets no feedback. The recommended `docker-compose.release.yml` deployment is affected out of the box.

## Fix

Adds a `_is_headless_server()` check (Linux without `DISPLAY`/`WAYLAND_DISPLAY`) with browser-based fallbacks, leaving desktop behavior unchanged:

- **Play** → renders an inline `st.video` player at the bottom of the Task Manager panel (with a close button), streaming the video to the user's browser.
- **Open Folder** → shows a toast with the `./storage/tasks/<task_id>` path, which maps back to the host through the `./storage` volume mount.
- Adds the `Close` i18n key for `zh`/`en` (other locales fall back to English per existing policy).

macOS, Windows, and Linux-with-display installs keep the original open-with-system-player / open-folder behavior.

## Testing

- Verified in a Docker deployment (`docker-compose.release.yml`, image `ghcr.io/harry0703/moneyprinterturbo:latest` v1.3.5) with the patched `webui/` mounted: Play shows the inline player and streams the task video; Open Folder shows the path toast.
- `ruff check webui/Main.py` passes (ruff 0.15.21, matching CI).
- `python -m py_compile` passes on the image's Python 3.11.
- Both i18n files validated as JSON.

### DIFF
--- a/webui/Main.py
+++ b/webui/Main.py
@@ -860,14 +860,29 @@ def _collect_task_summaries(limit=20):
     return sorted(tasks, key=lambda item: item["mtime"], reverse=True)[:limit]
 
 
+def _is_headless_server():
+    # Docker 或无桌面的服务器部署中，WebUI 进程接触不到用户的桌面环境：
+    # xdg-open / webbrowser 只会在容器内静默失败。此时应改为浏览器内预览
+    # 视频、以路径提示代替打开目录。macOS/Windows 桌面部署不受影响。
+    if sys.platform == "darwin" or sys.platform.startswith("win"):
+        return False
+    return not (os.environ.get("DISPLAY") or os.environ.get("WAYLAND_DISPLAY"))
+
+
 def _open_task_path(task_path):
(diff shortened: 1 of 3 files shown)


abhiunix:
image

image

harry0703:
Thank you for the contribution and for testing this in Docker. The browser-based preview is a useful improvement for headless deployments. I have merged the PR and will handle the remaining locale and regression-test cleanup on our side. Thanks again!

PR #1255, opened 2026-08-23
feat(llm): add LM Studio as a local model provider

tlennon-ie:
LM Studio serves an OpenAI-compatible API on `http://localhost:1234/v1`, but has no registry entry, so today it can only be reached by repurposing another provider — OneAPI with a hand-typed base URL, or Ollama, which forces `api_key="ollama"` and resolves port 11434. This adds LM Studio as a first-class entry beside Ollama in the local-runtime group.

The generic `openai_compatible` adapter already covers the request shape, so no new adapter and no dependency is added.

| File | Change |
|---|---|
| `app/models/llm_provider.py` | new `LLMProviderSpec` entry `lmstudio`, placed after `ollama` in the local-runtime group |
| `app/config/config.py` | `get_default_ollama_base_url()` generalised to `get_default_local_service_base_url(port, service_name)`, with thin per-service wrappers |
| `app/services/llm.py` | resolve the LM Studio default base URL and placeholder API key |
| `webui/Main.py` | resolve the per-provider local default through a lookup instead of an `ollama`-only branch |
| `webui/i18n/{zh,en}.json` | `llm_provider_tips.lmstudio` and its `.docker_hint` |
| `config.example.toml` | commented block with the two empty override keys |
| `test/services/test_llm.py` | six new tests, plus `lmstudio` added to the registry-order assertion |
| `README.md` / `README-en.md` / `README-ja.md` | provider list |

## Why the config helper was generalised

Like Ollama, LM Studio binds loopback only, so `localhost` inside a container points at the container rather than the host. `get_default_ollama_base_url()` carried that resolution inline: probe `host.docker.internal`, fall back to the Linux container default gateway parsed from `/proc/net/route`, then fall back to the hostname anyway.

That logic is identical for any host-local inference server — only the port and the name in the log line differ. Rather than copy roughly twenty lines and let the two drift, it is now `get_default_local_service_base_url(port, service_name)` with two one-line wrappers:

```python
def get_default_ollama_base_url() -> str:
    return get_default_local_service_base_url(OLLAMA_DEFAULT_PORT, "Ollama")

def get_default_lmstudio_base_url() -> str:
    return get_default_local_service_base_url(LMSTUDIO_DEFAULT_PORT, "LM Studio")
```

Ollama's resolved URLs are unchanged in all four environments, and its four existing tests were left untouched precisely so they'd catch a regression here.

`webui/Main.py` had a matching `if llm_provider == "ollama"` branch for the base URL and the Docker hint. That is now a small resolver lookup, so each provider gets its own `llm_provider_tips.<id>.docker_hint` without a second branch.

## One deliberate difference from Ollama

Ollama's branch overwrites `api_key` unconditionally. LM Studio's uses `api_key or "lm-studio"`.

Both need a non-empty placeholder because the OpenAI SDK rejects an empty `api_key`, and neither validates credentials itself. But LM Studio is commonly put behind an authenticating reverse proxy when exposed beyond loopback, and overwriting the key would make that setup impossible to configure. A configured `lmstudio_api_key` is therefore passed through. The field is hidden in the WebUI (`show_api_key=False`) since the local case needs nothing, so this is a `config.toml`-only affordance.

I did not change Ollama's behaviour to match — that is a separate decision and not this PR's to make.

## Locale keys

Only `zh.json` and `en.json` get the tips keys, matching the existing pattern: `test_secondary_locales_do_not_duplicate_provider_tips` forbids `llm_provider_tips.*` in the secondary locales, and `get_llm_provider_tips()` falls back to English for them at runtime.

No `llm_provider_label.lmstudio` key is added, so `default_label` supplies "LM Studio" in every language and there is no label to keep in sync.

## Model name

No default. LM Studio's model IDs depend entirely on what the user has downloaded, so any default would be wrong for almost everyone; `requires_model_name` stays `True` and the existing validation raises a clear error when it is empty. The tips block points at the Developer tab, where the ID is shown.

## Tests

Six new tests in `test/services/test_llm.py`, each observed failing before the implementation existed:

- base URL resolution on the host (`http://localhost:1234/v1`)
- inside Docker Desktop (`host.docker.internal`)
- on native Linux where `host.docker.internal` does not resolve (container default gateway, port 1234 — this is the one that would catch a copy-paste of Ollama's port)
- an explicit `lmstudio_base_url` override winning over container detection
- the placeholder API key when none is configured
- a configured API key being passed through rather than overwritten

Full suite on this branch: `718 passed, 10 skipped, 6084 subtests passed`, `ruff check` clean.

One test fails on my machine both on this branch and on an unmodified checkout of the base commit: `test_webui_task.py::test_worker_logs_are_available_without_streamlit_session_state`. It asserts a log line matches a regex containing `./test/services/...`, and Windows produces `./test\services\...`. It is a path-separator assumption in the test, unrelated to this change, and should pass on Linux CI.


harry0703:
Hi @tlennon-ie, thank you for the thoughtful implementation and the thorough tests. I applied the PR on top of the latest main and verified that Ruff, the focused tests, and the full test suite pass. I also confirmed the LM Studio request path and optional bearer token against a real local OpenAI-compatible HTTP endpoint.

There is one Docker usability issue to address before merging. LM Studio binds to `127.0.0.1` by default, so on native Linux Docker the container gateway URL will still be unreachable unless the user enables **Serve on Local Network** or starts the server with `lms server start --bind 0.0.0.0`. Please mention this in the Docker guidance.

When LM Studio is exposed beyond loopback, authentication is recommended. The implementation supports `lmstudio_api_key`, but the WebUI hides the field and `config.example.toml` does not include it. Could you please either expose it as an optional field while keeping `requires_api_key=False`, or clearly document the config-only option and add `lmstudio_api_key = ""` to the example config?

The rest of the implementation looks good, and I will be happy to retest the update. Thanks again!

tlennon-ie:
Thanks for testing this against a real endpoint — both points are fair, and both are addressed in `bc397b0`.

## Docker guidance

You're right that the gateway URL alone was not a complete answer. The resolution logic finds a routable address for the *host*, but LM Studio isn't listening on it, so on native Linux Docker the connection fails no matter which address is resolved. The guidance now names the bind and both fixes:

> **Reaching it from Docker**: LM Studio binds to `127.0.0.1` by default, so a container cannot connect even through the host gateway. Enable **Serve on Local Network** in the Developer tab, or start the server with `lms server start --bind 0.0.0.0`.

This is in `llm_provider_tips.lmstudio` for both `zh` and `en`, and in the `config.example.toml` block.

I also amended the container-detection hint. It previously read as a promise that the default Base Url would work once a container was detected, which is exactly the misleading case you identified:

> Container environment detected. If Base Url is empty, `http://host.docker.internal:1234/v1` is used by default. **LM Studio must also be listening beyond `127.0.0.1` for that address to connect.**

## Optional API key

I took the first option — exposed as an optional field, `requires_api_key` unchanged at `False`. Hiding it while supporting it was the wrong combination: the setup that needs a token is the one where the user is least likely to be hand-editing `config.toml`.

- `show_api_key` now defaults to `True` for the entry, so the field renders. `requires_api_key=False` means a blank value stays valid and a local server needs no input, and the existing `api_key or "lm-studio"` fallback keeps the placeholder behaviour for the local case.
- `lmstudio_api_key = ""` is in `config.example.toml`, with a comment explaining that LM Studio itself does not check credentials and the value is for a proxy in front of it.
- The tips say the same, so the field doesn't read as a missing credential.

One thing I added beyond the request: once LM Studio listens beyond loopback it's reachable by anything on the network, so the guidance states the authenticating proxy as a recommendation rather than mentioning it in passing. Happy to soften that if it reads as too strong for the WebUI.

## Tests

Four regression tests, each confirmed failing before the change:

- `test_lmstudio_offers_the_api_key_as_an_optional_field` — `requires_api_key` stays `False` while `show_api_key` is `True`, so a future edit can't quietly make the key mandatory or hide it again.
- `test_lmstudio_example_config_offers_the_optional_api_key` — the key is present and empty, so the WebUI field and the example config can't drift apart.
- `test_lmstudio_tips_explain_the_default_loopback_bind` — both locales mention `127.0.0.1`, *Serve on Local Network*, and `lms server start --bind 0.0.0.0`.
- `test_lmstudio_docker_hint_does_not_promise_connectivity_alone` — the container hint mentions the bind.

Full suite: `722 passed, 10 skipped, 6088 subtests passed`, `ruff check` clean.

One disclosure, unchanged from the original PR: `test_webui_task.py::test_worker_logs_are_available_without_streamlit_session_state` fails on my machine on this branch *and* on an unmodified base commit. It asserts a log line matches a regex containing `./test/services/...` and Windows produces `./test\services\...`. It's a path-separator assumption in the test, unrelated to this work, and should pass on Linux CI.

I don't have a Linux Docker host to verify the `--bind 0.0.0.0` path end to end — that part is from LM Studio's documented behaviour and your report rather than my own test, so it's worth a look when you retest. Thanks again for the careful review.

harry0703:
Hi @tlennon-ie, thank you for addressing the previous feedback and for adding the regression tests. I retested the latest commit on top of main, and the full local suite passed: 757 passed and 11 skipped.

There are still a few items to address before merging:

- The PR includes seven unrelated WAV files under resource/comfyui/voices, adding about 4.25 MB. Please remove them from this PR.
- The tips and example config say that LM Studio itself does not check credentials and that the API key is only for a proxy. LM Studio 0.4.0 and later support built-in API tokens through Require Authentication. The implementation already sends the bearer token correctly, so only the wording needs to be updated to cover both the built-in token and proxy cases: https://lmstudio.ai/docs/developer/core/authentication
- The latest commit currently has no CI checks attached. Please rerun or trigger CI after the cleanup.

The provider implementation itself looks good. Once these points are addressed, I will be happy to do a final check. Thanks again!

PR #1259, opened 2026-08-24
fix(webui): decouple upload_post_enabled from auto_upload in settings dialog

### LINKED ISSUE #1258: [Bug]: Opening the Settings dialog rewrites upload_post_enabled and upload_post_auto_upload when the two keys differ
### 问题描述 | Current Behavior

The Auto-Publish settings tab added in #1209 binds two config keys to one checkbox. `webui/Main.py` computes the checkbox value as `upload_post_enabled and upload_post_auto_upload`, then writes **both** keys whenever the checkbox differs from either one:

```python
is_enabled = config.app.get("upload_post_enabled", False)
is_auto = config.app.get("upload_post_auto_upload", False)
ui_state = is_enabled and is_auto

upload_post_enabled = st.checkbox(
    tr("Enable Auto-Publish"),
    value=ui_state,
    key="upload_post_enabled_checkbox"
)
if upload_post_enabled != is_enabled or upload_post_enabled != is_auto:
    _set_runtime_config("app", "upload_post_enabled", upload_post_enabled)
    _set_runtime_config("app", "upload_post_auto_upload", upload_post_enabled)
```

With `upload_post_enabled = true` and `upload_post_auto_upload = false` in `config.toml`, the checkbox renders unchecked (`true and false`), the mismatch branch fires on that same render, and both keys get persisted as `false`. Opening the Settings dialog is enough; the user never touches the checkbox. `st.dialog` renders every tab, so opening Settings to edit a Pexels or LLM key also triggers it.

That key combination is a deliberate configuration, not a corrupt state: `upload_post_enabled = true` with `auto_upload = false` lets an external tool call `cross_post_video` (or gate on `upload_post_enabled`) while the WebUI's automatic post-render upload stays off. `config.example.toml` documents them as independent options. In my setup, a scheduled pipeline publishes through a manual review step; after this change, one visit to the Settings dialog silently disabled publishing, and the failure only surfaced as scheduled posts no-oping.

The reverse direction also loses information: re-ticking the single checkbox from the split state writes `auto_upload = true`, opting the user into automatic publishing they had turned off.

### 重现步骤 | Steps to Reproduce

1. In `config.toml`, set `upload_post_enabled = true` and `upload_post_auto_upload = false`.
2. Launch the WebUI and open the Settings dialog (any tab).
3. Close it without changing anything.
4. `config.toml` now has `upload_post_enabled = false`.

### 预期行为 | Expected Behavior

Rendering the Settings dialog leaves `config.toml` unchanged. Each of the two keys keeps its own value unless the user edits that specific setting.

### 建议修复 | Suggested Fix

Two checkboxes, each bound to one key, each writing only its own key on change:

```python
upload_post_enabled = st.checkbox(
    tr("Enable Upload-Post Integration"),
    value=is_enabled,
    key="upload_post_enabled_checkbox"
)
if upload_post_enabled != is_enabled:
    _set_runtime_config("app", "upload_post_enabled", upload_post_enabled)

upload_post_auto_upload = st.checkbox(
    tr("Enable Auto-Publish"),
    value=is_auto,
    key="upload_post_auto_upload_checkbox"
)
if upload_post_auto_upload != is_auto:
    _set_runtime_config("app", "upload_post_auto_upload", upload_post_auto_upload)
```

I run this as a local patch, including an update to `test_webui_upload_post_checkbox_syncs_persisted_mismatch` asserting the dialog leaves a split config untouched. Happy to submit it as a PR.

### 版本信息 | Version

main @ `110997c`, macOS, Python 3.11.

mcgeand:
Fixes #1258.

## Problem

The Auto-Publish settings tab renders one checkbox for two config keys. It computes the checkbox value as `upload_post_enabled and upload_post_auto_upload`, then writes **both** keys whenever the checkbox differs from either. A config with `upload_post_enabled = true` and `upload_post_auto_upload = false` (external tooling uses the Upload-Post integration while post-render auto-upload stays off) mismatches on every render, so opening the Settings dialog persists both keys as `false` without any click. Re-ticking the box from that state sets `auto_upload = true`, which the user had turned off. Full analysis and reproduction in #1258.

## Change

- `webui/Main.py`: two checkboxes, one per key. Each reflects its own saved value and writes only its own key on change.
- `test/services/test_webui_upload_post_settings.py`: the checkbox test now asserts that opening Settings preserves a split config (`enabled = true` / `auto_upload = false`) and that toggling auto-upload leaves `enabled` untouched.
- `webui/i18n/en.json` + `test/services/test_webui_i18n.py`: new label `Enable Upload-Post Integration`, registered in `ENGLISH_FALLBACK_KEYS` so secondary locales fall back to English per the existing convention.

## Testing

Full suite passes with the change: `744 passed, 11 skipped, 6951 subtests` (`python -m pytest test -q`). The rewritten AppTest case renders the real dialog against a split config and fails on the pre-fix behavior.

https://claude.ai/code/session_01AcyHNEBEz7V7LdDz2L6wWQ

### DIFF
--- a/test/services/test_webui_i18n.py
+++ b/test/services/test_webui_i18n.py
@@ -31,6 +31,7 @@ ENGLISH_FALLBACK_KEYS = frozenset(
         "Confirm AI Video Charge Help",
         "Confirm AI Video Charge Required",
         "Custom API Endpoint",
+        "Enable Upload-Post Integration",
         "API Platform",
         "llm_provider_endpoint_selector.moonshot",
         "llm_provider_endpoint_selector_help.moonshot",
--- a/test/services/test_webui_upload_post_settings.py
+++ b/test/services/test_webui_upload_post_settings.py
@@ -34,8 +34,11 @@ def test_webui_upload_post_setup_guide_links_to_required_pages():
         assert "https://app.upload-post.com/manage-users" in setup_guide.value
 
 
-def test_webui_upload_post_checkbox_syncs_persisted_mismatch():
-    # Persisted mismatch: enabled=True, auto_upload=False
+def test_webui_upload_post_checkboxes_stay_decoupled():
+    # enabled=True / auto_upload=False is a deliberate split: an external
+    # pipeline may call Upload-Post while auto-publish stays off. Opening the
+    # settings dialog must not rewrite either key, and each checkbox must
+    # reflect and write only its own key.
     test_app_config = dict(
         config.app,
         upload_post_enabled=True,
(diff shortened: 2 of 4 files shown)


harry0703:
Thanks for the quick fix. I verified the decoupled controls locally across all four config combinations, and the full test suite and CI pass.

There is one small issue before merging: `Enable Upload-Post Integration` is currently added to `ENGLISH_FALLBACK_KEYS`, so it appears in English in every non-English UI. Since this is a stable settings label, could you please add translations for the supported locales and remove it from the fallback list?

Thanks again!

harry0703:
Thanks again for the fix and the clear regression test. I have merged the PR. We will add the remaining translations on our side.

PR #1190, opened 2026-08-18
docs: clarify video material sources

HaningZS:
## Summary

- explain how Pexels, Pixabay, Coverr, and local files supply existing visual material
- document both paid AI-generation paths: WaveSpeed AI and Shengsuan Cloud / LoomLoom
- add equivalent source explanations to the Chinese, English, and Japanese READMEs
- preserve the current feature-list wording and sponsor sections

## Why

The introduction can lead readers to expect that every workflow synthesizes new video frames. The new overview makes the selected video source and shared post-production pipeline explicit, sets license expectations for stock and local material, and covers both AI video integrations currently available on `main`. It also clarifies the source question raised in #1181 without claiming support for arbitrary creator or channel URLs.

## Testing

- rebased onto current `upstream/main` (`05c0baf`)
- docs-only boundary check: 33 insertions, 0 deletions across `README.md`, `README-en.md`, and `README-ja.md`
- `git diff --check upstream/main...HEAD`
- `uv run --no-sync python -m compileall -q app cli.py main.py webui test`
- `uv run --no-sync ruff check app cli.py main.py webui test`
- `uv run --no-sync python -X utf8 -m coverage run -m pytest -q test`: 712 passed, 11 skipped, 6084 subtests passed

Closes #1183

PR #1262, opened 2026-08-25
docs: clarify visual material sources

### LINKED ISSUE #1183: Suggestion: Clarify the video generation mechanism in README
### Background

MoneyPrinterTurbo is presented as an “AI short video generation tool” and provides an impressive automated workflow from topic/keywords to a finished video.

However, I think the current README may cause some users to misunderstand how the actual video generation works.

In particular, users familiar with modern AI video generation models such as **Seedance, Wan, Kling, or Sora** may interpret “AI video generation” as meaning that the system generates new video frames/content from text or images using a generative video model.

From the current project description, however, the core workflow appears to be substantially different:

* AI models are used to generate the video script and search keywords.
* The system searches for / retrieves existing video footage from sources such as **Pexels, Pixabay, and Coverr**.
* These existing video clips are then selected, arranged, subtitled, combined with TTS and background music, and rendered into the final video.
* In other words, the main visual content is **retrieved from existing footage and then edited/composited**, rather than being generated pixel-by-pixel by a text-to-video or image-to-video foundation model.

This is a perfectly legitimate and useful approach, but it is technically different from generative video models.

### Why this matters

The distinction is important because users may otherwise expect MoneyPrinterTurbo to:

> Prompt → AI generates completely new video content

while the actual workflow is closer to:

> Prompt → AI generates script/search keywords → retrieve existing footage → automatically edit/composite → final video

These are two different categories of video-generation systems.

For example, **Seedance and Wan** are generative video models whose primary purpose is to synthesize new video content from text/images. MoneyPrinterTurbo, by contrast, is primarily an **AI-assisted automated video creation/editing pipeline based on retrieved video assets**.

### Suggested README clarification

I suggest adding a prominent note near the beginning of the README, for example:

> **Important: MoneyPrinterTurbo is not a text-to-video (T2V) or image-to-video (I2V) generative video model such as Seedance or Wan.**
>
> MoneyPrinterTurbo is an AI-assisted automated video creation tool. It uses LLMs to generate scripts and search keywords, retrieves existing video footage from sources such as Pexels, Pixabay, and Coverr, and then automatically edits/composites the footage with subtitles, voice-over, and background music to produce the final video.
>
> If you are looking for a model that directly generates new video content from text or images, please use a dedicated video generation model such as Seedance, Wan, Kling, Sora, etc.

I think adding such a clarification would make the project's technical positioning much clearer and help users form the correct expectations before trying it.

This is not a criticism of the implementation itself—the automated retrieval + editing workflow is useful. The suggestion is mainly about making the distinction between **AI-assisted video editing/composition** and **generative video models** explicit in the README.

Suggestion: Clarify the video generation mechanism in README

ketpatil77:
## Summary

- explain that the default workflow retrieves local or stock footage and edits it automatically
- distinguish the optional WaveSpeed AI text-to-video material source
- keep the clarification aligned across Chinese, English, and Japanese READMEs

Closes #1183.

## Verification

- `git diff --check`
- reviewed rendered Markdown structure in all three README variants


ketpatil77:
Closing as a duplicate of #1190. That earlier PR already covers the same three README files, includes both supported AI video sources, has full local verification, and has incorporated maintainer feedback. Keeping one focused implementation avoids splitting review effort.

PR #1266, opened 2026-08-26
fix(api): omit Redis auth segment when password is not configured

Mihir7027:
## Summary

- When `redis_password` is unset (the default `None`), the previous f-string
  produced `redis://:None@host:port/db` — Python's `None` got stringified verbatim,
  so redis-py sent `AUTH None` to Redis and failed on every password-less server.
- Extract a small `_build_redis_url()` helper that only includes the `:password@`
  auth segment when a password is actually provided.
- Covers `None`, empty string (both produce a no-auth URL), and a real password
  value with unit tests.

## Test plan

- [ ] `TestBuildRedisUrl.test_no_password_omits_auth_segment` — `None` and `""` both produce `redis://localhost:6379/0`
- [ ] `TestBuildRedisUrl.test_password_is_included_in_url` — real password produces correct URL
- [ ] Existing `TestVideoControllerTasks` and `TestRedisTaskManager` suites still pass

### DIFF
--- a/app/controllers/v1/video.py
+++ b/app/controllers/v1/video.py
@@ -48,7 +48,13 @@ _redis_password = config.app.get("redis_password", None)
 _max_concurrent_tasks = config.app.get("max_concurrent_tasks", 5)
 _max_queued_tasks = config.app.get("max_queued_tasks", 100)
 
-redis_url = f"redis://:{_redis_password}@{_redis_host}:{_redis_port}/{_redis_db}"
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for the fix. I verified it with a real passwordless Redis container, including connection, read/write, and task queue round-trip tests. Everything worked as expected, and the PR has been merged. Appreciate your contribution!

PR #1267, opened 2026-08-26
fix(llm): suppress spurious "trying again" warning on the final retry attempt

Mihir7027:
generate_script and generate_terms guarded the "trying again…" log with if i < _max_retries:. Since the loop is for i in range(_max_retries), i maxes out at _max_retries - 1, making the condition always True — the misleading warning fired even on the last attempt when no retry would follow.

Changed the guard to if i < _max_retries - 1:, matching the identical pattern already present in generate_social_metadata (line 958). Added unit tests verifying the warning fires exactly _max_retries - 1 times when every attempt fails.

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -569,7 +569,7 @@ def generate_script(
         except Exception as e:
             logger.error(f"failed to generate script: {e}")
 
-        if i < _max_retries:
+        if i < _max_retries - 1:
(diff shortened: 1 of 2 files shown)


harry0703:
Thank you for this focused fix. We verified the retry boundaries for both script and search-term generation, including single-attempt, exhausted-retry, and eventual-success cases. The warning count now matches the actual number of retries, and the PR has been merged. Appreciate your contribution!

PR #1268, opened 2026-08-26
fix(voice): anchor SiliconFlow subtitle timeline end to full audio duration

Mihir7027:
## Problem

`siliconflow_tts` built its subtitle offsets with a hand-rolled loop:

    sentence_duration = int(sentence_chars * char_duration)

Accumulated integer truncation across multiple sentences means the last
subtitle always ends a few 100-nanosecond units before the actual audio
end. Users see a subtitle gap at the tail of every SiliconFlow-narrated
video.

## Fix

Replaced the ad-hoc loop with `populate_legacy_submaker_with_full_text`,
the shared helper that all other providers (`gemini_tts`, `mimo_tts`,
`minimax_tts`) already use. It explicitly anchors the last subtitle entry
to the real `audio_duration_100ns`, eliminating the truncation gap.

Also removed a redundant local `from moviepy import AudioFileClip` import
(already imported at module level) and wrapped `audio_clip.close()` in a
`finally` block so the handle is always released.

## Tests

Added a unit test that mocks the HTTP response and `AudioFileClip`,
calls `siliconflow_tts` with a multi-sentence script, and asserts that
`offsets[-1][1] == int(audio_duration_seconds * 10_000_000)`.

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -948,74 +948,24 @@ def siliconflow_tts(
                 with open(voice_file, "wb") as f:
                     f.write(response.content)
 
-                # 这里仍然沿用项目原有的字幕结构，因此需要补齐旧字段。
                 sub_maker = ensure_legacy_submaker_fields(SubMaker())
 
-                # 获取音频文件的实际长度
                 try:
-                    # 尝试使用moviepy获取音频长度
-                    from moviepy import AudioFileClip
-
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the fix and the regression test! I rechecked it against the latest main, and the full test suite passes. Merged.

PR #1269, opened 2026-08-26
fix(tts): ensure AudioFileClip is always closed in ElevenLabs, Chatterbox, and Fish Audio

Mihir7027:
## Summary
- `elevenlabs_tts`, `chatterbox_tts`, and `fish_audio_tts` each open an
  `AudioFileClip` (spawning an FFmpeg subprocess) and call `.close()` in
  plain sequential code, with no `try/finally`.  If anything raises between
  construction and close the subprocess leaks.
- The pre-existing `_write_validated_minimax_audio` already wraps
  `.duration` in `try/finally`; this PR extends that pattern to the three
  remaining providers.

## Changes
- Wrapped `audio_clip.duration` in `try/finally: audio_clip.close()` in
  `elevenlabs_tts`, `chatterbox_tts`, and `fish_audio_tts`
  (`app/services/voice.py`)
- Added three unit tests that mock `.duration` to raise a `RuntimeError`
  and assert `.close()` is still called

## Test plan
- [ ] `test_elevenlabs_tts_audio_clip_closed_on_duration_error` passes
- [ ] `test_chatterbox_tts_audio_clip_closed_on_duration_error` passes
- [ ] `test_fish_audio_tts_audio_clip_closed_on_duration_error` passes
- [ ] Existing ElevenLabs/Chatterbox TTS success tests still pass

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -1686,8 +1686,10 @@ def elevenlabs_tts(
                 f.write(response.content)
 
             audio_clip = AudioFileClip(voice_file)
-            audio_duration = audio_clip.duration
-            audio_clip.close()
+            try:
+                audio_duration = audio_clip.duration
+            finally:
+                audio_clip.close()
 
             sub_maker = ensure_legacy_submaker_fields(SubMaker())
             logger.success(f"elevenlabs tts succeeded: {voice_file}")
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for fixing these resource cleanup paths and adding regression tests for all three providers. I verified the change against the latest main, and the focused and full test suites pass. Merged!

PR #1270, opened 2026-08-26
fix(llm): use non-greedy quantifier when stripping bracket and paren groups from script

Mihir7027:
## Summary
- `format_response` in `generate_script` used `re.sub(r"\[.*\]", ...)` and
  `re.sub(r"\(.*\)", ...)`. The greedy `.*` matches from the first opener
  to the **last** closer on the same line, deleting all text in between.
- Examples of silent data loss:
  - `"[Intro] Great content [end]"` → `"."`
  - `"Save (at least) 10% of income (monthly)."` → `"Save ."`
- Replaced `.*` with `.*?` so each bracket/paren group is removed
  independently, leaving surrounding words intact.

## Test plan
- [ ] `test_generate_script_strips_each_bracket_group_independently` passes
- [ ] Existing `test_generate_script_sends_custom_prompt_to_llm` passes

### DIFF
--- a/app/services/llm.py
+++ b/app/services/llm.py
@@ -536,9 +536,11 @@ def generate_script(
         response = response.replace("*", "")
         response = response.replace("#", "")
 
-        # Remove markdown syntax
-        response = re.sub(r"\[.*\]", "", response)
-        response = re.sub(r"\(.*\)", "", response)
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for catching this silent script truncation issue and adding a focused regression test. I verified the fix against the latest main, including several bracket and parenthesis cases, and the full test suite passes. Merged!

PR #1276, opened 2026-08-27
feat(material): add native Volcano Engine Ark Seedance

### LINKED ISSUE #1271: [Feature]: Add native Volcano Engine Seedance API integration
### 需求描述 | Problem Statement

MoneyPrinterTurbo does not currently provide a native integration entry for Seedance through Volcano Engine. The Seedance API is publicly available, but users cannot use Volcano Engine's official capability directly from this project. They must rely on indirect integrations or maintain their own adapter, which increases configuration, deployment, and maintenance costs.

This is a focused request for one provider, following the scope guidance in #854. A previous Seedance/Jimeng API request was discussed in #830, where the concern was the size and maintenance scope of a first-class integration.

Could you clarify the main reasons why the native Volcano Engine Seedance entry has not been added yet (for example, API stability, authentication, regional or quota limitations, or maintenance cost)? Is there a plan or timeline to support it in a future release?🐱🐱

Please consider adding Volcano Engine Seedance as an optional video-generation provider, using the official public API and the project's existing provider configuration pattern:

- Allow Access Key, Secret Key, region, and model/version parameters to be configured through environment variables or a config file.
- Implement task submission, status polling, result retrieval, and clear error handling.
- Expose provider selection in the UI or configuration while keeping the current implementation as the default or fallback.
- Document account setup, authentication, quotas, and usage limitations in the README.

The integration could initially be kept narrowly scoped to the core API flow, with optional UI support added incrementally. If there is no short-term plan, it would still be helpful to understand the technical or product considerations so the community can assess whether a focused contribution would be feasible.

lihuiyang1024:
## Summary

- add the official Volcano Engine Ark Seedance asynchronous text-to-video API as an opt-in material provider
- support private `config.toml`, the existing VolcEngine Ark LLM key, `VOLCENGINE_ARK_API_KEY`, and `ARK_API_KEY`
- make Ark base URL, model/Endpoint ID, resolution, duration bounds, polling interval, timeout, and watermark configurable
- map existing MoneyPrinterTurbo prompts, aspect ratios, and clip durations to Ark create-task requests
- poll the same task ID to a terminal state, download signed output URLs with bounded retries, and persist sanitized material provenance
- expose the provider through API task parameters, CLI/batch mode, WebUI settings, and the Agent Skill
- document setup and limitations in English, Chinese, and Japanese

## Paid-task safety

Ark video generation is billed per submitted task. This integration therefore:

- never retries task submission after a network error or 5xx because a paid task may already exist
- retries only polling of the same remote task ID for transient 429/5xx/network failures
- records the remote task ID in logs and task failure state when the outcome is unknown
- stops later paid submissions after an unconfirmed task or generated-video download failure
- requires explicit `--confirm-seedance-charge` in CLI/batch mode and an unchecked-by-default confirmation in WebUI
- redacts Ark keys and proxy credentials from request errors

## Scope

The initial provider supports text-to-video only. Image/video/audio references and Seedance 2-specific tools are intentionally left for follow-up work so the first integration remains isolated and maintainable. Existing material providers are unchanged.

Official API references:

- Create task: https://www.volcengine.com/docs/82379/1520757
- Query task: https://api.volcengine.com/api-docs/view?action=GetContentsGenerationsTask&serviceCode=ark&version=2024-01-01

## Tests

- changed-area regression: `266 passed, 2 skipped, 6747 subtests`
- Seedance protocol/material/task tests: `17 passed` plus retry, timeout, redaction, charge, and failure-path subtests
- complete CLI suite: `81 passed, 20 subtests`
- WebUI provider/confirmation integration, settings backup, config, i18n, and Agent Skill tests pass
- full Ruff, compileall, and diff checks pass

The suite uses deterministic HTTP mocks and performs no billed calls. I do not have a billed Ark account, so real-account validation is still required before merge.

Closes #1271.


harry0703:
Thanks for the contribution! We’ve incorporated and further validated the Seedance integration on main, while preserving your authorship in the final commit. The implementation has passed the full CI suite and real-account generation testing. Thanks again for your work!

PR #1287, opened 2026-08-30
feat(video): add configurable source clip fit mode

emanuelsc3s:
## Summary

- add a `VideoFitMode` with `cover` and `contain` options
- use centered cover/crop by default so mismatched source clips fill the output canvas
- retain the previous letterbox behavior through the `contain` option
- expose the setting in the WebUI, CLI, API/task schema, saved presets, and restored tasks
- add localized WebUI labels and regression coverage for rendering and persistence

## Motivation

When a landscape source clip is used for a portrait output, the current compositor scales the complete source frame down and places it on a black portrait canvas. For a 16:9 source rendered as 9:16, this leaves large black bars above and below the footage.

This PR adds an explicit fit policy instead of hard-coding that behavior. Users can choose between filling the portrait canvas or preserving the complete source frame.

## Behavior

| Mode | Behavior |
| --- | --- |
| `cover` | Scale until the output canvas is fully covered, then crop excess edges symmetrically from the center. |
| `contain` | Preserve the complete source frame and add black bars when the aspect ratios differ. |

The fit operation runs on each source clip before transitions, subtitles, and final audio mixing. It never stretches the image.

## User-facing changes

- WebUI: adds **Source Frame Fit** with **Fill (center crop)** and **Fit (black bars)** options.
- WebUI preferences: persists the selected fit mode and restores it in new sessions.
- Settings presets and task restoration: round-trip the new field.
- CLI: adds `--video-fit-mode {cover,contain}`.
- API/task schema: adds `video_fit_mode` with `cover` as the default.
- Example configuration: documents `[ui].video_fit_mode`.

## Compatibility and design note

The new `combine_videos` argument is appended to the function signature so existing positional callers keep their previous argument mapping. The previous rendering behavior remains available as `contain`.

This PR intentionally changes the default for mismatched aspect ratios to `cover`, because it avoids letterboxing in the common landscape-to-portrait workflow. If preserving the historical default is preferred, the default can be switched to `contain` while keeping the new user-facing control and implementation.

## Validation

Commands run locally using the same checks as CI:

```bash
uv run --no-sync python -m compileall app cli.py main.py webui test
uv run --no-sync ruff check app cli.py main.py webui test
uv run --no-sync python -X utf8 -m coverage run -m pytest -q test
uv run --no-sync python -m coverage report
```

Results:

- `828 passed, 11 skipped`
- `7,321` subtests passed
- total coverage: `79%`
- real 16:9 material validated from `1280x720` to `1080x1920`
- FFmpeg crop detection reported the full output canvas: `crop=1080:1920:0:0`


harry0703:
Thank you for the contribution! I tested both cover and contain through a full WebUI video generation flow, and both worked as expected. Merged.

PR #1288, opened 2026-08-31
fix(webui): avoid Windows-reserved download names

Sushanth012:
## Summary
- detect Windows device names after sanitizing video download subjects
- prefix reserved names such as `CON`, `AUX`, and `LPT1` so downloads remain cross-platform safe
- cover reserved names with focused task-history tests

## Testing
- `uv run --no-sync python -X utf8 -m pytest -q test/services/test_webui_task_history.py test/services/test_webui_task.py` (25 passed)
- `uv run --no-sync ruff check webui/Main.py test/services/test_webui_task_history.py test/services/test_webui_task.py`
- `uv run --no-sync python -m compileall webui test/services/test_webui_task_history.py`

### DIFF
--- a/test/services/test_webui_task.py
+++ b/test/services/test_webui_task.py
@@ -156,6 +156,7 @@ def test_completed_task_renders_subject_named_video_download(
     selected_nodes = []
     target_names = {
         "_DOWNLOAD_FILENAME_INVALID_PATTERN",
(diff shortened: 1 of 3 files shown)


harry0703:
Thank you for the contribution! This has been merged. We also added a small follow-up to cover the remaining Windows reserved-name variants and expanded the regression tests.

PR #1316, opened 2026-09-04
feat: trim long TTS silence gaps and shift subtitles

Efe-Tuncel:
Edge TTS leaves ~1.2s pauses after each sentence, so short-sentence scripts get seconds of dead air. This trims inter-cue gaps over 0.6s down to 0.3s and shifts later subtitle cues back so audio and subtitles stay in sync. Any failure keeps the original files and the pipeline continues. New trim_silence flag on VideoParams (default true). Tests: test/services/test_silence.py (5 passed).


Efe-Tuncel:
\Author request: closing"

PR #1317, opened 2026-09-04
feat: two-lane stock terms with safe fallback

Efe-Tuncel:
Stock pools have no brand, person, logo, or chart footage, so proper-noun search terms return generic irrelevant clips. This makes search terms stock-safe and adds a fallback net. Two lanes: generic filmable scenes go to stock search, specific facts go to overlays. Changes: STOCK-SAFE rule in the term prompt, bad-term guard with dedup plus one round of LLM replacements in generate_terms, replacement-candidate retry in material download (both plain and script-order modes), match_materials_to_script default true (opt out any time). All fallbacks are fail-safe and keep existing candidates on error. Tests: test/services/test_stock_terms.py (7 passed), test_schema and test_task green.


Efe-Tuncel:
\Author request: closing"

PR #1319, opened 2026-09-04
feat(llm): add Claude Code provider using a Claude subscription

fred-ng:
Claude Pro / Max / Team subscriptions do not issue API keys, and their credentials are only usable by Anthropic's own clients, so the existing `anthropic` provider is unavailable to subscription-only users.

This adds a `claude_code` provider that shells out to the locally installed, logged-in `claude` CLI in headless mode (`claude -p --output-format json`) instead of calling the HTTP API, letting the CLI perform authentication. Requests consume subscription usage rather than API credit.

Details:
- Registry entry needs no API key, no base URL, and no model name; an empty model name follows whatever default the CLI currently uses.
- The adapter replaces Claude Code's coding-agent system prompt with a copywriter prompt, disables all tools and external MCP config, and runs in a temporary working directory so a stray CLAUDE.md cannot leak into script output. Each call is a deterministic single-turn text generation.
- Failures are reported from the CLI's JSON payload (`is_error` / `result`) rather than the exit code alone, since auth and quota errors arrive as a non-zero exit with a readable message on stdout.
- Dockerfile.claude layers Node.js and the claude CLI onto the published image; docker-compose.claude.yml mounts the project and injects CLAUDE_CODE_OAUTH_TOKEN, which is how the CLI authenticates in a container where no keychain is available.


harry0703:
Thanks for contributing this! Reusing a local Claude Code login is a useful option. I reviewed `4f46943` and ran the relevant tests locally. A few things need addressing before merging:

1. The subprocess inherits API credentials and provider settings from its environment. Claude Code can prioritize these over subscription login, so the current claim that calls never consume API credit isn't guaranteed. Please detect conflicting authentication settings or explicitly isolate the subscription authentication path.
2. The tool denylist doesn't disable every tool, and a temporary working directory doesn't isolate user-level settings, hooks, or skills. Please use `--tools ""` and explicitly control configuration/skill loading so these calls stay text-only on the supported CLI versions.
3. `claude_code_timeout = 300` in TOML fails with `'int' object has no attribute 'strip'`. Please accept numeric values as well as strings and reject non-finite values such as `nan` and `inf`.
4. The relevant test run returned **87 passed, 3 failed, 1 skipped**; Ruff passed. The failures concern the provider list, duplicated timeout default, and provider help template. Please update those checks appropriately and add adapter tests for success, missing CLI/login, exhausted quota, timeout, and malformed output.

I haven't verified a real subscription call or the Docker image yet. A successful end-to-end example with the CLI version and authentication method would also help. Thanks again!

harry0703:
Thanks for the updates and the added tests! I rechecked `7de58d2`: the relevant tests now pass (107 passed, 1 skipped), and Ruff passes too. I also confirmed that Claude Code 2.1.260 supports the isolation flags.

Two small gaps remain:

1. Please address the remaining subscription-auth isolation gap before merging: `CLAUDE_CODE_USE_FOUNDRY` and its credentials still reach the subprocess. On a machine configured for Foundry, this can select cloud billing instead of subscription login, or fail authentication. Please cover this path in the environment filtering and regression tests.
2. A minor timeout edge case: numeric `0` and `false` are replaced by the 300-second default by `or field.default_value` before validation, while string `"0"` is rejected. Please apply the default only to empty values so validation stays consistent. This isn't a blocker on its own.

I haven't verified a real subscription generation or the Docker image yet. Thanks again for following up!

fred-ng:
Thanks for your comments. I tried with my docker and claude code team subscription that working as expected. Thanks for the amazing tool.
image

harry0703:
Merged! Thanks for adding this provider and patiently addressing the review feedback. The full local test suite passed (1000 passed, 11 skipped), and I checked the WebUI settings too. I appreciate the authentication isolation improvements and the regression coverage.

PR #1321, opened 2026-09-05
Fix webui.sh port check breaking when project path contains spaces

tailieu2212:
## What changed

`webui.sh`'s port probe called `find_available_port $PORT_CHECK_CMD` with an **unquoted** variable holding the venv's python path. When the project directory path contains spaces (e.g. a folder named `My Project/MoneyPrinterTurbo`), shell word-splitting breaks that single path into several bogus arguments, so the Python port-probe script never actually runs and `webui.sh` fails with:

```
***** No available WebUI port found in 8501-8599 for 127.0.0.1. *****
```

even though port 8501 is free.

## Why

Ran into this while setting up MoneyPrinterTurbo locally in a folder with spaces in its path. `webui.sh` worked fine when invoking the venv's `streamlit` binary directly, which confirmed the port-check subshell itself was the broken part.

## Fix

Replaced the single `PORT_CHECK_CMD` string with a small `find_port()` wrapper function defined in each of the three python/uv/streamlit detection branches. Each wrapper calls `find_available_port` with its arguments already correctly quoted/split (the venv path as one argument, `uv run python` as three), so paths with spaces are no longer mangled by word-splitting.

## Testing

- `sh -n webui.sh` — syntax OK
- Ran `./webui.sh` from a project directory with spaces in its path: previously failed with "No available WebUI port found"; after the fix it correctly reports `WebUI address: http://127.0.0.1:8501` and Streamlit starts and serves (verified with `curl` returning HTTP 200).
- Ran `./webui.sh` from a path without spaces as a regression check: still works as before.

No functional changes for users on paths without spaces.

### DIFF
--- a/webui.sh
+++ b/webui.sh
@@ -17,15 +17,15 @@ MPT_WEBUI_HOST="${MPT_WEBUI_HOST:-127.0.0.1}"
 MPT_WEBUI_PORT="${MPT_WEBUI_PORT:-8501}"
 
 if [ -x "$CURRENT_DIR/.venv/bin/python" ]; then
-  PORT_CHECK_CMD="$CURRENT_DIR/.venv/bin/python"
   set -- "$CURRENT_DIR/.venv/bin/python" -m streamlit
+  find_port() { find_available_port "$CURRENT_DIR/.venv/bin/python"; }
 elif command -v uv >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


harry0703:
Thanks for the fix! I verified startup locally with paths containing spaces, normal paths, and an occupied preferred port. Everything worked as expected. Merged — thanks for contributing!

PR #1323, opened 2026-09-06
Add NVIDIA NIM as an LLM provider, with model auto-load

Searchadx360:
## Summary
- Adds NVIDIA NIM (`https://integrate.api.nvidia.com/v1`) as a new LLM provider option, using the existing OpenAI-compatible adapter, with EN/ZH configuration tips, config.example.toml entries, README updates, and tests.
- Generalizes the Groq-only "load available models" dropdown into a registry-driven `supports_model_list` flag so any OpenAI-compatible provider (now Groq and NVIDIA NIM) can offer a model picker fetched from `GET {base_url}/models`, instead of requiring users to hand-type model IDs. Falls back to free text entry if the list can't be loaded.
- Fixes a pre-existing bug where the headless "Open Task Folder" toast showed a backslash-separated path on Windows hosts (`os.path.relpath` returns OS-native separators, but the message is hardcoded to a `/`-joined form).
- Forces UTF-8 stdout/stderr in `cli.py`/`webui.bat`/`webui.sh` so loguru's Unicode log symbols don't crash on Windows' default codepage.

## Test plan
- [x] `pytest test/` — 768 passed, 10 skipped, 0 failed
- [x] Manually smoke-tested via Streamlit `AppTest`: selecting NVIDIA NIM in the LLM Provider dropdown renders correctly (tips, API key/base URL/model fields, model-list fallback caption when no key is set)
- [ ] Live NVIDIA NIM API key not available in this environment — the `GET /models` auto-load path reuses Groq's existing, already-proven code path, but hasn't been exercised against NVIDIA's real endpoint

Note: NVIDIA's `wan-ai/wan2.2-animate-2-14b` NIM is a video-generation model, not a chat model, so it intentionally isn't wired in as a model choice here (this app only calls `/v1/chat/completions`).


Searchadx360:
Closing this for now. Thanks for the project!

PR #1326, opened 2026-09-06
fix(docker): migrate base image from EOL Debian 11 (bullseye) to Debian 12 (bookworm)

eualanteixeira:
## Problem

Debian 11 "bullseye" reached the end of its LTS support on 2026-08-31. As a result, `deb.debian.org/debian-security` has started dropping some bullseye-security package files (observed on arm64: `libtiff5`, `libavfilter7`, `libgbm1`), which breaks `docker compose build` with 404 errors while installing `git`/`ffmpeg`.

`archive.debian.org`, the usual fallback for EOL Debian releases, does not (yet) host an archived bullseye-security tree either, so pointing apt at it cannot satisfy the security-patched dependency versions that `git`/`ffmpeg` need there — apt fails with "you have held broken packages" instead.

This currently breaks a fresh `docker compose up -d --build` for anyone building the image today.

## Fix

- Bump the base image from `python:3.11-slim-bullseye` to `python:3.11-slim-bookworm` (Debian 12, currently fully supported with live mirrors).
- Update the mirror fallback chain (Aliyun -> Tsinghua -> official Debian mirrors) to use bookworm's suites.
- Add a retry loop at every mirror tier instead of only the first one, for resilience against transient CDN hiccups.
- Set `DOCKER_BUILD_MIRROR: default` as a build arg in `docker-compose.yml`, since users outside China otherwise waste build time waiting on mirrors they can't reach before falling through to the official Debian mirror.

## Testing

`docker compose build --no-cache` succeeds for both the `webui` and `api` services on Apple Silicon (arm64) macOS with Docker Desktop, and the WebUI comes up correctly at `http://127.0.0.1:8501`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Use an official Python runtime as a parent image
-FROM python:3.11-slim-bullseye
+FROM python:3.11-slim-bookworm
 
 # Set the working directory in the container
 WORKDIR /MoneyPrinterTurbo
@@ -23,7 +23,7 @@ RUN set -u; \
     write_debian_sources() { \
         main_url="$1"; \
         security_url="$2"; \
-        printf 'deb %s bullseye main\ndeb %s bullseye-updates main\ndeb %s bullseye-security main\n' \
+        printf 'deb %s bookworm main\ndeb %s bookworm-updates main\ndeb %s bookworm-security main\n' \
             "$main_url" "$main_url" "$security_url" > /etc/apt/sources.list; \
         rm -rf /var/lib/apt/lists/*; \
     }; \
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for updating the base image and testing the build on Apple Silicon! Moving to Bookworm makes sense. Two small adjustments before merging:

1. Bookworm also ships `/etc/apt/sources.list.d/debian.sources`. Writing only `/etc/apt/sources.list` leaves the original HTTP sources active alongside the selected mirror. Could you update that file, or remove it before writing the replacement sources, so each fallback actually switches sources?
2. Please make the Compose mirror argument configurable instead of hard-coding `default`, ideally `${DOCKER_BUILD_MIRROR:-china}` to preserve the existing default while allowing users outside China to select the official mirrors.

Our local Docker engine is currently unavailable, so we have not independently verified the container build yet. Thanks again for tackling this!

PR #1328, opened 2026-09-07
feat: improve Shengsuan Cloud video generation

huanglvjing:
## Summary

  Improve the Shengsuan Cloud (LoomLoom) AI video generation workflow and make it safer, clearer, and more reliable.

  ## Changes

  - Load available AI video models dynamically from the current account capability profile.
  - Display locally maintained reference prices for known models while keeping unknown/new models fully compatible.
  - Automatically refresh quotes when billable parameters change.
  - Preserve the original single-click script and keyword generation workflow without exposing unnecessary LoomLoom script quotation
  controls.
  - Automatically recommend the number of AI video materials based on the estimated voiceover duration.
  - Recalculate the recommendation after an accurate full voiceover preview becomes available.
  - Display estimated material coverage and the expected black-screen shortfall.
  - Append black-screen padding when generated footage is shorter than the narration instead of looping paid AI-generated clips.
  - Improve LoomLoom polling, timeout handling, validation, payment confirmation, and retry behavior.
  - Preserve upstream video fitting and newly added provider functionality during the rebase.
  - Add and update service, video-processing, WebUI, and localization tests.

  ## Compatibility

  - Existing non-LoomLoom video sources keep their original looping behavior.
  - Newly added or unknown Shengsuan Cloud models remain selectable; they simply do not display a local reference price.
  - Manual material-count adjustments are preserved after the automatic recommendation is applied.

  ## Testing

  - 183 tests passed
  - 3 tests skipped
  - 8,133 localization subtests passed
  - Ruff, Python compilation, and Git diff checks passed


harry0703:
Thanks for improving the Shengsuan Cloud integration! Dynamic model selection and automatic quotes are useful additions. I tested this locally and found a few issues to address before merging:

1. Selecting “Shengsuan Cloud Batch (Paid)” still calls the current LLM provider because `_render_script_settings()` now always calls `_render_local_script_generation()`. With OpenAI configured, it calls OpenAI rather than LoomLoom. In that combination, selecting LoomLoom video also hides the independent API key input. Please preserve the existing batch-script workflow and keep its removal out of this PR.
2. Black padding can appear even when there is enough footage. In a real render with a 6-second source, 5-second narration, sequential mode, and a 3-second clip limit, only the first 3 seconds of footage were used; the rest was black. Please use the remaining footage before treating it as a shortfall. I would also prefer to keep the existing looping default and discuss black padding separately.
3. `test_loomloom_video_source_quotes_then_passes_secret_in_confirmed_request` fails on a clean configuration: it expects two quote calls, but the initial scene count is already 1, so setting it to 1 again produces only one call. Please set the initial UI configuration explicitly in the test.

The related suite returned 169 passed, 1 failed, and 3 skipped, with 8,133 localization subtests passing. This included local rendering checks, but no live paid Shengsuan Cloud API calls. Thanks again!

harry0703:
Thanks for the contribution! I incorporated and refined this in 5ceffd0 on main, with you credited as co-author. We kept the existing batch-script and looping behavior, improved quote retries, and added regression tests. The local suite passed with 1,088 tests, and live API checks plus portrait/landscape video generation passed too. Closing this PR since the changes are now integrated. Thanks again!

PR #1330, opened 2026-09-07
feat: added native support for pauses on the video script

### LINKED ISSUE #1320: [Feature]: Native support for pauses in the video script using Edge TTS
### 需求描述 | Problem Statement

[Feature Request] Support custom pause/silence tags in video scripts ([pause: Xs] / [pausa: Xs] / [停顿: X秒])

Currently, there is no way to insert a custom pause or silence of a specific duration (e.g., 2 or 3 seconds) between sentences directly in the video script. 

- **Natural punctuation** (`...`, `\n\n`, `.`) only provides short, fixed pauses (~0.3s to 0.7s) that cannot be customized.
- **SSML break tags** (like `<break time="2s" />`) do not work with Edge TTS (the default free engine) because `edge-tts` escapes XML characters, causing the synthesizer to either read the tag aloud phonetically (*"break time equals two seconds..."*) or fail.
- Creating pauses currently requires editing custom audio externally in third-party software before uploading.

---

Native multi-language pause tag support directly in the script input (e.g. `[pause: 2s]`, `[pausa: 2s]`, `[停顿: 2秒]`, `[silence: 1.5s]`, `[silencio: 500ms]`).

**How it should work:**
1. **Script Parsing:** Intercept and parse tags into speech blocks and pause durations.
2. **Audio Generation & Concatenation:** 
   - Synthesize speech blocks with the configured TTS engine.
   - Generate exact silence clips using FFmpeg (`anullsrc`).
   - Concatenate the clips with FFmpeg's `concat` filter to ensure consistent sample rates without audio glitches.
3. **Subtitle Synchronization:** Shift `SubMaker` cues (`start`/`end`) and legacy offsets by the accumulated duration of preceding pauses and audio chunks, ensuring subtitles remain 100% in sync without displaying technical pause tags on screen.
4. **Keyword Clean-up:** Automatically clean pause tags before sending text to the LLM for stock video keyword extraction (`generate_terms`) and subtitle sentence matching (`normalize_script_for_subtitle_matching`).

---

### Supported Syntax Examples

- **English:** `[pause: 2s]`, `[silence: 1s]`, `[pause: 2.5]`, `[pause: 500ms]`
- **Spanish:** `[pausa: 2s]`, `[silencio: 1.5s]`, `[pausa: 500ms]`
- **Chinese:** `[停顿: 2秒]`, `[暂停: 1.5s]`, `[静音: 500毫秒]`
- **Japanese / Korean / others:** `[一時停止: 2秒]`, `[ポーズ: 1.5s]`, `[일시중지: 2s]`, `[stille: 2s]`

---

### Additional Context

I have already implemented and locally tested this feature with unit tests (`test_voice.py`, `test_webui_i18n.py`) and real Edge TTS end-to-end integration tests. I can submit a Pull Request with the implementation if the maintainers are interested!

IsmaelCamC:
Resolves #1320
### Summary of Changes
1. **Scoped to Azure TTS v1 (Edge TTS) & Zero Regression:**
   - Normal scripts without pause tags bypass pause processing and call \`_single_tts\` directly with zero overhead and 100% existing behavior preserved.
   - Added test \`test_tts_without_pauses_calls_single_tts_directly\`.
2. **Documented Tag Set & Syntax Flexibility:**
   - Core keywords: \`pause\`, \`pausa\`, \`silence\`, \`silencio\`, \`停顿\`, \`暂停\`, \`静音\`, \`일시중지\`, \`ポーズ\`, etc.
   - Flexible syntax: supports standard \`[pause: 2s]\`, optional colon \`[pause 2s]\`, parenthesis \`(pause: 2s)\`, and optional unit defaulting to seconds (\`[pause 2]\`).
   - Supports seconds (\`s\`, \`sec\`, \`seconds\`, \`秒\`) and milliseconds (\`ms\`, \`毫秒\`).
3. **Duration Validation and Edge Cases:**
   - **Invalid durations:** Durations \`<= 0s\` (e.g. \`[pause: 0s]\`, \`[pause: -1s]\`) are safely ignored.
   - **Excessive durations:** Clamped at \`MAX_PAUSE_DURATION_SECONDS = 10.0\` with a warning log.
   - **Consecutive pauses:** Automatically merged into a single pause segment (e.g. \`[pause: 1s] [pause: 2s]\` -> \`3.0s\`), avoiding redundant silence files and FFmpeg overhead.
   - **Leading & trailing pauses:** Properly offsets speech audio and subtitle cues for leading pauses, and extends total audio length for trailing pauses without leaving dangling subtitle cues.
   - **Scripts with only pauses:** Safely generates silent audio and returns a clean SubMaker without speech cues.
   - **Subtitle sync:** Tested in both \`sentence\` mode and \`word_by_word\` mode.
4. **AI-Generated Scripts Clarification:**
   - Initial v1 support is specifically targeted at **manually entered or edited scripts** in the WebUI and API. Since \`format_response\` in \`llm.py\` strips bracketed stage directions by default, preserving LLM-generated pauses can be introduced in a subsequent PR if desired.
### Tests
All 71 unit tests in \`test/services/test_voice.py\` pass.


harry0703:
Thanks for the implementation and demo! I tested this locally, including real Edge TTS calls and FFmpeg concatenation. A few issues still need addressing before merging:

1. Subtitle offsets accumulate MP3 duration differences. With twelve 1-second audio clips and eleven 0.5-second pauses, the decoded output was 17.50 seconds, but the last subtitle started at 17.71 instead of 16.50 and ended at 18.71. Please calculate offsets from the actual decoded samples, ideally using PCM/WAV intermediates and encoding once at the end. A real multi-segment concatenation test would help catch this; the current mocks hide the difference.
2. Invalid tags are still passed to TTS: `[pause: -2s]` and `[pause: nope]` remain unchanged, while `[pause: 0s]` falls back to `_single_tts()` with the original text. Please reject or consistently remove invalid tags before synthesis. Also, a requested 1ms pause currently becomes 100ms, so please validate or document the minimum duration.
3. The new `tts()` wrapper checks only for pause tags, not the provider. Kokoro, Fish Audio, Gemini and SiliconFlow also enter the split-synthesis path. Please keep this first version limited to Azure TTS v1 (Edge TTS), as discussed, to avoid changing other providers’ request counts and behavior.

The full local suite passed: 1,097 tests, with 19 skipped. These issues came from additional boundary checks and real audio processing. Thanks again for working on this!

IsmaelCamC:
Thanks for the thorough review and benchmarking with real audio calls! All three points have been addressed and covered with new automated test cases.

---

### 1. Subtitle Offsets & Multi-Segment Drift (PCM/WAV Pipeline)
* **Root Cause**: Concatenating intermediate MP3 files accumulated encoder delay and bit-reservoir padding frames (~20–50 ms per clip) from `libmp3lame`. Over 12 speech clips and 11 pauses, this compounded into a +1.21s offset drift between subtitles and the decoded audio stream.
* **Changes**:
  - In `_tts_with_pauses()` and `_concat_audio_files()`, all intermediate speech chunks are decoded to standard **24,000 Hz 16-bit mono PCM WAV**.
  - Silence segments are generated directly as uncompressed PCM samples (`round(pause_duration * SAMPLE_RATE)`), eliminating framing padding.
  - Subtitle offsets are calculated strictly from the decoded PCM sample count: `cumulative_samples / float(SAMPLE_RATE)`.
  - MP3 encoding via FFmpeg is performed **only once at the very end** over the concatenated PCM master track.
* **Verification**:
  - Added `test_real_multi_segment_concatenation_no_drift` in `test/services/test_voice.py` without mocking concatenation or durations.
  - With 12 real 1.0s speech clips and 11 real 0.5s pauses, the decoded MP3 output is **17.50s**, and the 12th subtitle cue starts exactly at **16.50s** (ending at 17.50s) instead of drifting to 17.71s.

---

### 2. Invalid Tag Sanitization & Minimum Pause Duration
* **Invalid Tags (`[pause: -2s]`, `[pause: nope]`, `[pause: 0s]`)**:
  - Updated `PAUSE_TAG_PATTERN` to capture any tag starting with pause keywords regardless of arguments, including flexible syntax (e.g. `[pausa 1s]`, `(pausa: 2s)`).
  - `remove_pause_tags()` now cleans all pause tags—both valid and invalid—ensuring nothing leaks into TTS engines or subtitles.
  - `parse_script_with_pauses()` validates arguments; non-numeric values (`nope`) and non-positive durations (`<= 0s`) log a warning and are discarded from both pauses and speech text. If all tags are invalid, it safely falls back to `_single_tts()` with the cleaned text.
* **Minimum Duration (100 ms)**:
  - Formally defined and documented `MIN_PAUSE_DURATION_SECONDS = 0.1` (100 ms) and `MAX_PAUSE_DURATION_SECONDS = 10.0` in `app/utils/utils.py`.
  - Micro-pauses below 100 ms (such as `[pause: 1ms]`) log a warning and are clamped to the 100 ms threshold.

---

### 3. Provider Restriction (Limited to Azure TTS v1 / Edge TTS)
* Added `is_azure_v1_voice(voice_name)` in `app/services/voice.py`.
* Only Azure TTS v1 (Edge TTS) enters the split-synthesis path `_tts_with_pauses()`.
* Other providers (**Kokoro**, **Fish Audio**, **Gemini**, **SiliconFlow**, **ElevenLabs**, etc.) strip pause tags via `remove_pause_tags()` and route directly to `_single_tts()` in a single API call, preserving their default behavior, quotas, and request counts.

---

### Test Suite

<details>
<summary><b>75/75 tests passing in <code>test/services/test_voice.py</code></b> (click to expand)</summary>

```bash
$ python -m unittest test.services.test_voice
...........................................................................
----------------------------------------------------------------------
Ran 75 tests in 1.052s

OK (skipped=3)

harry0703:
Thanks for the updates! I retested the three previous issues: the PCM timing fix, invalid-tag cleanup, and provider restriction now work in the cases tested. Real Edge TTS synthesis also succeeded, and the full local suite passed with 1,101 tests and 19 skipped.

One blocker remains in `_tts_with_pauses()`: when a speech chunk cannot be decoded, the fallback around lines 895–905 generates silence from the subtitle duration and continues as if synthesis succeeded. With corrupted audio and valid subtitle cues, I reproduced a successful result containing two subtitle cues but three seconds of completely silent audio.

Please fail with a clear error when a speech chunk is empty or cannot be decoded, rather than replacing narration with silence. The tests should supply valid audio fixtures instead of requiring production fallbacks for mock files. Please add a regression test for this failure path as well.

Two smaller adjustments would also help: pass the original text through unchanged when there are no pause tags, and mention Azure TTS V1 (Edge TTS) plus the 0.1–10 second pause limits in the WebUI help text.

This is getting close—thanks again for working through the feedback!

IsmaelCamC:
Thanks for the thorough review and for catching that fallback behavior! All requested changes and adjustments have been implemented, tested, and pushed to `feature/pauses`:

### 1. Fail Fast on Empty or Undecodable Speech Audio
- **Removed Silence Fallback**: In `_tts_with_pauses()`, the fallback generating silent audio upon chunk decoding issues has been removed.
- **Strict Error Handling**: If a speech chunk audio file is missing, 0-byte, fails FFmpeg decoding to PCM WAV, or produces 0 frames, `_tts_with_pauses()` immediately logs an error and returns `None` rather than substituting narration with silence.
- **Regression Tests Added**:
  - `test_tts_with_pauses_fails_on_empty_chunk_audio`: ensures synthesis fails and returns `None` when a chunk file is empty.
  - `test_tts_with_pauses_fails_on_corrupted_chunk_audio`: ensures synthesis fails and returns `None` when audio data is corrupted.

### 2. Valid Audio Fixtures in Tests
- Replaced mock `Path(voice_file).touch()` empty files with `_write_test_wav()` fixtures generating valid 16-bit mono 24 kHz PCM WAV data across unit tests (`test_tts_with_pauses_shifts_submaker_timeline`, `test_pause_leading_and_trailing`, `test_pause_script_with_only_pauses`).
- Production code no longer needs to accommodate 0-byte mock files.

### 3. Original Text Passthrough Without Pauses
- In `tts()`, if `not utils.has_pause_tags(text)`, the script text is now passed through directly to `_single_tts()` completely untouched, preventing any unnecessary regex replacement, whitespace alterations, or bracket modification.
- Added regression test `test_tts_passes_original_text_unchanged_without_pauses`.

### 4. WebUI Help Text Updates
- Updated `"Video Script Help"` across all 12 localization files (`webui/i18n/*.json`) to explicitly document:
  - Supported provider: **Azure TTS V1 (Edge TTS)**
  - Allowed pause duration limits: **0.1–10 seconds**

---

### Test Suite Results
- Full local test suite (`pytest`): **1,104 passed, 19 skipped** (all green).
- Localization key consistency (`test_webui_i18n`): **13 passed, 0 failed**.
- Voice tests (`test_voice`): **78 passed, 3 skipped**.

harry0703:
Thanks for the contribution and for working through the feedback! Merged after local end-to-end testing, including real video generation through the CLI and API. We’ll follow up on the small empty-subtitle log issue for scripts containing only pauses.

PR #1332, opened 2026-09-08
Use full clip - Local file

unk0vn:
I needed a feature that would allow creating a short video using just a single clip (from start to finish). In the default v1.3.6, the clip changes at most every 10 seconds, either sequentially or randomly. I added this change, and when using a local file, the video now generates without restarting the same clip in a loop every 10 seconds—instead, we get one full clip from start to finish. I believe this is a useful feature and would love to see it included in the main release so I don't have to edit it manually in future updates.

Bez tytułu

final video(Compression):

https://github.com/user-attachments/assets/14059108-cc03-46e9-b76c-939e86afb54d

**Summary of Changes**

**1. `app/services/video.py` — `combine_videos()**`

* **Goal:** Allow the use of the entire downloaded footage instead of splitting it into fixed-length segments, controlled by the `"full"` value passed from the UI.
* Added flag: `use_full_clip = isinstance(max_clip_duration, str) and max_clip_duration.strip().lower() == "full"`.
* Loop building `subclipped_items`: if `use_full_clip` → adds the entire clip (`start_time=0`, `end_time=clip_duration`) without splitting. Otherwise, the original logic splitting into segments of `source_clip_duration` seconds applies (restored from the version prior to our first change).
* Second location trimming an individual clip: `if not use_full_clip and clip.duration > max_clip_duration: clip = clip.subclipped(0, max_clip_duration)` — trimming only applies when `"full"` is not selected.

**2. `webui/Main.py**`

* **Goal:** Add a "use full clip" option in the UI alongside existing numerical values.
* Options list (~line 4634): `[2, 3, 4, 5, 6, 7, 8, 9, 10]` → `[2, 3, 4, 5, 6, 7, 8, 9, 10, "full"]` (only for sources other than `metaso_minimax` — MiniMax retains its own time range).
* `format_func` in `stable_selectbox` for `video_clip_duration_select` — displays `"Full clip / Unlimited"` instead of `"full"`, and the rest as `"X sec"`.
* New helper function `_resolve_estimate_clip_duration(video_clip_duration)` — safely converts the value to an integer, with a fallback to `5` when the value is `"full"` or invalid.
* Replaced `clip_duration = max(int(params.video_clip_duration or 1), 1)` with `clip_duration = _resolve_estimate_clip_duration(params.video_clip_duration)` in four locations: `_render_wavespeed_video_settings`, `_render_seedance_video_settings`, `_render_ofox_video_settings`, and `_render_metaso_minimax_video_settings` — without this, selecting `"full"` for these sources would trigger a `ValueError` (as `int("full")` throws an exception).


harry0703:
Thanks for the contribution! Being able to use a local video continuously without splitting it into short segments is a useful option.

There are a few blockers in the current patch:

- `webui/Main.py` removes the existing Kokoro controls and parts of the Shengsuan Cloud workflow. These are unrelated to full-clip support and need to be preserved. The same targeted regression suite passes all 40 tests on main, but 36 fail on this branch.
- `video_clip_duration` still requires an integer in `VideoParams`, so `"full"` fails API/model validation and preset round-trip validation.
- The option is also available for cloud providers, but the helper only fixes the UI estimate. I tested the Seedance material entry point with `"full"`; it rejects it with `Seedance clip duration must be a positive integer` before generation.

Could you update the patch against the latest main and keep it focused on local videos first, with tests for parameter validation, presets, and full-clip playback? Please also keep the implementation summary in the PR description rather than above the README heading, and avoid the whole-file line-ending changes in `video.py`.

Happy to review the revised version. Thanks again!

unk0vn:
check now @harry0703

harry0703:
Thanks for the update! The full-clip validation and preset round-trip tests now pass, and limiting the UI option to local materials is a good improvement. Three issues still need fixing:

- The diff still removes Kokoro controls and parts of the Shengsuan Cloud workflow. The same 36 regression tests still fail; please restore those unrelated changes, including the current Shengsuan backend call arguments.
- Local materials can also be images. With a real PNG, numeric duration rendered successfully, but `"full"` failed with `could not convert string to float: 'full'`. Images need a numeric display duration, including when mixed with videos.
- The API model accepts `"full"` for cloud sources despite the UI restriction. `volcengine_seedance + full` passes validation, then fails in the material backend. Please reject unsupported source/duration combinations during validation.

Please also remove the whole-file line-ending changes to keep the diff focused. Thanks again for working on this!

PR #1333, opened 2026-09-08
fix(tts): handle non-finite voice rates

Sushanth012:
## Summary
- treat NaN and infinite voice rates as invalid input
- fall back to the normal +0% TTS rate instead of raising during rounding
- add regression coverage for both non-finite float values

## Why
Voice rate fields accept floats. NaN and infinity pass the existing non-positive guard, then round() raises ValueError or OverflowError and aborts voice generation. This keeps the existing invalid-input fallback behavior consistent for all unusable rates.

## Testing
- uv run python -X utf8 -m pytest -q test/services/test_voice.py
- 59 passed, 3 skipped, 2 subtests passed

### DIFF
--- a/app/services/voice.py
+++ b/app/services/voice.py
@@ -638,7 +638,7 @@ def convert_rate_to_percent(rate: float) -> str:
         rate = float(rate)
     except (TypeError, ValueError):
         rate = 1.0
-    if rate <= 0:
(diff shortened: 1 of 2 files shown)


harry0703:
Thanks for the focused fix and regression tests! I verified the non-finite inputs locally and confirmed normal rate conversion is unchanged. All CI checks passed. Merged!

PR #1339, opened 2026-09-09
feat(i18n): add Azerbaijani (az) WebUI translation

jamalkamaladdin:
This PR adds an Azerbaijani (az) translation.

New file webui/i18n/az.json: 409 strings, all filled in.
test/services/test_webui_i18n.py: added az to SECONDARY_LOCALES.


harry0703:
Thanks for adding Azerbaijani support! The translation checks passed, and we also checked the WebUI locally. Merged — thanks for your contribution!

PR #1340, opened 2026-09-10
[fix] Avoid repeated material selection across batch videos

giteePro:
## Problem

Generating multiple videos with `match_materials_to_script=true` passes the same downloaded material list to every sequential composition. Each output consumes the same prefix, so a batch can produce identical videos even when additional candidates have already been downloaded.

## Change

- Preserve keyword order while preferring less-used candidates within each keyword group for successive outputs.
- In random batch composition, prefer sources not yet used by earlier outputs while retaining the existing preference for distinct sources within each output.
- Track sources actually included in the composition, accounting for playback speed, short clips, processing failures, and safety-margin clips trimmed from the output.
- Persist per-output selections and display a translated warning when available candidates require reuse.

This applies to Pexels, Pixabay, Coverr, and local-material batches. Single-output behavior and AI-generation providers are unchanged. Reuse remains allowed when candidates are insufficient; the change does not guarantee unique outputs for an exhausted material pool or change semantic matching.

## Validation

- `python -m pytest test/services/test_batch_material_allocation.py test/services/test_video.py test/services/test_task.py test/services/test_webui_task.py -q --disable-warnings`: 141 passed, 3 skipped, 62 subtests passed (Python 3.11).
- `ruff check app cli.py main.py webui test`: passed.
- `python -m compileall -q app cli.py main.py webui test`: passed.
- `git diff --check`: passed.

Regression coverage includes ordered candidate rotation, random batches, material shortages, playback speeds, short-source looping, failed clips, trimmed safety margins, single-output compatibility, unaffected generated-video providers, and the result warning. Browser interaction and external provider integration tests were not run.

### DIFF
--- a/app/services/task.py
+++ b/app/services/task.py
@@ -1,3 +1,4 @@
+import json
 import math
 import os
 import re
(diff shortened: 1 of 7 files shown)


harry0703:
Thanks for the fix! Merged. We verified both ordered and random batches with real local video renders and filled in the remaining translations. The full local test suite passes.

PR #1341, opened 2026-09-10
docs: sync Edge TTS voice list with the built-in voice catalog

dajiaohuang:
`docs/voice-list.txt` is linked from the READMEs as the list of available Edge TTS voices, but it had not been regenerated since it was first added, so it drifted from `app/services/data/azure_voices.json` — the catalog the WebUI actually offers:

- **11 Edge voices were missing**, including the Inuktitut (`iu-Cans-CA-*`, `iu-Latn-CA-*`) voices and the `en-US` multilingual set.
- **3 listed voices no longer exist** and appear to have been superseded by their multilingual successors: `it-IT-GiuseppeNeural`, `ko-KR-HyunsuNeural`, `pt-BR-ThalitaNeural`.

Picking one of those three from the documented list fails at synthesis time, and the missing entries hide voices the app already supports.

This regenerates the file from that catalog, keeping its existing scope (Edge TTS only — Azure TTS V2 entries are still excluded), its `Name:`/`Gender:` layout and its sort order. 314 entries → 322.

Verification: re-derived the file from the catalog and compared byte-for-byte — all 314 pre-existing entries are reproduced unchanged, with no entries missing or stale afterwards. Documentation-only change; no source or test file reads this document.

### DIFF
--- a/docs/voice-list.txt
+++ b/docs/voice-list.txt
@@ -283,20 +283,32 @@ Gender: Female
 Name: en-US-AnaNeural
 Gender: Female
 
+Name: en-US-AndrewMultilingualNeural
+Gender: Male
+
 Name: en-US-AndrewNeural
 Gender: Male
 
 Name: en-US-AriaNeural
 Gender: Female
 
+Name: en-US-AvaMultilingualNeural
+Gender: Female
+
 Name: en-US-AvaNeural
 Gender: Female
+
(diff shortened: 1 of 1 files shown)


harry0703:
Thanks for updating the voice list! Verified it matches the built-in Edge TTS catalog. Merged.

PR #1343, opened 2026-09-12
feat: add VoxCPM TTS provider

lottshin:
## What does this PR do?

Adds VoxCPM TTS support through the ModelBest Audio Speech API.

## Changes

- Adds a VoxCPM provider to the existing voice provider framework
- Adds configurable API base URL, API key, model, voice, and timeout settings
- Adds WebUI configuration and provider selection
- Supports streamed SSE audio responses and WAV to MP3 conversion
- Adds configuration, localization, unit tests, and documentation

## Verification

- Ruff check passed
- Full test suite passed: 1114 passed, 19 skipped
- Manually verified a real voice preview through the ModelBest API

## Current scope

- Standard text to speech with the configured voice
- Reference-audio voice cloning is not included
- Numeric speed is not sent because the API does not expose a speed field


harry0703:
Thanks for adding VoxCPM support! The integration looks useful. I found a few things to address:

1. The API key can be cleared on reconnect: when Streamlit replays an empty password widget state, `voxcpm_api_key_input` overwrites the saved key. I reproduced this with AppTest. Please reuse the existing saved-credential handling and add a regression test before merging.
2. The speed selector remains enabled even though `voice_rate` is ignored. Disabling it for VoxCPM with a short explanation would make this clearer.
3. A 401 response triggers three identical requests. Please fail immediately for authentication/invalid-parameter errors and reserve retries, with a delay, for transient failures.

The focused tests passed (125 passed, 3 skipped), as did Ruff. I also checked real WAV-to-MP3 conversion and subtitle generation with a simulated SSE response; those passed. I haven't tested the live ModelBest API yet.

lottshin:
Thanks for the review. Addressed all three points in commit efc86c7.

- Preserved the saved VoxCPM API key when Streamlit replays an empty password widget state; added an AppTest regression.
- Disabled the VoxCPM speed selector and explained that numeric speed adjustment is not supported.
- Authentication and invalid-parameter responses (400, 401, 403, 404, 422) now fail immediately. Retries are reserved for transient HTTP failures and request exceptions, with backoff. Deterministic SSE/WAV conversion errors are not retried.

Validation:

- Focused tests: 36 passed
- Ruff: passed
- git diff --check: passed

harry0703:
Thanks for the contribution and for addressing the feedback! Tested Chinese and English speech with the live ModelBest API, plus a complete video render with subtitles. Merged.

PR #1356, opened 2026-09-13
refactor(video): centralize video source registry across entry points

shzshzshz123:
Each entry point used to maintain its own list of video_source values (WebUI VIDEO_SOURCE_GROUPS and the label map, CLI _CLI_VIDEO_SOURCES, the dispatch chain in material.download_videos and the batch allocation set in task.generate_final_videos). That duplication already caused real bugs: openai_image was once only available from a single entry point.

- Add app/models/video_sources.py as the single source of truth for source identity and stable capabilities (kind, UI group, CLI exposure), following the llm_provider registry convention.

- material.py derives its dispatch tables from that registry and now fails fast when a source is declared as generated but has no downloader, instead of silently degrading to stock search.

- task.py reads capabilities (batch allocation, local materials, search terms) from the registry instead of hardcoded string sets.

- cli.py derives the --video-source choices and webui/Main.py derives the group and label maps from the registry.

- Add test_video_sources_registry.py asserting the registry stays in sync with every consumer, so a new source that misses one entry point fails the suite.

Behavior is unchanged: same source set, same UI order, same CLI choices.


harry0703:
Thanks for the contribution! Centralizing the source registry makes sense. I tested bc28734 locally and found two issues to address before merging:

1. The image prompt changes are unrelated to the registry refactor. They now add cinematic styling by default, including when `openai_image_prompt_style` is empty, changing existing generation behavior. Could you remove those changes from this PR and keep the current prompt behavior?
2. `_STOCK_SEARCHERS` captures function references at import time, so existing tests that patch the Pexels/Coverr functions no longer intercept dispatch. Please update the mocks to patch the dispatch entries (or resolve the functions at call time), keeping these tests independent of credentials and real HTTP requests.

The targeted suite had 269 passed, 6 failed, and 3 skipped: three failures from the prompt changes and three from dispatch mocking. The same material tests all passed on main. Real HTTP requests were blocked during testing.

Happy to take another look once these are fixed. Thanks!

PR #1357, opened 2026-09-13
fix(cli): allow the zoom video transition modes

c020627:
## Problem

`--video-transition-mode` cannot select the zoom transitions, even though the WebUI and the render pipeline fully support them.

`app/models/schema.py:23` defines eight `VideoTransitionMode` members:

```
None, Shuffle, FadeIn, FadeOut, SlideIn, SlideOut, ZoomIn, ZoomOut
```

`webui/Main.py` offers all eight in the *Video Transition Mode* selectbox, and `app/services/video.py:875` dispatches all eight to `app/services/utils/video_effects.py`, which implements `zoomin_transition` and `zoomout_transition`.

`cli.py` however keeps its own hand-written table for the flag, and it stopped at the slide modes:

```python
_TRANSITION_MODE_VALUES = {none, shuffle, fade-in, fade-out, slide-in, slide-out}
```

The table was introduced in `92e411e` (2026-06-15, *"Add CLI flags for GUI-backed video params"*). `ZoomIn`/`ZoomOut` were added to the enum, the WebUI and `video_effects` in `e5dec5a` (2026-07-13, *"feat(video): add ZoomIn/ZoomOut transitions"*) — `cli.py` was not part of that change, and it has been edited ten more times since without picking the two values up, so this is a missed sync rather than a deliberate restriction of the CLI.

Result: the run aborts before it starts, and the error message enumerates the incomplete list:

```
error: argument --video-transition-mode: video-transition-mode must be one of:
none, shuffle, fade-in, fade-out, slide-in, slide-out
```

The config-file path is unaffected — it is documented with the enum spellings (`config.example.toml:619`, `# video_transition_mode = "None"     # "None", "Shuffle", "FadeIn", etc.`), so the CLI was the only entry point that could not reach `ZoomIn`/`ZoomOut`.

## Solution

Add `zoom-in` / `zoom-out` to the CLI table and to the flag `metavar`, following the kebab-case convention already used for `fade-in` / `slide-in` (`zoom-in` -> `ZoomIn`), which is how `_transition_mode` maps each spelling onto the enum.

## Changes

- `cli.py` — two new entries in `_TRANSITION_MODE_VALUES` (`cli.py:143-144`); the flag `metavar` now lists all eight spellings (`cli.py:358`).
- `test/services/test_cli.py` — one new test (`test/services/test_cli.py:54`) that derives the expected set from `VideoTransitionMode`, so a transition added to the enum and the pipeline can no longer be silently unreachable from the CLI, and then parses `--video-transition-mode zoom-in` end to end to `VideoTransitionMode.zoom_in`.

## Testing

- The new test fails on `main` with `AssertionError: Items in the second set but not the first: 'ZoomIn'` (and `'ZoomOut'`), and passes with the fix — a real regression test for the two missing entries.
- `test/services/test_cli.py`: `91 passed, 20 subtests passed`.
- Full suite: `1 failed, 1185 passed, 19 skipped, 9652 subtests passed`. The single failure is `test_webui_headless_task_actions.py::test_headless_open_folder_shows_host_mapped_path`, which also fails on a clean `main` in this environment (it patches `sys.platform` to `linux` while the local NumPy build reaches an `os.uname()` path that only exists on POSIX) and is unrelated to this change.
- `python -m compileall app cli.py main.py webui test` OK; `ruff check app cli.py main.py webui test` (0.15.21, the version pinned in `uv.lock`) -> `All checks passed!`.

## Notes for Reviewer

- `cli.py` is touched by three open PRs: #1206 (hunks at lines 253 and 695), #1322 (line 176) and #1356 (lines 13 and 35-52, around `_CLI_VIDEO_SOURCES`). This change is confined to lines 143-144 and 358, so no hunk overlaps with any of them in either merge order.
- `test/services/test_cli.py` is touched by #1315 (line 576) and #1322 (line 1489); the new test sits at line 54, far from both.
- Scope is a single defect: the CLI could not reach two transitions that are otherwise fully wired. No other entry point or transition mode was changed.


harry0703:
Thanks for catching this and adding the regression test! Verified both zoom modes locally, including batch inheritance and overrides. Merged.

PR #1358, opened 2026-09-14
fix(cli): allow the ElevenLabs background music mode

c020627:
## Problem

`--bgm-type` cannot select the ElevenLabs background music mode, so the CLI is the only entry point that cannot use it:

```console
$ python cli.py --video-subject "test" --bgm-type elevenlabs
cli.py: error: argument --bgm-type: bgm-type must be one of: none, random, custom, sonilo
$ echo $?
2
```

The run aborts during argument parsing, before any generation starts. The other entry points all accept the value:

| Entry point | ElevenLabs background music |
| --- | --- |
| Runtime dispatch (`app/services/task.py:82` `_VIDEO_MUSIC_PROVIDERS["elevenlabs"]`) | yes |
| WebUI selectbox (`webui/Main.py:6053`, `("ElevenLabs Background Music", "elevenlabs")`) | yes |
| API (`VideoParams.bgm_type` is an unconstrained `Optional[str]`) | yes |
| **CLI (`--bgm-type`)** | **no** |

## Root cause

`cli.py` keeps its own whitelist for `--bgm-type` instead of reading the runtime registry:

- `_bgm_type()` (`cli.py:171`) accepts `{"", "random", "custom", "sonilo"}`
- the same literal is repeated in `_validate_batch_task_params()` (`cli.py:1129`)
- the flag metavar (`cli.py:430`) lists the same four values

Sonilo was wired into the CLI by `281b5eb` (2026-07-14, "feat(bgm): add Sonilo video-matched music"), which touched `cli.py`. ElevenLabs landed four days later in `d5f31da` (2026-07-18, "feat(bgm): add ElevenLabs video-matched music") and updated `app/models/schema.py`, `app/services/task.py`, `app/services/elevenlabs_music.py`, `config.example.toml`, `webui/Main.py`, the i18n files and four test files — **but not `cli.py`**. The value was also re-typed by `b3599f6` (2026-08-22, "feat: add CLI batch manifest mode") when the batch validation was added, and `cli.py` has been edited twelve more times since, so the two lists simply drifted apart.

## Solution

Accept `elevenlabs` in the CLI whitelist, in the batch validation and in the flag metavar, so the three lists match the runtime registry again. `app/services/bgm.py:48 should_use_bgm()` is already provider independent (`should_use_bgm("elevenlabs", 0.2)` returns `True`) and `app/services/elevenlabs_music.py:363 generate_bgm()` already works with an empty prompt, so no runtime change is needed.

## Changes

- `cli.py:171` — accept `"elevenlabs"` in `_bgm_type()` and list it in the error message
- `cli.py:430` — extend the `--bgm-type` metavar and mention where ElevenLabs reads its key (`config.toml` or `ELEVENLABS_API_KEY`), mirroring the existing Sonilo sentence
- `cli.py:1129` — accept `"elevenlabs"` in `_validate_batch_task_params()` and list it in the error message
- `test/services/test_cli.py:559` — new `test_video_music_providers_are_selectable_from_the_cli`, which derives the expected providers from `app/services/task.py::_VIDEO_MUSIC_PROVIDERS` so the two lists cannot drift apart again

## Testing

Before/after on the same branch, using a small script that prints the runtime registry, the WebUI options and the CLI result:

```text
=== [1] runtime providers ===
    _VIDEO_MUSIC_PROVIDERS = ['elevenlabs', 'sonilo']
=== [3] should_use_bgm (provider independent) ===
    elevenlabs   -> True
    sonilo       -> True
=== [4] CLI: _bgm_type ===
  before: elevenlabs   -> ArgumentTypeError: bgm-type must be one of: none, random, custom, sonilo
  after:  elevenlabs   -> 'elevenlabs'
=== [5] CLI: parse_args --bgm-type elevenlabs ===
  before: SystemExit code = 2 (aborted before any generation)
  after:  accepted -> params.bgm_type='elevenlabs'
```

The new test fails on the unmodified `cli.py`:

```text
E       argparse.ArgumentTypeError: bgm-type must be one of: none, random, custom, sonilo
cli.py:173: ArgumentTypeError
```

- `test/services/test_cli.py`: `92 passed` (91 before)
- Full suite: `1 failed, 1186 passed, 19 skipped, 9652 subtests passed`. The single failure is `test/services/test_webui_headless_task_actions.py::test_headless_open_folder_shows_host_mapped_path`, a pre-existing win32 environment artifact that also fails on a clean `main`.
- `python -m compileall cli.py test/services/test_cli.py` and `ruff check app cli.py main.py webui test` (0.15.21, the version pinned in `uv.lock`) are clean.

## Notes for Reviewer

**Overlap with open PRs.** `cli.py` is also touched by #1206, #1322 and #1356, but no changed line is shared: #1206 rewrites lines 253 and 695-706, #1356 rewrites lines 13-22 and 35-52, and #1322 rewrites the parser description on line 181. This change touches lines 171-175, 430-436 and 1129-1133; the nearest neighbour (#1322) is six lines below the last line changed here, so any merge order is clean. In `test/services/test_cli.py` the new test sits at line 559 and the closest open hunk (#1315) starts at line 576.

**Scope boundary.** `--sonilo-bgm-prompt` lets CLI users pass a music-style prompt for Sonilo; there is no CLI flag for the provider-agnostic `video_music_prompt` field that the WebUI populates for ElevenLabs. That is a separate change (it introduces a new public flag and a naming decision), so it is deliberately left out here — this PR only makes the mode selectable, which is what the CLI currently rejects. ElevenLabs still generates video-matched music without a prompt, since `_request_bgm()` only sets `description` when the prompt is non-empty.


harry0703:
Thanks for fixing this and adding the regression test! Verified CLI parsing and batch handling locally, and CI is green. Merged.

PR #1359, opened 2026-09-14
fix(cli): allow the two-thirds subtitle position

c020627:
## Problem

`--subtitle-position` cannot select the "2/3 from bottom" position, so the CLI is the only entry point that cannot reach it:

```console
$ python cli.py --video-subject "test" --subtitle-position two_thirds_bottom
cli.py: error: argument --subtitle-position: invalid choice: 'two_thirds_bottom' (choose from top, center, bottom, custom)
$ echo $?
2
```

The run aborts during argument parsing, before any generation starts. Passing the position through `config.toml` is worse, because it fails silently — the saved value is thrown away and replaced with `bottom`:

```text
config.toml [ui].subtitle_position = two_thirds_bottom
cli.build_video_params().subtitle_position = bottom      # <- silently replaced
```

`webui/Main.py:7027` persists the dropdown selection into `config.toml` (`_set_runtime_config("ui", "subtitle_position", params.subtitle_position)`), so a user who picks "2/3 from Bottom" in the WebUI and then renders the same project from the CLI gets a different video, with no warning. The other entry points all support the value:

| Entry point | `two_thirds_bottom` |
| --- | --- |
| Runtime render (`app/services/video.py:1406`) | yes |
| Schema comment (`app/models/schema.py:142` — `# top, bottom, center, custom, two_thirds_bottom`) | yes |
| WebUI selectbox (`webui/Main.py:7005`, `(tr("2/3 from Bottom"), "two_thirds_bottom")`) | yes |
| API (`SubtitleRequest.subtitle_position`, unconstrained `Optional[str]`) | yes |
| **CLI (`--subtitle-position`, batch manifest)** | **no** |

## Root cause

`cli.py` keeps its own copy of the accepted subtitle positions in three places:

- `_subtitle_position()` (`cli.py:117`) — also the checker used when the value is read back from `config.toml` (`cli.py:842`)
- the `--subtitle-position` `choices` list (`cli.py:481`)
- `_validate_batch_task_params()` (`cli.py:1076`)

`50e9123` (2026-09-03, "feat(subtitle): add word-by-word display with pop-up spring animation") added the option — its message explicitly lists "Add 2/3 from bottom positioning option" — and updated `app/models/schema.py`, `app/services/subtitle.py`, `app/services/task.py`, `app/services/video.py`, `app/services/voice.py`, `webui/Main.py`, `webui/i18n/en.json` and `webui/i18n/zh.json`. **`cli.py` is not in that file list**, and it has been edited several times since (`bd34691`, `da4d3cb`, `e021e63`, `ec1fc34`) without picking the value up, so the lists drifted apart rather than being deliberately narrower.

This also breaks the stated contract of the config reader: `_ui_config_value()` (`cli.py:682`) documents that the checker functions "复用命令行使用的同一批类型函数 … 因此保存值和命令行参数遵循完全相同的取值规则". With a check that is narrower than what the WebUI can save, an unusable-looking value quietly degrades to the default instead of being reported.

## Solution

- Add `two_thirds_bottom` to the CLI through a single `_SUBTITLE_POSITION_VALUES` tuple that all three sites read, following the convention this file already uses for `_CLI_VIDEO_SOURCES` ("单任务 argparse 和批量清单必须共享同一来源集合 … 新增 Provider 不会再次遗漏批量校验"). Sharing one tuple also removes the drift itself, not just this instance of it.
- Update the documented value list in `config.example.toml`, which only mentioned four positions.

No runtime change is needed: `app/services/video.py:1406` already renders the value.

## Changes

- `cli.py:151` — new `_SUBTITLE_POSITION_VALUES` tuple, ordered like the WebUI dropdown
- `cli.py:117` — `_subtitle_position()` validates against the tuple and lists the same values in its error message; because this function is also the `config.toml` checker, a saved `two_thirds_bottom` is now applied instead of being discarded
- `cli.py:491` — `--subtitle-position` uses `choices=_SUBTITLE_POSITION_VALUES`
- `cli.py:1088` — batch manifest validation uses the same tuple
- `config.example.toml:640` — document `two_thirds_bottom`
- `test/services/test_cli.py:1908` — new `test_two_thirds_bottom_subtitle_position_is_supported`, covering both the explicit flag and the saved-config path
- `test/services/test_cli.py:1935` — new `test_subtitle_positions_stay_aligned_with_the_webui`, which extracts the dropdown values from `webui/Main.py` and asserts the CLI accepts every one of them, so a future position cannot be added to the WebUI alone

## Testing

Before/after on the same branch, using a small script that reads the saved config value the same way the CLI does:

```text
=== [after] ===
config.toml [ui].subtitle_position = two_thirds_bottom
cli.build_video_params().subtitle_position = two_thirds_bottom
=== [before] ===
config.toml [ui].subtitle_position = two_thirds_bottom
cli.build_video_params().subtitle_position = bottom
```

Both new tests fail on the unmodified `cli.py`:

```text
FAILED test/services/test_cli.py::TestCliUiDefaults::test_subtitle_positions_stay_aligned_with_the_webui
FAILED test/services/test_cli.py::TestCliUiDefaults::test_two_thirds_bottom_subtitle_position_is_supported
__main__.py: error: argument --subtitle-position: invalid choice: 'two_thirds_bottom' (choose from top, center, bottom, custom)
```

- `test/services/test_cli.py`: `94 passed` (92 before)
- Full suite: `1 failed, 1188 passed, 19 skipped, 9652 subtests passed`. The single failure is `test/services/test_webui_headless_task_actions.py::test_headless_open_folder_shows_host_mapped_path`, a pre-existing win32 environment artifact that also fails on a clean `main`.
- `python -m compileall cli.py test/services/test_cli.py` and `ruff check app cli.py main.py webui test` (0.15.21, the version pinned in `uv.lock`) are clean.

## Notes for Reviewer

**Overlap with open PRs.** `cli.py` is also touched by #1206, #1322 and #1356, but no changed line is shared: #1206 rewrites lines 253-271 and 695-706, #1356 rewrites lines 13-22 and 35-52, and #1322 rewrites the parser description on line 181. This change touches lines 117-123, 151-157, 491 and 1085-1091, so any merge order is clean. In `test/services/test_cli.py` the new tests sit at lines 1908-1956; the closest open hunks are #1315 at line 576 and #1322 at line 1489. `config.example.toml:640` is ~150 lines away from the nearest open hunk (`#1315`, lines 464-488).

**Scope boundary.** `app/services/video.py:1406` also tolerates the undocumented aliases `two_thirds` and `2/3_bottom`. This PR only accepts the canonical `two_thirds_bottom` that the WebUI writes, because nothing in the project emits the aliases and adding them would let the CLI accept spellings that the WebUI cannot save. `subtitle_display_mode` and `subtitle_animation` need no CLI change: `cli.py` never overrides them, so they keep flowing through the `VideoParams` defaults validated by `schema.py::_get_valid_ui_choice`.


harry0703:
Thanks for catching the saved-config fallback too! Verified the CLI option, saved settings, and batch handling locally. CI is green, and this is merged.

PR #1360, opened 2026-09-14
fix(cli): let --video-music-prompt reach every AI music provider

c020627:
## Problem

The CLI exposes a music style prompt only through `--sonilo-bgm-prompt`, which is hard-wired to Sonilo. A user choosing `--bgm-type elevenlabs` therefore has **no way to pass a prompt at all**.

The runtime reads the provider-agnostic field `params.video_music_prompt` (`_get_video_music_prompt()` in `app/services/task.py`), and the WebUI writes that same field for **both** AI music providers (`webui/Main.py`, sonilo and elevenlabs branches). The CLI never set it, so ElevenLabs generated background music from an empty prompt — the style the user typed was silently dropped.

## Solution

- Add `--video-music-prompt`, accepted for any AI music provider.
- Reject it unless `--bgm-type` is `sonilo` or `elevenlabs`, since only those branches consume it. The value is *not* used to infer the provider (unlike `--sonilo-bgm-prompt`), because inferring could pick the wrong one.
- Hoist the provider list into `_CLI_MUSIC_BGM_TYPES`, kept in sync with `_VIDEO_MUSIC_PROVIDERS` in `app/services/task.py`, and derive the `--bgm-type` metavar, the error text and the batch validation from it. This removes the duplication that allowed the gap in the first place.

Batch manifests keep working: `video_music_prompt` is now part of the per-task optional field list and is validated the same way.

## Testing

- `pytest test` → `1190 passed, 19 skipped`. The single failure (`test_headless_open_folder_shows_host_mapped_path`) is pre-existing and unrelated — it calls `os.uname`, which does not exist on Windows.
- `ruff check` clean, `compileall` clean.
- New regression tests:
  - `test_video_music_prompt_reaches_every_ai_music_provider` — derives the expected providers from `task_service._VIDEO_MUSIC_PROVIDERS` and asserts the prompt reaches each of them.
  - `test_video_music_prompt_requires_an_ai_music_provider` — `--bgm-type random` / `none` with a prompt exits with code 2 and a clear message, while the same `--bgm-type random` without a prompt still parses.
- Both new tests fail on the pre-fix code (`unrecognized arguments: --video-music-prompt`), confirming they actually guard the regression.


harry0703:
Thanks for adding the shared music prompt and keeping the old Sonilo flag compatible! Verified both providers, prompt precedence, and batch handling locally. CI is green. Merged.

PR #1361, opened 2026-09-14
fix(cli): allow the video clip speed from the command line

c020627:
## Problem

`video_clip_speed` can be set from every entry point **except** the single-task CLI:

| Entry point | Can set it |
| --- | --- |
| WebUI slider (`webui/Main.py:5186`, range 0.5–2.0) | yes — and it persists the value to `[ui].video_clip_speed` (`webui/Main.py:5195`) |
| Batch manifest (`cli.py:1190` allows every `VideoParams` field, range-checked at `cli.py:1157`) | yes |
| Runtime (`app/services/task.py:917` → `app/services/video.py:852 with_speed_scaled`) | consumes it |
| **Single-task CLI** | **no** |

`cli.py` has no `--video-clip-speed` flag, and unlike most persisted fields it
never picks up the saved value either: `VideoParams.video_clip_speed` is
hard-coded to `1.0` (`app/models/schema.py:114`), whereas its neighbours read
`config.ui` — `subtitle_position` (`schema.py:140`) and `custom_position`
(`schema.py:149`) via `config.ui.get(...)`, `subtitle_display_mode` /
`subtitle_animation` (`schema.py:143`, `:146`) via `_get_valid_ui_choice(...)`.

So a user who sets 1.5x in the WebUI and then runs the same `config.toml` through
`cli.py` silently gets 1.0x, with no warning.

Root cause: the feature was added without touching `cli.py`. `396b045`
("feat(video): global clip speed multiplier from the UI", 2026-07-13) changed
`schema.py`, `task.py`, `video.py`, `utils.py`, `webui/Main.py`, the i18n files
and the tests — **`cli.py` is not in that file list**. `cli.py` has been touched
repeatedly since (`eb8c237`, `eb9c237`, `ba49f34`, `44ce430`, `1727d7a`, `bd34691`,
…), and the batch path even grew a dedicated range check for this very field
(`50091bd`), but the single-task path never did.

## Solution

- Add `--video-clip-speed`, validated by a new `_clip_speed` checker.
- When the flag is absent, read `[ui].video_clip_speed`, so the WebUI saved value
  reaches the CLI exactly like `subtitle_position` already does.
- Hoist the 0.5–2.0 range into `_CLIP_SPEED_MIN` / `_CLIP_SPEED_MAX` and reuse it
  from the batch validation too, so the two paths cannot drift apart again.
- Add `video_clip_speed` to `optional_arg_names`; batch manifests keep working
  unchanged.

## Testing

- `pytest test/services/test_cli.py` → **98 passed** (was 96).
- New regression tests:
  - `test_video_clip_speed_is_reachable_from_the_cli` — asserts **both** paths:
    the explicit flag, and the value saved in `[ui]`.
  - `test_video_clip_speed_range_matches_the_runtime_normalizer` — asserts
    `cli._CLIP_SPEED_MIN` / `_CLIP_SPEED_MAX` equal `app/utils/utils.py`'s
    (the constants `normalize_clip_speed()` uses), and that `0.4`, `3.0`, `nan`
    and `inf` all exit with code 2.
- Both new tests fail on the pre-fix code
  (`error: unrecognized arguments: --video-clip-speed`).
- `ruff check cli.py test/services/test_cli.py` → All checks passed;
  `compileall` clean; `cli.py --help` now lists the flag.
- Full suite result: `1 failed, 1192 passed, 19 skipped, 9652 subtests passed in 394.24s`.
  The single failure (`test_webui_headless_task_actions.py::test_headless_open_folder_shows_host_mapped_path`)
  is pre-existing and unrelated — it calls `os.uname`, which does not exist on Windows.

## Notes for Reviewer

- Scope is deliberately limited to the one field that is **completely
  unreachable** from the single-task CLI. Other persisted UI values that the CLI
  already exposes as flags (e.g. `video_clip_duration`) are left alone —
  whether the CLI should also inherit those saved values is a broader design
  question, not this bug.
- No new dependency, no public signature change.


harry0703:
Thanks for the fix and regression tests! Merged after CI passed and local end-to-end checks with seven generated videos confirmed clip speeds, saved settings, and batch overrides work as expected, with audio unchanged.

PR #1313, opened 2026-09-04
feat(subtitle): add word-by-word display with pop-up spring animation

farhanahmad15:
This pull request adds advanced subtitle customization features to the video generation workflow, including new display modes, animation options, and positioning for subtitles. It introduces support for word-by-word subtitles, a pop-up spring animation effect, and a new "2/3 from bottom" subtitle position. The changes span backend logic, UI controls, and internationalization.

**Subtitle Display and Animation Enhancements**
- Added support for a "word by word" subtitle display mode, allowing subtitles to appear one word at a time, and a "sentence" mode for traditional display. This includes backend logic to generate word-level subtitles for both Whisper and Edge providers (`app/services/subtitle.py`, `app/services/voice.py`, `app/services/task.py`, `app/models/schema.py`). [[1]](diffhunk://#diff-1aa60fc713ea39063b95097acfcc1c4809f91103a4f05540a11213d24237e767L22-R22) [[2]](diffhunk://#diff-1aa60fc713ea39063b95097acfcc1c4809f91103a4f05540a11213d24237e767R84-R90) [[3]](diffhunk://#diff-b913bcda0a38a5f8839b3395669199064a8deb177f7338510e0ec3dbaa19fb6dR599-R606) [[4]](diffhunk://#diff-b913bcda0a38a5f8839b3395669199064a8deb177f7338510e0ec3dbaa19fb6dL615-R625) [[5]](diffhunk://#diff-a1f6f81383dcee945232117bda16726d20077e4ca3138e601a8ffb63ef8f57c6L2113-R2188) [[6]](diffhunk://#diff-9ca641fc0072695ae41a94c4f94ad1dc329148be558102357bf38ac13365fc57L124-R130) [[7]](diffhunk://#diff-9ca641fc0072695ae41a94c4f94ad1dc329148be558102357bf38ac13365fc57R156-R159)
[...]

PR #1362, opened 2026-09-14
feat(cli): expose the subtitle display mode and animation

c020627:
## Problem

`subtitle_display_mode` ("sentence" vs "word_by_word") and `subtitle_animation` ("none" vs "pop_spring") were added by #1313 together with the word-by-word renderer. That PR touched `app/models/schema.py`, `app/services/{task,subtitle,video,voice}.py`, `webui/Main.py` and the i18n files — **not `cli.py`**.

As a result the single-task CLI is the only entry point that cannot set them:

| Entry point | `subtitle_display_mode` / `subtitle_animation` |
| --- | --- |
| WebUI selectboxes (`webui/Main.py:7032`, `:7062`) persist them to `[ui]` | ✅ |
| Batch manifest (`cli.py`, field whitelist = `set(VideoParams.model_fields)`) | ✅ |
| `config.toml` `[ui]` (`_get_valid_ui_choice()` in `schema.py:143`, `:146`) | ✅ |
| Runtime consumes them (`task.py:600`, `video.py:1399`) | ✅ |
| **Single-task CLI** | ❌ no flag for either |

The CLI already exposes **ten** subtitle options (`--subtitle-position`, `--custom-position`, `--font-name`, `--font-size`, `--text-fore-color`, `--stroke-color`, `--stroke-width`, `--subtitle-enabled`, `--subtitle-background-enabled`, `--subtitle-background-color`, `--rounded-subtitle-background`) — these two were simply missed, even though `subtitle_position` is the closest sibling and does have one.

There is also a second, subtler half: the CLI reads every other persisted subtitle setting explicitly, because `VideoParams` resolves `config.ui` **once at import time** (`cli.py:888-892` documents exactly this for `subtitle_position` / `custom_position` / `video_clip_speed`). These two fields were not in that list either, so the command line depended on that import-time freeze instead of re-reading `[ui]`.

## Solution

- Add `--subtitle-display-mode` and `--subtitle-animation`, taking `choices` from two new shared tuples `_SUBTITLE_DISPLAY_MODE_VALUES` / `_SUBTITLE_ANIMATION_VALUES`, kept in sync with `_SUBTITLE_DISPLAY_MODES` / `_SUBTITLE_ANIMATIONS` in `app/models/schema.py` (a test enforces this, so a future mode cannot silently miss the CLI again).
- Register both in `optional_arg_names`, so an explicit flag reaches `VideoParams`.
- Add both to `ui_defaults`, with checkers mirroring `_subtitle_position`, so `python cli.py` alone resolves flag → saved `[ui]` value → built-in default exactly like the neighbouring subtitle fields, rather than relying on the import-time freeze.

Batch manifests were already validated for these fields by Pydantic (`SubtitleDisplayMode` / `SubtitleAnimation` are `Literal`s, and `cli.py:1237` uses `VideoParams.model_validate`), so no extra batch validation is needed — unlike `subtitle_position`, which is typed `Optional[str]` and therefore validated by hand in `_validate_batch_task_params`.

## Testing

- `pytest test/services/test_cli.py` → **100 passed, 22 subtests passed** (98 before).
- Full suite result: `1 failed, 1194 passed, 19 skipped, 9654 subtests passed in 405.14s`. The single failure (`test_headless_open_folder_shows_host_mapped_path`) is pre-existing and unrelated — it calls `os.uname`, which does not exist on Windows.
- `ruff check cli.py test/services/test_cli.py` clean, `compileall` clean.
- `python cli.py --help` now lists `[--subtitle-display-mode {sentence,word_by_word}]` and `[--subtitle-animation {none,pop_spring}]`.
- New regression tests:
  - `test_subtitle_style_options_follow_the_flag_then_the_saved_value` — an explicit flag sets the field, and a saved `[ui]` value is applied when the flag is absent.
  - `test_subtitle_style_values_stay_aligned_with_the_model_and_the_webui` — asserts the CLI tuples equal `app_schema._SUBTITLE_DISPLAY_MODES` / `_SUBTITLE_ANIMATIONS`, and that the WebUI selectbox options (read with `ast`, as `test_subtitle_positions_stay_aligned_with_the_webui` already does) are a subset of them.
- Both new tests fail on the pre-change code (`unrecognized arguments: --subtitle-display-mode word_by_word`), confirming they guard the regression.

2 files, no new dependencies, no behaviour change for the WebUI or the API.


harry0703:
Thanks for adding these CLI options and the regression tests! Merged after CI and local checks passed, including subtitle combinations, saved settings, and invalid inputs.

PR #1331, opened 2026-09-07
feat: AI media integrations, avatar overlay, cover gen, OmniVoice Studio, and MP3 export

3djimenez:
## Summary

This PR adds several major features and hardening fixes to MoneyPrinterTurbo.

### New capabilities

- **AI media integrations** (`app/integrations/`): unified provider router with fallback chain across fal, getimg, D-ID, HeyGen, Stability AI, Replicate, Shotstack, and BannerBear; SSRF-hardened HTTP helpers; per-provider API key resolution
- **Avatar/character pipeline**: `character_overlay.py` + `character_lipsync.py` for compositing avatar images with lipsync over generated video; bridge in `task.py` wires position/offset UI controls into the pipeline
- **Cover generation**: `cover.py` generates AI thumbnails via the provider router
- **Narration sources**: `narration_source.py` resolves script text from prompt, uploaded PDF, or custom text
- **Voice preview**: `voice_preview.py` isolated preview endpoint so users can audition a voice before full generation
- **Background task helpers**: `bg_tasks.py` for task lifecycle management
- **External webhooks**: `app/controllers/external.py` handles async callbacks from HeyGen, D-ID, etc.
- **New WebUI pages**: Text-to-Voice (page 2), OmniVoice Studio (page 3 — voice profile manager with edge_tts)
- **MP3 export**: `POST /export` accepts `output_format` field to convert WAV→MP3/FLAC via ffmpeg; new `GET /export/audio-as` endpoint streams converted audio for browser downloads; MP3 button added next to the WAV download icon on every history row

### Fixes & hardening

- Suppress MoviePy video probe output
- Timeout hanging edge-tts streams
- Harden external request security (SSRF, TLS)
- LiteLLM empty-message guard; LiteLLM added as LLM provider
- Harden task folder opening on Windows
- `start.sh`: single-command launcher for WebUI + TTS service
- `.gitignore`: exclude generated mp4/mp3/wav outputs, runtime logs, and local tooling dirs

## Test plan

- [ ] `python -m pytest test/` — all tests pass
- [ ] `start.sh` starts TTS service (port 8090) and WebUI (port 8080) cleanly
- [ ] Video generation pipeline runs end-to-end (local materials + Pexels)
- [ ] OmniVoice Studio page loads; voice preview works
- [ ] Avatar overlay renders at correct position when `avatar_image_file` is set
- [ ] Cover thumbnail generated when enabled
- [ ] MP3 export: clicking the MP3 button on a history row downloads a valid MP3
[...]

PR #1297, opened 2026-09-02
feat(webui): add evidence-grounded trend discovery

raahul09:
## Summary
- add evidence-grounded trend discovery for YouTube Shorts, TikTok, and Instagram Reels across US, IN, GB, CA, and AU
- rank opportunities with transparent predictive scoring, filtering, shortlisting, and Studio handoff
- add bounded source retries, snapshot validation, stale-cache fallback, localized UI labels, and focused tests

## Verification
- full repository test suite: passed
- focused trend/WebUI suite: 127 passed, 1 skipped
- locale contract: 12 passed, 5,954 subtests passed
- compileall, scoped Ruff, and git diff checks: passed
- live Google Trends RSS and real Streamlit browser acceptance recorded in docs/testing/2026-09-01-trend-discovery-acceptance.md

## Limits
- TikTok and Instagram Reels rankings are explicitly inferred from public Google evidence, not native platform measurements
- key-backed YouTube acceptance remains credential-gated
- retention potential is predictive, not measured viewer retention

PR #1363, opened 2026-09-14
docs(config): document the ElevenLabs music base URL

c020627:
## Problem

`config.example.toml` documents `music_model_id` and `music_timeout` for the
ElevenLabs Video-to-Music integration, but not `music_base_url` — even though the
code reads it (`app/services/elevenlabs_music.py:57`):

```python
config.elevenlabs.get("music_base_url", DEFAULT_BASE_URL)
```

Unlike the API key — which falls back to `ELEVENLABS_API_KEY` inside
`get_api_key()` — the endpoint has **no environment-variable fallback**, so the
example file is the only place a user could discover the setting. Anyone who
needs to route Video-to-Music through a proxy or a self-hosted gateway (an
internal network, or a region where `api.elevenlabs.io` is slow) currently has no
documented way to do it.

Three signals point at this being an oversight rather than a deliberate omission:

1. **Sibling parity inside the section**: `music_model_id` and `music_timeout`
   are both listed; `music_base_url` is the only `music_*` key that is not.
2. **Cross-provider parity**: every other provider section that supports a custom
   endpoint lists it — `[chatterbox]`, `[kokoro]`, `[voxcpm]` and `[minimax_tts]`
   all have `base_url`.
3. **Timeline**: `d5f31da` ("feat(bgm): add ElevenLabs video-matched music")
   introduced the reader *and* edited `config.example.toml`, adding
   `music_model_id` / `music_timeout` while missing `music_base_url`.

## Solution

- Add `music_base_url = "https://api.elevenlabs.io"` to `[elevenlabs]`, with a
  comment noting that it only affects the Video-to-Music endpoint, so the voice
  API keeps its default host.
- Extend `test_example_config_documents_runtime_settings` to assert the whole
  `[elevenlabs]` section. That test already pins the `[voxcpm]` section verbatim,
  which is exactly why this gap was never caught: only the newest provider was
  covered, and the section-specific assertions are hand-written.

The documented value equals `DEFAULT_BASE_URL` in `app/services/elevenlabs_music.py`,
so behaviour is unchanged until a user actually overrides it.

## Testing

- `pytest test/services/test_config.py` → `19 passed`.
- The new assertion **fails on the unpatched example file** with
  `Right contains 1 more item: {'music_base_url': 'https://api.elevenlabs.io'}`,
  so it genuinely guards the regression.
- Full suite: `1 failed, 1194 passed, 19 skipped, 9654 subtests passed`. The single
  failure is pre-existing and unrelated — `test_headless_open_folder_shows_host_mapped_path`
  calls `os.uname`, which does not exist on Windows.
- `ruff check` and `compileall` clean; the example file still parses with
  `tomllib` and the documented value equals the code default.

## Notes for reviewer

`config.example.toml` is also touched by open PRs #1206, #1255, #1297, #1315,
#1322 and #1331. Their hunks sit between lines 1 and 464 of that file, while this
change is at line 536, so no merge order can conflict.


harry0703:
Thanks for documenting the missing setting and adding the regression check! Local tests and CI passed. Merged.

PR #1364, opened 2026-09-14
docs(config): document [ui] subtitle/voice defaults in example config

c020627:
## Problem

`config.example.toml` ships a fully commented `[ui]` block. It is the only place a user can pin initial defaults by hand — the WebUI persists these keys on write, but the model schema and the CLI read them back as the starting value of each control. The block documents 35 keys, yet exactly the 3 keys the code reads from `[ui]` for the subtitle/voice controls are missing:

| missing key | read from `[ui]` by | fallback default |
| --- | --- | --- |
| `voice_mode` | `cli.py` (`_ui_config_value`), `app/models/schema.py` (`config.ui.get`) | `"tts"` |
| `subtitle_display_mode` | `app/models/schema.py` (`_get_valid_ui_choice`) | `"sentence"` |
| `subtitle_animation` | `app/models/schema.py` (`_get_valid_ui_choice`) | `"none"` |

Every *other* subtitle sibling is already listed (`subtitle_enabled`, `font_name`, `font_size`, `text_fore_color`, `subtitle_position`, `custom_position`, `subtitle_background_*`), so the omission reads as an oversight rather than a deliberate exclusion. The gap dates back to the word-by-word subtitle feature (`a79e047`), which touched the schema, WebUI, exporter and i18n but not the example config — leaving users who copy the sample block with no way to discover these three knobs.

## Solution

Add the 3 commented lines to the `[ui]` sample block, following the existing inline style (value + allowed values), and add a regression test that derives the expected set from the source instead of hard-coding it — so a future `[ui]`-readable field fails the test until it is documented.

```toml
# voice_mode = "tts"                 # "tts", "upload", or "none"
# subtitle_display_mode = "sentence" # "sentence" or "word_by_word"
# subtitle_animation = "none"        # "none" or "pop_spring"
```

The test `test_example_ui_block_documents_every_pinnable_default` collects the "pinnable" key set by scanning `app/models/schema.py` for `config.ui.get("…")` and `_get_valid_ui_choice("…")`, and `cli.py` for `_ui_config_value(…, "…")`, then asserts every one of them appears in the `[ui]` comment block.

## Testing

- New test fails **before** the fix with `AssertionError: ['subtitle_animation', 'subtitle_display_mode', 'voice_mode']`, passes after.
- `test/services/test_config.py`: 20 passed (19 before).
- Full suite: 1 failed, 1195 passed, 19 skipped, 9654 subtests passed. The single failure is `test_headless_open_folder_shows_host_mapped_path`, a pre-existing Windows environment artifact (`AttributeError: module 'os' has no attribute 'uname'` inside numpy's `hugepage_setup`), unrelated to this change and reproducible on a clean checkout.
- `ruff check` clean; `python -m compileall` clean; `tomllib` parses the edited file.

## Notes for reviewer

Comments + test only — no runtime behavior change. Two other open PRs touch `config.example.toml`, but their hunks sit at ≤ line 488 while this change is at lines 628–640, so there is no overlap.


harry0703:
Thanks for filling in the missing defaults and adding the regression test! Local checks and CI passed. Merged.

PR #1365, opened 2026-09-14
fix: decode subprocess output as UTF-8 instead of the system locale

britorj:
## Problem

`subprocess.run(..., text=True)` is used at nine call sites without an explicit `encoding`. Python then decodes with `locale.getpreferredencoding()`, which on non-English Windows is an ANSI codepage (cp1252 here), not UTF-8. Every helper process this project spawns emits UTF-8, so non-ASCII output was silently mangled.

The `claude_code` LLM provider is the worst case, because its output *becomes the video script*. The CLI returns UTF-8 JSON, so a Portuguese script came back double-encoded — `ã` was stored as the bytes `\xc3\x83\xc2\xa3`, i.e. the characters `Ã£`.

Two failures followed, neither of which pointed back at the encoding:

1. **The narration read the mojibake out loud.** Edge TTS pronounced the stray `£` and `©` as words. I only found the root cause because Whisper transcribed the audio as `"em An Libra O"` where the script said `não`.

2. **Subtitles vanished silently.** The mojibake broke Edge cue alignment, which logged `edge cues still have unmatched text after aggregation` / `failed, sub_items len: 12, script_lines len: 16`, returned `""`, and the video rendered with no subtitles. The code deliberately does not fall back to Whisper here (to avoid an unexpected multi-GB download), so the only symptom a user sees is a subtitle-less video.

Reproducing needs only a non-English Windows locale and any non-ASCII script language.

## Fix

Pass `encoding="utf-8", errors="replace"` at every `subprocess.run` call site — the claude CLI plus the ffmpeg/ffprobe helpers.

`errors="replace"` is deliberate: it keeps a helper process that emits unexpected bytes from raising `UnicodeDecodeError` in the middle of a render.

The ffmpeg/ffprobe calls parse mostly ASCII output today, so they are not currently broken. They are fixed anyway because their output embeds file paths, and an accented path would hit exactly the same bug.

## Tests

Four test doubles shadow `subprocess.run` with a fixed `(command, capture_output, text, check)` signature and raised `TypeError` on the new keywords. They now accept `**kwargs`.

Full suite on Windows: **1193 passed, 19 skipped, 9654 subtests passed**, same as before the change. The two remaining failures are pre-existing and unrelated to this PR — both are Windows-only issues in the tests themselves:

- `test_webui_headless_task_actions.py` patches `sys.platform` to `"linux"`, and numpy (first imported inside that patch) takes its Linux branch and calls `os.uname()`, which does not exist on Windows;
- `test_webui_loomloom_regressions.py:375` uses `Path.read_text()` with no `encoding=`, so it hits the very same locale-decoding problem this PR is about, on the i18n JSON files.

I left that second one alone to keep this diff focused, but it is the same bug class and is a one-line fix if you would like it included.

## Verification

After the fix, on the same machine: the script is stored as single-encoded UTF-8 (`\xc3\xa3`), Edge subtitle alignment succeeds, and a full 9:16 render completes with correctly accented burned-in subtitles.


harry0703:
Thanks for tracking down the encoding issue and verifying the fix on Windows! Local tests passed, and our subprocess checks confirmed that non-ASCII text stays intact. Merged.

PR #1366, opened 2026-09-15
fix(skill): decode piped uv output as UTF-8

c020627:
## Problem

`run_checked()` in `docs/skill/mpt_agent.py` pipes `uv sync`'s combined
stdout/stderr back into Python so it can print the last 30 lines when the
dependency sync fails:

| call site | stdout target | decodes with |
| --- | --- | --- |
| `run_checked` (`docs/skill/mpt_agent.py:582`) | `subprocess.PIPE` | host locale (`text=True`, no `encoding`) — **bug** |
| `generate_video` (`docs/skill/mpt_agent.py:652`) | a log file | nothing (`result.stdout` is `None`) — fine |

With `text=True` and no explicit `encoding`, Python decodes the pipe with the
host locale (`locale.getpreferredencoding(False)` — `cp936` on a Chinese
Windows box, `cp1252` on a Western one, and exactly what the `windows-latest`
CI job runs under). `uv` emits UTF-8, so any non-ASCII dependency error turns
into mojibake before it reaches the failure tail, e.g.
`依赖安装失败：网络不可达` arrives as `渚濊禆瀹夎�呭け璐ワ細缃戠粶涓嶅彲杈�`.

PR #1365 already fixed this exact defect class for every other
PIPE-decoding `subprocess.run` call (`app/services/llm.py`,
`app/services/video.py`, `app/services/voice.py`,
`app/services/elevenlabs_music.py`, `app/services/sonilo.py`) — `run_checked`
is the call site that slipped through.

## Solution

Add `encoding="utf-8"` to the `run_checked` call (`errors="replace"` was
already there), matching the exact pattern #1365 used everywhere else.

The sibling call in `generate_video` is deliberately **not** changed: its
`stdout` is redirected to a log file, so `result.stdout` is `None` and
`text=True` never decodes anything — adding `encoding` there would be a no-op.

## Testing

- Runtime reproduction under a non-UTF-8 locale (`PYTHONUTF8=0`, preferred
  encoding resolves to `cp936`), piping UTF-8 Chinese out of a child process
  exactly like `run_checked` does:

  - before fix: `repr: '渚濊禆瀹夎�呭け璐ワ細缃戠粶涓嶅彲杈�\n'` (mojibake)
  - after fix: `repr: '依赖安装失败：网络不可达\n'` (correct)

- New regression test
  `test_run_checked_decodes_piped_output_as_utf8` mocks `subprocess.run` and
  asserts the call carries `stdout=subprocess.PIPE`, `text=True` and
  `encoding="utf-8"`. It fails **before** the fix with:

  ```
  AssertionError: None != 'utf-8' : piped output must be decoded as UTF-8, not the host locale
  ```

  and passes after.

- `test/services/test_mpt_agent_skill.py`: 29 passed (28 before).
- Full suite: `1 failed, 1196 passed, 19 skipped, 9654 subtests passed`. The
  single failure is `test_headless_open_folder_shows_host_mapped_path`, the
  pre-existing Windows environment artifact (`os.uname` inside numpy's
  `hugepage_setup`), reproducible on a clean checkout and unrelated here.
- `ruff check` (0.15.21) clean; `compileall` clean.

## Notes for reviewer

- Scope is one kwarg on one call site plus one test — no behavior change on
  UTF-8 hosts, and non-UTF-8 hosts go from mojibake to readable output.
- Why the second `subprocess.run` in this file is untouched: `generate_video`
  redirects `stdout` to `.agent-logs/.../run-<task>.log`; the log is later
  re-read with an explicit `encoding="utf-8"` (`log_path.read_text(...)`), so
  the UTF-8 guarantee already holds on that path. Pinning `encoding` on a
  file-redirected call would be dead code.
- No open PR touches `docs/skill/mpt_agent.py` (checked all 18 open PRs).


harry0703:
Thanks for catching this missed call site and adding the regression test! Verified on macOS, Windows 11, and Windows 10. CI passed, and this is now merged.

PR #1367, opened 2026-09-15
feat: add MuAPI video material source

Anil-matcha:
## Summary

- Add MuAPI as a configurable asynchronous text-to-video material source.
- Use the budget `seedance-lite-t2v` route by default, while allowing the endpoint, base URL, resolution, duration bounds, polling interval, and timeout to be configured.
- Wire the source through the material pipeline, task recovery, CLI, WebUI, sample configuration, and video-generation skill.
- Require explicit charge confirmation and never retry an ambiguous submission, since the remote task may already have been accepted and billed.
- Preserve the remote request ID when polling or downloading fails so a user can recover a completed task.

MuAPI's documented submit-then-poll contract is implemented here: `POST /api/v1/{endpoint}` returns a request ID, followed by `GET /api/v1/predictions/{request_id}/result`. See the [MuAPI API reference](https://muapi.ai/docs/api-reference).

## Validation

- `uv run ruff check ...`
- `uv run pytest -q`
- Full suite: 1227 passed, 19 skipped.
- All MuAPI tests use mocked HTTP responses; no live or billable MuAPI request was made.


harry0703:
Thanks for adding MuAPI support and the tests! The full suite passed locally (1,227 passed, 19 skipped). Two things to address before merging:

1. The material loop counts the requested duration instead of the downloaded video’s actual duration. With a real 2-second MP4 returned for a 5-second request, it stopped after one clip despite needing 5 seconds. Please measure the downloaded file and count its usable duration.
2. Submission and polling requests follow redirects by default. A local 307 test forwarded the dummy API key to another origin and replayed the POST. Please disable automatic redirects for both requests and handle unexpected redirects without resubmitting the paid job.

These were local reproductions, not observations of MuAPI’s live service. We haven’t tested live paid generation yet. Could you also add regression coverage for these cases?

Anil-matcha:
Thanks for the concrete reproductions. Addressed in `1d7ebf0`:

- MuAPI submission and polling requests now use `allow_redirects=False`. Unexpected 3xx responses are treated as unconfirmed task state, with no retry or paid resubmission.
- The MuAPI material loop now probes the downloaded media file and accumulates `min(requested_clip_duration, actual_downloaded_duration)`. A shorter returned file therefore causes the next script term to be generated when more coverage is needed.
- Added mocked regression coverage for POST redirects, polling redirects, and a 2-second download returned for a 5-second request.

Validation:

- `uv run pytest` — 1230 passed, 19 skipped
- `uv run ruff check app/services/muapi.py app/services/material.py test/services/test_muapi.py`

No live or billable MuAPI request was made.

harry0703:
Thanks for adding MuAPI support and addressing both issues! The local test suite and our real-file/HTTP regression checks passed. Merged.

PR #1368, opened 2026-09-15
fix(cli): allow the WaveSpeed video source from the command line

c020627:
## Problem

The WebUI treats WaveSpeed as a first-class video source. It is a selectable entry in the source dropdown (`webui/Main.py:123`), it has a dedicated settings panel (`Main.py:5243`), a per-request charge checkbox (`Main.py:5270-5283`) and its own validation branch (`Main.py:7378-7387`). `config.example.toml:65` lists `"wavespeed"` as a valid value and documents the billing model at `:68`, and `app/services/material.py:1696` already dispatches to `_download_videos_wavespeed_on_demand()`.

The CLI rejected it outright. `_CLI_VIDEO_SOURCES` (in `cli.py`) never learned about the source, so argparse failed before any of the existing validation could run:

```
$ python cli.py --video-subject test --video-source wavespeed
usage: cli.py [-h] ...
cli.py: error: argument --video-source: invalid choice: 'wavespeed'
       (choose from 'pexels', 'pixabay', 'coverr', 'volcengine_seedance',
       'ofox', 'metaso_minimax', 'openai_image', 'local')
```

Every other per-request billed generator that the CLI does expose (`volcengine_seedance`, `ofox`, `metaso_minimax`) is registered there, has a `_CLI_VIDEO_SOURCES` entry, and is guarded by an explicit `--confirm-<source>-charge` flag. So this was an oversight rather than an intentional restriction.

Note that `loomloom` is intentionally kept out of the CLI, as the comment above `_CLI_VIDEO_SOURCES` explains: it drives an interactive WebUI quote (`LoomLoomConfirmedVideoRequest`) and carries substantial `st.session_state` coupling. WaveSpeed is different — it is configured by a plain `wavespeed_api_keys` list and needs no interactive session state, which is exactly the shape the CLI already supports for the three siblings.

## Solution

Register the source and mirror the sibling contract exactly:

1. Add `"wavespeed"` to `_CLI_VIDEO_SOURCES`, next to `coverr`.
2. Add `--confirm-wavespeed-charge`, required once the pipeline reaches the `materials` or `video` stage.
3. Enforce it in **both** paths the other billed generators use:
   - the single-task path, via `parser.error(...)`, placed directly before the existing `volcengine_seedance` branch;
   - the batch manifest path, via `_validate_batch_task_params(..., wavespeed_charge_confirmed=...)`, placed directly before the existing `volcengine_seedance` check.

Change: `cli.py` +29 lines, `test/services/test_cli.py` +68 lines — 2 files, 97 insertions, no deletions, no new dependencies, no refactoring.

## Testing

New tests in `test/services/test_cli.py`:

- `test_wavespeed_video_source_requires_explicit_charge_confirmation` — without the flag, `SystemExit` with code 2; with it, `build_video_params(args).video_source == "wavespeed"`.
- `test_wavespeed_confirmation_is_not_required_before_material_stage` — `--stop-at script` needs no flag.
- `test_batch_wavespeed_source_uses_global_charge_confirmation` — batch manifest rejects without the flag (`run_cli` returns 2, `task.start` not called) and succeeds with it (returns 0, `start.assert_called_once()`).

Confirmation that the tests fail **before** the fix and pass **after**:

```
# before (with only the new tests applied)
$ pytest test/services/test_cli.py -k wavespeed
3 failed, 100 deselected
# failure cause:
#   SystemExit: 2 ... error: argument --video-source: invalid choice: 'wavespeed'

# after
$ pytest test/services/test_cli.py -k wavespeed
3 passed

$ pytest test/services/test_cli.py
103 passed, 22 subtests passed

$ pytest test/services/test_cli.py test/services/test_config.py test/services/test_material.py
165 passed, 27 subtests passed

$ ruff check cli.py test/services/test_cli.py
All checks passed!

$ python -m compileall cli.py test/services/test_cli.py
OK
```

No change was needed in `app/services/task.py`: unlike `volcengine_seedance`, WaveSpeed has no preflight branch there (`grep -n wavespeed app/services/task.py` returns nothing), so this PR deliberately does not touch it.

## Notes for reviewer

- The placement of each new block is chosen to sit immediately before its `volcengine_seedance` counterpart, so a future refactor that unifies the charge-confirmation path (see the direction taken in #1356) can treat all four sources mechanically.
- This PR is purely additive: no existing behaviour changes, and no source that was previously accepted is now rejected.
- If you would prefer the flag to be named differently (e.g. `--confirm-charge` with a per-source value), I am happy to adjust.


harry0703:
Thanks for fixing this CLI omission and covering both single and batch tasks! Local regression tests and CI passed. Merged.

PR #1369, opened 2026-09-15
docs(readme): list the Fluxion AI provider in the provider lists

c020627:
## Problem

The provider lists in `README.md`, `README-en.md` and `README-ja.md` are complete enumerations of the gateways and local runtimes the app can talk to, which is why `API Route` and `Claude Code` were added to them when those providers landed. `Fluxion AI` is registered in `app/models/llm_provider.py` and is selectable in the WebUI, but it is missing from all three lists.

The commit that introduced the provider (`b66674b feat(llm): add Fluxion AI provider`) did touch all three READMEs, but only to add a sponsor entry. The enumeration itself was left at 26 of the 27 registry entries.

## Solution

Add `Fluxion AI` between `API Route` and `Ollama`, which is exactly the registry order asserted by `test/services/test_llm.py::test_provider_registry_preserves_product_group_order`:

```python
... "openrouter", "api_route", "fluxionai", "ollama", "claude_code", ...
```

The link is the registry's `api_key_url`, the same convention the neighbouring entries follow (Shengsuan Cloud and APIMart also carry their referral URLs there).

## Testing

Enumeration coverage was checked mechanically: all 27 provider ids of `LLM_PROVIDER_REGISTRY` were probed against the `- [x]` feature lines of each README.

```
before: README-en.md 26/27 (missing fluxionai)
        README-ja.md 26/27 (missing fluxionai)
        README.md    25/27 (missing qwen, fluxionai)
after:  README-en.md 27/27
        README-ja.md 27/27
        README.md    26/27 (only the qwen probe string remains)
```

The remaining `qwen` hit on `README.md` is a probe artefact, not a gap: the Chinese list names that provider `阿里云通义千问`, so it has no ASCII `qwen` token.

No Python or test file is touched. `grep -rn "README" --include=*.py test/` returns nothing, and `README*.md` appears in `docker-ghcr.yml` under `paths-ignore` only, so CI behaviour is unchanged.

## Notes for reviewer

- The other surfaces were checked as well: `config.example.toml` (`fluxionai_api_key`), `webui/i18n/en.json`, `webui/i18n/zh.json` and the provider registry all mention Fluxion AI. The README lists were the only place it was missing.
- The change is one line per file with no reflow; line endings are untouched (all three files are CRLF and stay CRLF).
- Scope is a single documentation fix, nothing is bundled with it.


harry0703:
Thanks for catching the omission and updating all three READMEs! CI passed, and this is now merged.

PR #1371, opened 2026-09-15
fix(skill): accept the WaveSpeed video source in the agent helper

c020627:
## Problem

`docs/skill/mpt_agent.py` keeps its own copy of the video-material source list and rejects anything
outside it. The comment directly above that copy (line 36) states the rule it is supposed to follow:

> Keep this list aligned with ``_CLI_VIDEO_SOURCES`` in cli.py. A source that the CLI accepts must
> not be rejected here as unsupported.

Once WaveSpeed was allowed on the command line (`cli.py:40-50`), the copy was left one entry short:

| Source list | Contents |
| --- | --- |
| `cli.py::_CLI_VIDEO_SOURCES` | 9 — `pexels`, `pixabay`, `coverr`, **`wavespeed`**, `volcengine_seedance`, `ofox`, `metaso_minimax`, `openai_image`, `local` |
| `mpt_agent.py::SUPPORTED_SOURCES` | 8 — the same list without `wavespeed` |

`wavespeed` is the only source the CLI accepts and the helper still rejects, so a run through the
helper dies inside `missing_config` before any of the designed exit-code-10 flow happens:

```
$ python mpt_agent.py ... --video-source wavespeed
SkillError: unsupported video source: wavespeed
```

The paid-generation guard is missing for it as well. `cli.py:721-730` (`parser.error`) and
`cli.py:1191-1200` (`build_video_params`) both refuse to start a WaveSpeed material stage without
`--confirm-wavespeed-charge`, but `missing_config` has no WaveSpeed branch. Even with the key
configured, the helper reported nothing missing and then let the CLI hard-fail on the same run,
instead of emitting `WAVESPEED_CHARGE_CONFIRMATION_REQUIRED` so the agent can ask the user for
explicit consent first. The three other paid generation sources — `volcengine_seedance`, `ofox` and
`metaso_minimax` — all have that branch.

## Solution

- Add `wavespeed` to `SUPPORTED_SOURCES` (in the position `_CLI_VIDEO_SOURCES` uses).
- Mirror the `seedance` branch in `missing_config`: require `wavespeed_api_keys`, and require
  `--confirm-wavespeed-charge`. The runtime reads the key from that config key only
  (`app/services/material.py:728` → `get_api_key`, which consults `config.app` and nothing else), so
  this branch deliberately keeps the same credential priority instead of inventing an environment
  fallback the runtime does not have.
- Report `WAVESPEED_CHARGE_CONFIRMATION_REQUIRED=--confirm-wavespeed-charge` from
  `report_missing_config`.
- Document the flag in `docs/skill/SKILL.md` next to the other three paid sources.

Every edit is an addition — 0 lines removed, no behavior change for any other source.

## Testing

Before / after with the same `config.toml`:

```
未配置 + 未确认 -> ('moonshot', ['wavespeed_api_keys', 'confirm_wavespeed_charge'])
已配置 + 未确认 -> ('moonshot', ['confirm_wavespeed_charge'])
已配置 + 已确认 -> ('moonshot', [])
```

`report_missing_config` now prints the flag, and still does not print the LLM provider option list —
the missing field is `wavespeed_api_keys`, which does not match the `_api_key` suffix that gates that
block:

```
MPT_NEEDS_INPUT
LLM_PROVIDER=moonshot
MISSING=wavespeed_api_keys
MISSING=confirm_wavespeed_charge
WAVESPEED_CHARGE_CONFIRMATION_REQUIRED=--confirm-wavespeed-charge
```

The new guard test derives its expectation from `cli.py` instead of hardcoding it, and it fails on the
previous revision:

```
cli.py _CLI_VIDEO_SOURCES = [coverr, local, metaso_minimax, ofox, openai_image, pexels, pixabay,
                            volcengine_seedance, wavespeed] (9 项)

修复前: 8 项 | set(SUPPORTED_SOURCES) == _CLI_VIDEO_SOURCES -> FAIL   差集: ['wavespeed']
修复后: 9 项 | set(SUPPORTED_SOURCES) == _CLI_VIDEO_SOURCES -> PASS
```

- `python -m pytest test/services/test_mpt_agent_skill.py -q` → **32 passed** (was 29; +3 new)
- `python -m compileall -q docs/skill/mpt_agent.py test/services/test_mpt_agent_skill.py` → OK
- `ruff check` on both touched Python files: identical rule set to the unmodified baseline. This
  machine has ruff 0.16.7 while CI runs the pinned version, so the signal is baseline-vs-patch rather
  than a zero count.

## Notes for reviewer

- `docs/` is not in the CI lint target (`ruff check app cli.py main.py webui test`), so
  `docs/skill/mpt_agent.py` was checked by hand with the commands above.
- No `MPT_WAVESPEED_API_KEY` environment override was added. WaveSpeed is the only source here whose
  runtime credential comes from config alone, and the rule stated in the neighbouring comments is
  that the helper keeps the same credential priority as the runtime.
- `openai_image` likewise has no `os.environ` fallback in `missing_config`; that is left unchanged.
- The open MuAPI PR #1367 touches these same three files. A local three-way merge of the two heads
  auto-merges all three with no conflicts, so either merge order is fine.


harry0703:
Thanks for aligning the helper with the CLI and adding the source-list regression test! Local tests and CI passed. Merged.

PR #1372, opened 2026-09-15
fix(i18n): stop the subtitle animation from reusing the transition "None" label

c020627:
## Problem

Two independent dropdowns in `webui/Main.py` rendered the same translation key:

- **video transition mode** — `webui/Main.py:5061`: `(tr("None"), VideoTransitionMode.none.value)`
- **subtitle entrance animation** — `webui/Main.py:7061`: `(tr("None"), "none")`

`webui/i18n/zh.json` defined that key **twice**:

```
:98    "None": "无转场",     # transition group
:198   "None": "无动画",     # animation group, added by the subtitle animation feature
```

JSON keeps the last definition, so the Chinese **video transition** dropdown displayed 无动画 ("no animation"), and `无转场` was unreachable. `en.json` carried the same duplicate pair (invisible there, both values are `"None"`).

The leak also went the other way: the locales that translate `"None"` with transition wording showed that wording inside the **animation** dropdown — `de` `Kein Übergang`, `it` `Nessuna transizione`, `pt` `Nenhuma Transição`, `ru` `Без перехода`, `vi` `Không Có Chuyển Đổi`.

## Solution

Give the animation option its own key so the two dropdowns stop sharing one string:

- `webui/Main.py:7061` → `tr("None (Animation)")`
- `webui/i18n/en.json:198` / `webui/i18n/zh.json:198` → `"None (Animation)"`; the duplicated `"None"` entry is gone, so the transition dropdown shows 无转场 again
- `ENGLISH_FALLBACK_KEYS` in `test/services/test_webui_i18n.py` gains `"None (Animation)"` — only `en`/`zh` maintain the new label, and the repo's i18n guard requires a new key to either exist in every secondary locale or be declared as an English fallback
- new test `test_locale_files_do_not_redefine_a_translation_key` parses every locale with `object_pairs_hook` and fails when one file defines a key twice, so this class of silent override cannot come back

## Testing

```
pytest -q test/services/test_webui_i18n.py      -> 14 passed
```

With the fix reverted the new test fails as intended (`duplicates == ['None']`), so it is a real regression guard rather than a tautology.

Full suite (`pytest -q test`): no new failures. The only failures are pre-existing environment ones and reproduce identically on `origin/main` without this change — `test_does_not_serve_symlink_to_file_outside_tasks` (needs symlink privileges on Windows), `test_stats_only_include_managed_top_level_regular_files` (counts leftover files in `storage/`), `test_headless_open_folder_shows_host_mapped_path` (Windows path specific).

## Notes for reviewer

- 4 files, `+31 / -3`, no new dependency, no behaviour change for English (both labels still read "None").
- The whitelist line is required by the repository's own i18n guard test, not an optional addition.
- `docs/skill/*` and the READMEs are deliberately untouched, so this does not collide with the other open PRs.


harry0703:
Thanks for fixing the shared translation key and adding a regression test for duplicate locale keys! The local i18n tests and Streamlit checks passed, and CI is green. Merged.

PR #1373, opened 2026-09-15
fix(api): accept every local material format the CLI already supports

c020627:
## Problem

Local material formats are declared in three places and they no longer agree:

| Source | Video extensions | Image extensions |
| --- | --- | --- |
| `app/models/const.py:34-35` (canonical) | `mp4`, `mov`, `mkv`, **`webm`** | `jpg`, `jpeg`, `png`, **`bmp`** |
| `cli.py:1459-1464` | the four above + `avi`, `flv` | the four above |
| `app/services/material_upload.py:21-22` | `mp4`, `mov`, `avi`, `flv`, `mkv` | `jpg`, `jpeg`, `png` |

`const.py` predates the others and is the only list the render pipeline reads:
`app/services/video.py:1592,1622` opens a local material as an image when its extension
is in `const.FILE_TYPE_IMAGES`. So `.webm` and `.bmp` materials are already handled end
to end, but the API layer drops them:

- `POST /video_materials` -> `save_material_upload()` returns
  `unsupported local material format; supported formats: mp4, mov, avi, flv, mkv, jpg, jpeg, png`
  for a `.webm` or `.bmp` upload that `cli.py` happily accepts from a manifest.
- `GET /video_materials` reuses `SUPPORTED_MATERIAL_EXTENSIONS` to glob `local_videos/`
  (`app/controllers/v1/video.py:392-399`), so a `.webm` or `.bmp` file that is already
  sitting in the managed directory is silently invisible to API clients.

`app/services/bgm.py:25-27` already documents the rule these lists are meant to follow —
a single source of truth so the upload control and the server cannot disagree. This PR
closes the same gap for local materials.

## Change

`app/services/material_upload.py`

- `.webm` joins `SUPPORTED_VIDEO_EXTENSIONS` and `.bmp` joins `SUPPORTED_IMAGE_EXTENSIONS`,
  matching `const.FILE_TYPE_VIDEOS` / `const.FILE_TYPE_IMAGES`.
- `.bmp` gains `".bmp": frozenset({"BMP"})` in `_IMAGE_FORMATS_BY_EXTENSION` — without it
  `_validate_image` would raise `KeyError` instead of returning `MaterialUploadError`.
- the now-stale `valid JPEG or PNG` error message names BMP.

## Tests

`test/services/test_material_upload.py`

- `test_webm_and_bmp_materials_are_accepted` — a `.webm` upload and a real BMP upload are
  persisted; a PNG renamed to `.bmp` is still rejected by the content/extension check.
- `test_supported_extensions_cover_cli_accepted_formats` — fails if
  `SUPPORTED_MATERIAL_EXTENSIONS` stops covering the formats `const.py` declares, and keeps
  `SUPPORTED_IMAGE_EXTENSIONS` in sync with `_IMAGE_FORMATS_BY_EXTENSION`.

Both new tests fail on `main` with `unsupported local material format ...` and pass with
this change. `pytest -q test` is green, `ruff check` and `ruff format --check` are clean.

## Scope

`webui/Main.py:194-203` (`LOCAL_MATERIAL_EXTENSIONS`) has the same gap on the Streamlit
side. It is intentionally left untouched here to keep this PR on the API layer.


harry0703:
Thanks for aligning the API material formats with the CLI and adding regression coverage! Local tests passed, including real WebM/BMP uploads, listing, preprocessing, and invalid-file rejection. Merged — appreciate the focused fix!

PR #1374, opened 2026-09-15
fix(bgm): enforce Sonilo resource limits like the ElevenLabs provider

c020627:
## Problem

`app/services/sonilo.py` and `app/services/elevenlabs_music.py` implement the same
video-to-music feature. PR #1131 (`3a96206`, *fix(bgm): enforce ElevenLabs resource
limits*) hardened one of them — and only one of them:

```
$ git show --stat 3a96206
 app/services/elevenlabs_music.py       | 59 ++++++++++++++++---------
 test/services/test_elevenlabs_music.py | 81 ++++++++++++++++++++++++++++++++++
```

Sonilo was merged ten days before that fix (`281b5eb`, 2026-07-14) and has been edited
since (`a5b4a39`, 2026-09-14, the UTF-8 decoding fix) without picking any of it up.
`git log -S '"-fs"' -- app/services/` returns exactly one commit, and it is #1131. Three
of the limits that commit added are therefore still missing on the Sonilo path:

| #1131 hardening | `elevenlabs_music.py` | `sonilo.py` (on `main`) |
| --- | --- | --- |
| bounded error body | `:87` reads one `MAX_ERROR_BODY_BYTES` chunk through `iter_content()` | `:78` called `response.text`, truncating to 500 characters only *after* the whole body had been materialised |
| streamed response closed reliably | `:120` `with requests.get(..., stream=True)` | `:92` plain `requests.get()` — the error body was downloaded eagerly, before any code could decide to bound it |
| generated proxy bounded during encoding | `:231` passes `-fs str(MAX_PROXY_BYTES)` | no `-fs`; the 300 MB cap was applied only post hoc at `:216` |

The third row is the one with a concrete cost. `-fs` makes FFmpeg stop at the cap, so
ElevenLabs fails fast on an oversized proxy. `sonilo.py` encoded the entire proxy first
and rejected it afterwards, so the work was already paid for — up to the 600 s subprocess
timeout and `MAX_PROXY_BYTES` of temporary disk — before the same outcome was reached.

The first two rows only bite on error paths, but there they bite hard: the music requests
are issued with `stream=True`, so `response.text` forces a full read of a body that may be
a large HTML error page. Both limits already exist in the module
(`MAX_PROXY_BYTES`, `MAX_GENERATED_AUDIO_BYTES`), so this is a missing implementation,
not a missing decision.

## Change

`app/services/sonilo.py`, mirroring #1131 one-for-one:

- adds `MAX_ERROR_BODY_BYTES = 500` and rewrites `_safe_response_error()` to pull a single
  bounded chunk via `iter_content()`. It no longer touches `response.text` at all, so the
  `[:500]` truncation is no longer a cap on what is *kept* but on what is *read*.
- `test_connection()` now uses `with requests.get(..., stream=True) as response:` and does
  the status check inside the block, matching the ElevenLabs implementation — the error
  body is only read when it is actually needed, and the connection is released
  deterministically instead of relying on garbage collection.
- `_create_video_proxy()` passes `-fs str(MAX_PROXY_BYTES)` to FFmpeg, which turns the
  existing `proxy_size > MAX_PROXY_BYTES` check at `:216` into a backstop rather than the
  only enforcement.

## Tests

`test/services/test_sonilo.py`

- new `test_safe_response_error_reads_only_one_bounded_chunk` — the stand-in response
  raises from a `.text` property and asserts the body is not read past the first chunk, so
  the test fails if the unbounded read comes back.
- `test_connection_uses_non_billing_services_endpoint` now asserts the request is issued
  with `stream=True` and that the response context manager actually ran.
- `test_create_video_proxy_uses_expected_ffmpeg_policy` now asserts the `-fs` argument
  carries `MAX_PROXY_BYTES`, the same assertion #1131 added for ElevenLabs.
- the shared `_StreamingResponse` fixture gains `encoding`, `closed` and `iter_content`,
  mirroring the fixture update in #1131.

These three tests fail on `main` and pass with this change:

```
$ pytest -q test/services/test_sonilo.py          # main
3 failed, 17 passed, 28 subtests passed

$ pytest -q test/services/test_sonilo.py          # this branch
20 passed, 28 subtests passed
```

`ruff check` is clean on both files. `ruff format --check` reports both files, but it
reports them on `main` too (as it does `elevenlabs_music.py`) — the repository does not
run `ruff format --check` in CI, and no new formatting complaint is introduced.

A full local `pytest -q test` on Windows also reports five failures that are unrelated to
this change and reproduce on `main`: two are environment artifacts (the sandbox turns
`symlink_to` into a regular file, and one test asserts a forward-slash `storage/` path),
and three Streamlit `AppTest` cases flake under load — those three pass in isolation both
with and without this change.

## Scope

Only the Sonilo path is touched. The ElevenLabs side already has all three limits, and
Sonilo's own parameter caps (`MAX_VIDEO_DURATION_SECONDS`, `MAX_PROMPT_LENGTH`,
`MAX_GENERATED_AUDIO_BYTES`) were already enforced, so nothing there changes.

This is a follow-up to #1131 rather than a new finding: the same fix was merged for the
sibling provider, and this PR closes the gap it left behind.


harry0703:
Thanks for bringing the Sonilo resource limits in line with ElevenLabs and adding focused regression coverage! The 51 related local tests passed, and real FFmpeg checks confirmed normal proxy generation and cleanup when the size limit is exceeded.

There is one small error-handling regression to address before merging. In `_safe_response_error()`, decoding with `response.encoding` can raise `LookupError` if an upstream error response declares an unknown charset. I reproduced this with a local HTTP server returning status 500 and `Content-Type: text/plain; charset=unknown-charset`: the new implementation raises `LookupError`, whereas the previous `response.text` implementation falls back. Since the generation path expects `SoniloError`, this can bypass the intended no-BGM fallback.

Could you catch `LookupError` around decoding and fall back to UTF-8 with replacement, while keeping the single bounded read? Please also add a regression test using an unknown charset to confirm the error remains a `SoniloError` at the caller and the response is closed. This looks like a rare edge case, but the fix should be small. Thanks again for the improvement!

c020627:
Fixed — thanks for the exact repro, the local-server trace pointed straight at it.

`_safe_response_error()` now wraps the decode in `try/except LookupError` and falls back to `body_bytes.decode("utf-8", errors="replace")`. That is what the old `response.text` did internally: `requests` catches `LookupError`/`TypeError` there and retries the decode without the declared charset, which is why replacing it with an explicit `.decode(response.encoding)` turned a tolerated edge case into a raised exception. `errors="replace"` was never the gap — it only covers `UnicodeDecodeError`, not an unknown encoding name. The single bounded `iter_content()` read is unchanged, so the resource side of the PR is untouched.

Regression test: `test_request_bgm_reports_unknown_charset_body_as_sonilo_error` starts a loopback `ThreadingHTTPServer` that answers `500` with `Content-Type: text/plain; charset=unknown-charset`, and drives the generation path `_request_bgm()` through a `trust_env = False` session (the same loopback-only pattern `test_youtube_audience.py` uses, so no dev-machine proxy is involved). It asserts:

- the caller raises `SoniloError` whose message contains the decoded body text, so the fallback really decoded the body instead of merely swallowing the failure;
- the response is closed — `requests.Response` itself has no `closed` attribute, so this asserts on `response.raw.closed`, which is `True` once the `with` block releases the connection;
- no staged `.sonilo-audio-*` file survives.

Before the change the test fails with `LookupError: unknown encoding: unknown-charset` raised from `sonilo.py:87`; after it, `test/services/test_sonilo.py` is 21 passed / 28 subtests, and `test_sonilo.py` + `test_elevenlabs_music.py` + `test_bgm.py` is 52 passed / 88 subtests. `ruff check` is clean and the added lines introduce no new `ruff format` diff.

One thing this turned up: `elevenlabs_music.py::_safe_response_error()` carries the same decode, introduced by #1131 (3a96206) and still on `main`, with the same reachable `LookupError`. I left it out here so this diff stays focused on the regression you reported — I'll send it as a small separate PR.

c020627:
Sent the sibling as #1378: the same `try/except LookupError` fallback in `elevenlabs_music.py::_safe_response_error()`, plus the mirrored loopback-server regression test, so both copies of the helper are covered. Happy to fold it in here instead if you would rather review them as one change.

harry0703:
Thanks for addressing the charset edge case and adding the loopback-server regression test! The updated local tests passed, including error handling, response closure, and temporary-file cleanup. CI is green, and this is now merged. Appreciate the careful follow-up and the resource-limit improvements!

PR #1375, opened 2026-09-15
fix(api): apply the same Windows filename rules to local material uploads

c020627:
## What this fixes

`POST /video_materials` and the background-music upload are the two endpoints on this
server that accept a client-supplied filename, and they disagree about which names they
take. `app/services/bgm.py:19-20` defines the Windows invalid-character set and the
device-name set, and `app/services/bgm.py:90 sanitize_upload_filename` enforces both.
`app/services/material_upload.py:87 sanitize_material_filename` was added later by
`7003aba` ("fix(api): harden local material uploads"), and that commit copied the
surrounding pipeline — the internal upload prefix, the chunked staging, the atomic
replace — but not these two checks.

Reproduced against the current `main` by calling both sanitizers directly:

```
filename          bgm (audio)                material (video)
clip.mp4          ACCEPT                     ACCEPT
CON.mp4           REJECT (BgmUploadError)    ACCEPT     <== divergent
con.mp4           REJECT (BgmUploadError)    ACCEPT     <== divergent
lpt1.mp4          REJECT (BgmUploadError)    ACCEPT     <== divergent
aux.extra.mp4     REJECT (BgmUploadError)    ACCEPT     <== divergent
bad:name.mp4      REJECT (BgmUploadError)    ACCEPT     <== divergent
bad?.mp4          REJECT (BgmUploadError)    ACCEPT     <== divergent
bad<1>.mp4        REJECT (BgmUploadError)    ACCEPT     <== divergent
quote".mp4        REJECT (BgmUploadError)    ACCEPT     <== divergent
pipe|name.mp4     REJECT (BgmUploadError)    ACCEPT     <== divergent
star*.mp4         REJECT (BgmUploadError)    ACCEPT     <== divergent
```

The material path is reachable from the public API: `app/controllers/v1/video.py:428-431`
passes the client filename straight into `save_material_upload`, which reaches
`sanitize_material_filename`. So a `CON.mp4` upload returns 200 from `/video_materials`
and 400 from the music endpoint.

## Why the UUID storage key is not a reason to skip this

`app/services/bgm.py:101-106` already answers that objection in the source:

> 即使服务端最终使用 UUID，提前拒绝这类名称也能保证 API 在不同平台上的输入行为一致

Both endpoints really do store under a UUID — `app/services/bgm.py:255` and
`app/services/material_upload.py:262` both build `f"{uuid4().hex}{suffix}"` — and
`material_upload.py` already rejects the `.material-upload-` internal prefix even though
that is also only display-relevant. The two validators should not differ.

## The change

`app/services/material_upload.py`

- adds `_WINDOWS_INVALID_FILENAME_CHARS` and `_WINDOWS_RESERVED_FILENAMES` next to the
  existing `_INTERNAL_UPLOAD_PREFIX`.
- `sanitize_material_filename` now rejects the invalid-character set and, using bgm's
  exact device-name form (the segment before the first dot, stripped of trailing dots
  and spaces, upper-cased), rejects `CON`-style basenames.

`test/services/test_material_upload.py`

- adds `test_sanitize_filename_rejects_windows_invalid_and_reserved_names`, mirroring the
  existing `test/services/test_bgm.py::test_sanitize_upload_filename_rejects_invalid_names_and_extensions`.

Names that merely contain a reserved word (`console.mp4`) or place an invalid character
outside the device-name segment (`C:\videos\aux-extra.png`) still upload normally, and that
is asserted in the same test.

## Verification

- The new test fails on `main` — 10 of its 12 names are accepted there — and passes with
  this change: `10 failed` → `29 passed, 61 subtests passed`.
- `test/services/test_material_upload.py`, `test/services/test_bgm.py` and
  `test/services/test_controller_video.py` together: `55 passed, 87 subtests passed`.
- The divergence probe above reports `ACCEPT`/`REJECT` in agreement for all 12 names after
  the change.
- `ruff check` and `ruff format --check` are both clean on the two changed files (the files
  were already `ruff format`-clean on `main`, so the multi-line `any(...)` this PR first
  added was collapsed before committing).

## Scope

The rule list is now defined in three places: `app/services/bgm.py:19-20`,
`app/services/material_upload.py:34-38` and `webui/Main.py:212` (which additionally covers
the Win32 superscript digits). Hoisting them into `app/utils/file_security.py` would remove
the duplication, but that module's test file is currently touched by other work, so this PR
keeps the fix local and reviewable. Consolidating all three copies is a separate change.


harry0703:
Thanks for aligning the local material filename rules with the background-music upload and adding regression tests! The related local tests passed, and we also checked the sanitizer cases on Windows 11 and Windows 10. Merged — appreciate the focused improvement!

PR #1376, opened 2026-09-15
fix(bgm): reject staged upload files when resolving background music

c020627:
## What

`bgm.resolve_bgm_file()` accepts the `.bgm-upload-*` staging files that
`bgm._list_bgm_files()` deliberately hides.

`app/services/bgm.py` writes every upload to a same-directory temp file prefixed with
`_INTERNAL_UPLOAD_PREFIX` before the atomic `os.replace()`. That prefix is guarded in two
of the three places a BGM name can enter the service:

| entry point | guards the staging prefix? |
| --- | --- |
| `sanitize_upload_filename()` | yes — rejects the uploaded name |
| `_list_bgm_files()` | yes — skips it while enumerating |
| `resolve_bgm_file()` | **no** |

`resolve_bgm_file()` is the one that turns a client-supplied value into a path, and it
backs both public entry points:

- `app/services/video.py:662` — `get_bgm_file()`, fed by `params.bgm_file` from the API request
- `cli.py:1521` — `--bgm-file`

So the same file gets two different answers inside one process. Probing a directory that
holds a staged file:

```text
list_bgm_files()                            -> ['builtin.mp3', 'user.flac']   # staged file hidden
resolve_bgm_file('.bgm-upload-abc123.m4a')  -> ACCEPTED .bgm-upload-abc123.m4a
```

The reason the enumeration hides it is already written in the source:

> 临时文件虽然带有合法音频扩展名，但尚未完成校验，不能被随机 BGM 列表提前选中。

An interrupted upload leaves exactly such a file behind — the `finally` cleanup in
`_stage_bgm_upload()` never runs when the process is killed — and a task can then be
pointed at a partially written audio file that was never validated.

## Change

`resolve_bgm_file()` now rejects a basename carrying `_INTERNAL_UPLOAD_PREFIX`. The
comparison is case-insensitive because Windows and macOS resolve these names that way, so
`.BGM-UPLOAD-pending.m4a` cannot slip past on the platforms where the file actually exists.

## Tests

`test_resolve_rejects_staged_upload_left_by_an_interrupted_write` fails on the current code
(3 subtests, `ValueError not raised`) and passes with the fix. It also asserts that an
ordinary uploaded file still resolves, so the guard cannot be widened by accident.

- `test/services/test_bgm.py`: 17 passed, 43 subtests passed
- `ruff check app/services/bgm.py test/services/test_bgm.py`: clean

`ruff format --check` reports pre-existing long-line splits in both files on `main` as well
(this repository is not format-clean and CI only runs `ruff check`), so no reformatting was
applied; the newly added lines are not part of that diff.


harry0703:
Thanks for closing this gap between BGM listing and explicit file resolution, and for adding regression coverage! The 201 related local tests passed, and real-file checks confirmed that staged uploads are rejected while normal uploaded and built-in music still resolve correctly. Merged — appreciate the focused fix!

PR #1378, opened 2026-09-16
fix(bgm): keep an unknown upstream charset from bypassing the ElevenLabs error path

c020627:
## What

`elevenlabs_music._safe_response_error()` decodes the bounded error body with `response.encoding`, which `requests` copies verbatim from the upstream `Content-Type` charset. When that charset is unknown — a proxy or CDN answering `Content-Type: text/plain; charset=unknown-charset`, for example — `codecs` raises `LookupError`. `errors="replace"` does not cover it: it only affects `UnicodeDecodeError`.

That exception escapes both callers, because each one only translates `requests.RequestException`:

- `_request_bgm()` (`elevenlabs_music.py:341`) — `generate_bgm()` callers degrade on `ElevenLabsMusicError`, so a `LookupError` bypasses the intended no-BGM fallback and fails the task instead;
- `test_connection()` — the WebUI probe at `webui/Main.py:6437-6443` only catches `ElevenLabsPaidPlanRequiredError` and `ElevenLabsMusicError`, so the raw codec error escapes the handler and the localized "Connection Test Failed" message is never shown.

## Why this is a regression against `response.text`

#1131 (3a96206) replaced `(response.text or "")` with the explicit bound read plus `.decode(response.encoding)`. `requests.Response.text` catches `LookupError`/`TypeError` internally and retries the decode without the declared charset, so the previous implementation tolerated exactly this response. This PR restores that behaviour with a `try/except LookupError` fallback to `utf-8` + `errors="replace"`. The single bounded `iter_content()` read — the resource limit #1131 added — is unchanged.

## Reproduction

A loopback `ThreadingHTTPServer` answering `500` with `Content-Type: text/plain; charset=unknown-charset`:

- before: `LookupError: unknown encoding: unknown-charset` raised from `elevenlabs_music.py:93`, straight out of `_request_bgm()`;
- after: `ElevenLabsMusicError: ElevenLabs generation failed (500): unknown-charset upstream failure`.

`test_request_bgm_reports_unknown_charset_body_as_music_error` drives the generation path through a `trust_env = False` session (the loopback-only pattern `test_youtube_audience.py` uses, so no dev-machine proxy is involved) and asserts:

- the caller raises `ElevenLabsMusicError` whose message contains the decoded body text, so the fallback really decoded the body instead of merely swallowing the failure;
- the response is closed — `requests.Response` has no `closed` attribute itself, so the test asserts on `response.raw.closed`;
- no staged `.elevenlabs-music-*` file survives.

Checks: `test/services/test_elevenlabs_music.py` 16 passed / 20 subtests; `test_elevenlabs_music.py` + `test_sonilo.py` + `test_bgm.py` 52 passed / 91 subtests; `ruff check` clean, and the added lines introduce no new `ruff format` diff.

## Note

`sonilo.py::_safe_response_error()` carries a copy of this helper with the same defect. That copy is fixed in #1374, which is where this was reported — I kept the two changes in separate PRs. Both copies are covered by a regression test now.


harry0703:
Thanks for spotting and fixing the same issue in the ElevenLabs provider! The local regression tests and additional connection-error checks passed, and CI is green. Merged — appreciate you keeping both providers consistent while preserving the bounded error-body read.

PR #1379, opened 2026-09-16
fix(skill): name the Pexels key environment variable in the credential report

c020627:
## What

`report_missing_config()` and `report_invalid_pexels_config()` now print
`PEXELS_API_KEY_ENV=MPT_PEXELS_API_KEY` next to the existing Pexels signup links.

## Why

`apply_environment_config()` can write five provider keys from the environment, and each of the other four is announced by its report branch:

| provider | line the report prints |
|---|---|
| volcengine_seedance | `VOLCENGINE_ARK_API_KEY_ENV=MPT_VOLCENGINE_ARK_API_KEY` |
| ofox | `OFOX_API_KEY_ENV=MPT_OFOX_API_KEY` |
| metaso_minimax | `METASO_MINIMAX_API_KEY_ENV=MPT_METASO_MINIMAX_API_KEY` |
| muapi | `MUAPI_API_KEY_ENV=MPT_MUAPI_API_KEY` |
| **pexels** | *(none)* |

Pexels reads `MPT_PEXELS_API_KEY` at `docs/skill/mpt_agent.py:232`, so the write path exists and the report is the only place that never names it. `SKILL.md` lists `MPT_PEXELS_API_KEY` among the variables to set, and `MPT_NEEDS_INPUT` is documented as the list the agent acts on ("Ask only for the listed values"), so the machine-readable hint is the one that has to be complete.

The omission costs more for Pexels than for the other four. Each generated source also has a runtime fallback to a bare provider variable (`OFOX_API_KEY`, `MUAPI_API_KEY`, `METASO_MINIMAX_API_KEY`, `VOLCENGINE_ARK_API_KEY`), so its printed line is a convenience. The material-search sources live in `app/services/material.py`, which reads no environment variable at all:

```
$ grep -n "os.getenv\|os.environ\|getenv" app/services/material.py
(no output)
```

`material.get_api_key()` (`app/services/material.py:175-181`) reads `config.app` and raises `ValueError` when the key is unset. So `MPT_PEXELS_API_KEY` is the only way to satisfy the Pexels check without editing `config.toml` by hand, and a caller that reruns with only the printed hints cannot.

Pexels is also the only material-search source `apply_environment_config()` can configure at all: `pixabay`, `coverr`, and `wavespeed` have no write path, so there is no variable to name for them and adding one would need a new channel. That is left out of this change on purpose.

## Tests

`test_only_missing_pexels_key_does_not_ask_for_llm_again` now asserts the new line, and `test_invalid_pexels_key_reports_the_same_key_environment` covers the rejected-keys path, which had no direct test before. Both fail before this change (`'PEXELS_API_KEY_ENV=MPT_PEXELS_API_KEY' not found`) and the file passes afterwards: `36 passed`.

No behaviour changes beyond the two extra lines of output.


c020627:
Gentle ping — this has been green and conflict-free since it was opened, and nothing is pending on my side.

If it is not a fit for the repo, feel free to close it; if you would like anything changed, I can pick it up right away.

Crystal666988:
这么吊

PR #1380, opened 2026-09-16
test(webui): run the headless task-action tests on Windows

c020627:
## What happens today

`test/services/test_webui_headless_task_actions.py` is red on Windows, in two independent ways.

### 1. The fixture spoofs `sys.platform` before the WebUI's import chain is loaded

```python
monkeypatch.setattr(sys, "platform", "linux")   # test_webui_headless_task_actions.py:35
```

The headless branch of `_open_task_path()` / `_open_task_video()` is selected by
`webui/Main.py:1051 _is_headless_server()`, which returns `False` early for
`darwin`/`win`, so the test has to pretend to be Linux. But `sys.platform` is a
process-global that third-party libraries also read at import time, and
`AppTest.from_file()` re-executes `webui/Main.py` *inside* the fixture's scope.
`webui/Main.py:47` pulls in `app.services.material`, which imports `moviepy`,
which resolves the FFmpeg binary at import time (`moviepy/config.py:41` →
`imageio/plugins/ffmpeg.py:173` → `imageio_ffmpeg/_utils.py`) and initialises
`numpy` (whose `hugepage_setup()` calls `os.uname()`). Both then branch on the
spoofed platform:

```
File ".../webui/Main.py", line 47, in <module>
File ".../app/services/material.py", line 15, in <module>
File ".../moviepy/config.py", line 41, in <module>
File ".../imageio/plugins/ffmpeg.py", line 173, in get_exe
File ".../imageio_ffmpeg/_utils.py", line 33, in get_ffmpeg_exe
RuntimeError: No ffmpeg exe could be found.
```

and, when `numpy` has not been imported yet,
`AttributeError: module 'os' has no attribute 'uname'`.

Because those modules are process-wide singletons, which failure appears depends
on what ran earlier. In a full `pytest test` run `numpy`/`moviepy` are already
imported, so only problem 2 shows up; running this file on its own — the normal
way to work on it — fails both tests. `test_video.py:1369` shows the pattern
used elsewhere in the suite, `with patch("sys.platform", "win32")`, scoped to
the call under test; this fixture is the only place in the repo that mutates
`sys.platform` for the whole process.

### 2. The assertion hardcodes a POSIX separator

`_open_task_path()` builds the hint as
`f"{tr('Open Task Folder')}: ./storage/{rel_path}"`, where
`rel_path = os.path.relpath(...)` follows the platform separator. On Windows the
toast reads `./storage/tasks\headless-test`, so the hardcoded
`"./storage/tasks/headless-test"` can never match.

This one is reachable without the fixture at all — it is the failure a full-suite
Windows run reports today.

## Why it is worth wiring up

`_is_headless_server()` exists precisely for the Docker/server deployment the
project supports; the README covers Windows and CI runs a `Windows smoke tests`
job. That job only runs seven hand-picked files, so this one is never checked
there, and a Windows contributor sees a red test that has nothing to do with
their change.

## Change

- Import the module whose import chain resolves the platform-dependent resources
  (`app.services.material`) before the fixture spoofs `sys.platform`, so FFmpeg
  and `numpy` are resolved for the host that is actually running the suite.
- Build the expected hint with `os.path.join()` so it matches however the
  platform renders the separator. The `./storage/` prefix and the
  `tasks`/`headless-test` segments stay pinned.

## Verification

Windows 11, Python 3.13.13, file run on its own:

| command | before | after |
| --- | --- | --- |
| `pytest -q test/services/test_webui_headless_task_actions.py` | **2 failed** | **2 passed** |

Both halves are independently necessary: with only the import fix in place the
file still reports `1 failed, 1 passed`, failing on the hardcoded separator.

Neighbouring WebUI test files stay green (`test_webui_headless_task_actions`,
`test_webui_task`, `test_webui_task_history`, `test_webui_bgm`,
`test_webui_startup` → 68 passed).

On Linux both `os.path.join("tasks", "headless-test")` and the toast render as
`tasks/headless-test`, so the assertion is unchanged there and the Python 3.11 /
3.13 jobs keep passing. `ruff check` is clean, and the file adds no new
`ruff format --diff` hunk (1 before, 1 after, both pre-existing).


harry0703:
Thanks for fixing the Windows import-order and path-separator issues while preserving the assertions! We verified the PR on Windows 11 with Python 3.12 and Windows 10 with Python 3.13 using the locked dependencies: both target tests passed independently, and all 45 related WebUI tests passed on each machine. CI is also green. Merged — appreciate the focused improvement!

PR #1381, opened 2026-09-16
fix(api): reject superscript Windows device names on upload

c020627:
## What happens today

The three copies of `_WINDOWS_RESERVED_FILENAMES` disagree about what a reserved name is.

`webui/Main.py:212` carries the full official list — 28 entries, including the Latin-1 superscript
device names `COM¹`/`COM²`/`COM³` and `LPT¹`/`LPT²`/`LPT³`. `d7d4a13`
(*"fix(webui): cover all Windows reserved names"*) expanded it, and
`test/services/test_webui_task_history.py` now pins that list explicitly.

`app/services/bgm.py:19` and `app/services/material_upload.py:34` hold the same constant under the
same name, but only 22 entries:

```python
| {f"COM{index}" for index in range(1, 10)}
| {f"LPT{index}" for index in range(1, 10)}
```

Three separate comments say this rule is meant to be one rule:

- `bgm.py` — the device-name rejection exists so that *"API 在不同平台上的输入行为一致"*.
- `material_upload.py` — *"Keep the same rule as `bgm.sanitize_upload_filename`"*.
- `test/services/test_material_upload.py` — *"the two endpoints must agree on which client-supplied
  names they accept instead of diverging by platform"*.

The result is that both upload sanitizers reject `COM1.mp3` but accept `COM¹.mp3`:

| filename | `bgm.sanitize_upload_filename` | `material_upload.sanitize_material_filename` |
| --- | --- | --- |
| `COM1.mp3` / `LPT1.wav` | rejected | rejected (`.mp4`/`.webm`) |
| `COM¹.mp3` / `LPT².wav` | **accepted** | **accepted** |
| `COM².mp3` / `COM³.mp3` | **accepted** | **accepted** |

Same rule, applied to a different set of names — and the upload endpoints disagree with the download
filename builder about which client-supplied names are reserved.

Both are reachable from the public API, not just the WebUI: `POST /bgm` →
`save_bgm_upload()` → `sanitize_upload_filename()`, and `POST /video_materials` →
`save_material_upload()` → `sanitize_material_filename()`.

## Change

Extend both lists to the same 28 entries the WebUI copy already uses, and cover the superscript names
in the existing reserved-name tests. The stem is still the only part compared against the list, so a
name that merely contains a superscript digit elsewhere — `song¹.mp3` — stays an ordinary upload;
that boundary is asserted too.

## Verification

`pytest -q test/services/test_bgm.py test/services/test_material_upload.py` → **30 passed, 70 subtests**
(was 30 passed before the test cases were added, so nothing else changed behaviour).

Both halves are independently pinned: reverting only the two service constants, and keeping the new
test cases, fails exactly the six added subtest cases and nothing else —

```
SUBFAILED(filename='COM¹.mp3') test_bgm.py::test_sanitize_upload_filename_rejects_invalid_names_and_extensions
SUBFAILED(filename='lpt².wav') ...
SUBFAILED(filename='com³.flac') ...
SUBFAILED(filename='COM¹.mp4') test_material_upload.py::test_sanitize_filename_rejects_windows_invalid_and_reserved_names
SUBFAILED(filename='lpt².webm') ...
SUBFAILED(filename='com³.png') ...
6 failed, 30 passed, 64 subtests passed
```

Every test module that imports either service is green — `test_bgm`, `test_cli`, `test_controller_video`,
`test_elevenlabs_music`, `test_material_upload`, `test_sonilo`, `test_webui_bgm`,
`test_webui_settings_transfer` → **234 passed, 180 subtests**.

`ruff check` (0.15.21, the version pinned in `pyproject.toml`) is clean, and the new lines add no
`ruff format --diff` hunk — hunk counts are unchanged from `783321a` for all four files.

## Notes

I would normally keep this to three files, but the change necessarily spans the two service-layer
copies plus each module's own test. No new dependency, no behaviour change beyond the rejection set,
and none of the four files is touched by an open PR.


harry0703:
Thank you for catching this inconsistency and for the focused fix and regression tests! Merged. We also verified the related test suites locally (234 tests and 180 subtests passed), along with 72 additional filename boundary cases. This keeps the upload validation consistent with the existing download filename rules while preserving ordinary filenames containing superscript digits. Thanks again for your contribution!

PR #1383, opened 2026-09-16
fix(queue): drain past queued tasks the worker cannot dispatch

c020627:
`RedisTaskManager.dequeue()` already discards a queued entry whose params no longer pass `VideoParams` validation and moves on to the next one. The function-name lookup right above it does not follow the same policy:

```python
task_info = json.loads(task_json)
# 将函数名称转换回函数对象
task_info["func"] = FUNC_MAP[task_info["func"]]   # <- KeyError escapes here
```

For a payload whose entry function is no longer in the map, this raises `KeyError` **before** the discard branch below can run.

## Why such payloads exist

`FUNC_MAP` membership changes over time, and the repo still shows it:

```python
FUNC_MAP = {
    "start": tm.start,
    # 'start_test': tm.start_test
}
```

That commented-out `start_test` entry is a leftover from 17b4a61 (2024-07). A Redis queue outlives a deployment — it is a separate service and `enqueue` persists JSON — so after an upgrade the queue can hold entries whose entry function this version cannot dispatch. This is the same migration scenario the neighbouring params check was written for ("queued under an older, more permissive schema, or corrupted").

## Why it is more than one dropped task

`check_queue()` calls `dequeue()` **while holding `self.lock`** (`app/controllers/manager/base_manager.py:66-72`), and it is reached only through `task_done()`:

```
run_task()  base_manager.py:60-64   try: func(...)  finally: self.task_done()
task_done() base_manager.py:92-95   with self.lock: current_tasks -= 1
                                    self.check_queue()
```

So a `KeyError` from `dequeue()` escapes `check_queue()`, then `task_done()`, then `run_task()`'s `finally` — the worker thread dies. Nothing else calls `check_queue()`, so the remaining queued tasks sit in `processing` until some later task happens to finish and re-enter `task_done()`.

Measured on the unfixed code (`max_concurrent_tasks=1`, one poison entry followed by a valid task, `task_done()` called on a finished task):

```
[C] task_done 抛出 KeyError: 'start_test'
[C] current_tasks=0
[C] execute_task 调用次数=0
```

The valid second task is never scheduled, and its state record stays `processing`.

## Change

Route the parse, the name lookup and the kwargs lookup through one guard, so every unusable shape is discarded exactly the way the params path already does it:

| queued entry | before | after |
| --- | --- | --- |
| truncated JSON payload | `JSONDecodeError` escapes `dequeue` | logged, discarded |
| JSON that is not an object | `TypeError` escapes `dequeue` | logged, discarded |
| unknown / missing entry function | `KeyError` escapes `dequeue` | logged, discarded, task marked `FAILED` |
| params failing current validation | discarded (unchanged) | discarded (unchanged) |

Entries whose `task_id` is readable are still converged to `FAILED` with `patch_task`, matching the existing branch — the task record is created before `enqueue` and defaults to `processing`, so dropping the queue entry alone would leave it spinning in the UI forever. Payloads that cannot be parsed at all have no readable `task_id`, so there is nothing to converge.

Same behaviour on the fixed code:

```
[C] current_tasks=1
[C] execute_task 调用次数=1
```

## Follow-up: three more payload shapes (commit `06c150a`)

Three cases still escaped the same discard policy. Each one now goes through it, and each has a regression test that queues a **valid task behind the broken entry** and asserts that draining continues.

| broken entry, valid task behind it | before | after |
| --- | --- | --- |
| empty value (`b""` / `""`) | `if not task_json` read it as "lpop found nothing" → `check_queue()` returned early and the valid task stayed queued | treated as an unusable entry, discarded; only `None` now means the queue is exhausted |
| `args: null` (also `{}`, `5`, `"x"`) | passed the parse; `*args` on `None` raised `TypeError: Value after * must be an iterable, not NoneType`, which `check_queue()`'s handler answered by **re-enqueueing** the entry and letting the exception escape | rejected as an unusable entry; a **missing** `args` field still defaults to `()`, and a valid `args` list is untouched |
| unknown function + `task_id: ["task-a"]` | `patch_task(["task-a"], ...)` reached redis and raised `DataError: Invalid input of type: 'list'. Convert to a bytes, string, int or float first.` | the state update is skipped when no usable string id is available; both discard paths carry the same guard |

Measured against a real Redis instance (`db 15`, unmocked queue and unmocked `RedisState`), one broken entry followed by one valid task, `check_queue()` driving the drain:

```
empty value
  before  check_queue() returns early, valid task still queued (dispatched=[])
  after   dispatched, queue empty
args: null
  before  TypeError escapes check_queue, malformed entry re-enqueued (queue grows)
  after   dispatched, queue empty
unknown func + list task_id
  before  DataError escapes dequeue, valid task still queued
  after   dispatched, queue empty
```

Regression control: an entry with an expired params payload and a valid **string** `task_id` is still converged to `FAILED` through the real `RedisState` Lua script, unchanged.

## Verification

| item | result |
| --- | --- |
| `test/services/test_task_manager.py` | 27 passed, 20 subtests (20 existing + 7 new) |
| only `app/controllers/manager/redis_manager.py` reverted to the previously reviewed revision, new tests kept | **exactly the 8 new cases fail** — 6 × `AssertionError: 1 != 2` (dequeue popped once and handed the broken entry over) and 2 × `Expected 'patch_task' to not have been called. Called 1 times. Calls: [call(['task-1'], ...)]`; everything else passes |
| modules referencing the task manager (`test_task_manager.py`, `test_controller_video.py`, `test_webui_task.py`) | 69 passed, 46 subtests |
| full suite | 11060 tests, 1 failure — the pre-existing Windows `symlink` case in `test_asgi_static_files` (fails standalone on this machine too, passes on Linux CI) |
| `ruff check` (locked 0.15.21) | All checks passed |
| `ruff format` baseline | same hunk count as the unmodified files |

2 files, +254/−18. No new dependency, no public signature change, no behaviour change on the happy path.

## Notes

- I did not touch `base_manager.py`. Its `check_queue()` already documents `dequeue()`'s skip-and-discard contract and handles the `None` return, so fixing `dequeue()` is the smaller change. The positional-argument container is validated in `dequeue()` for the same reason: that is where the other unusable-queued-entry shapes are already handled.
- I deliberately left the existing params-validation branch in place rather than folding the two paths into a helper: the duplicate three-line `patch_task` call is the price of a diff that stays reviewable. The follow-up keeps that shape and only adds the `isinstance` guard to both call sites.
- This branch was rebuilt on the current `main` (it now includes #1387, which touches `test/services/test_task_manager.py`), so the previous head `f3df033` is no longer on the branch; its content is commit `e161570` and the follow-up is `06c150a`.
- The follow-up delta on its own is +120/−6. Most of it is tests: each of the three cases needed a regression test that puts a valid task behind the broken entry and asserts the drain continues.


harry0703:
Thank you for the focused fix and the regression tests! We verified the related suites locally (62 tests and 26 subtests passed) and also tested against an isolated real Redis instance. The worker successfully drained past nine unusable entries, dispatched the following valid task, persisted the relevant failed states, and released its concurrency slot. The deleted-task case also behaved correctly.

During additional boundary testing of f3df033, we found three malformed-payload cases that can still interrupt draining:

1. An empty Redis value (an empty byte string) is treated as an empty queue by `if not task_json`. With a valid task behind it, dequeue returns None and leaves that task waiting. Only None should indicate that LPOP found no entry; an empty value should be discarded and draining should continue.
2. A payload with a known function and valid kwargs but `args: null` passes dequeue. Expanding `*args` then raises TypeError in check_queue, and its exception handler re-enqueues the malformed task. Please validate the positional-argument container before returning a task for dispatch, while preserving the existing default for a missing args field.
3. An unknown-function payload with a truthy, non-string task_id, such as a JSON array, reaches patch_task and raises redis.exceptions.DataError. This interrupts draining before the following valid task can run. Please validate the task ID before attempting the state update, and skip the update when no usable ID is available.

These are remaining malformed-data edge cases, not regressions introduced by this PR, and the normal enqueue path should not produce them. We consider them non-blocking follow-up hardening opportunities rather than a reason to discount the main fix. It would be helpful to cover each with a regression test that places a valid task afterward and verifies that draining continues, without re-enqueuing the unusable entry or recreating a deleted state record.

Thanks again for improving queue reliability!

c020627:
Thanks — all three reproduce on the reviewed revision, and they are fixed in `06c150a`. The branch has been rebuilt on the current `main` (it now includes #1387, which touches the same test file), so the delta since your review is that one commit: `app/controllers/manager/redis_manager.py` + `test/services/test_task_manager.py`, +120/−6.

| broken entry, valid task queued behind it | before | now |
| --- | --- | --- |
| empty value (`b""` / `""`) | `if not task_json` read it as "LPOP found nothing" → `check_queue()` returned early and the valid task stayed queued | discarded as an unusable entry; `is None` is now the only "queue exhausted" signal |
| `args: null` | parse passed, so `*args` on `None` raised `TypeError` in `check_queue()`, whose handler answered by **re-enqueueing** the entry and letting the exception escape | rejected before dispatch; a **missing** `args` field keeps the existing `()` default |
| unknown function + `task_id: ["task-a"]` | `patch_task(["task-a"], ...)` reached redis → `DataError: Invalid input of type: 'list'. Convert to a bytes, string, int or float first.` | the state update is skipped when no usable string id is available, on both discard paths |

Each case has a regression test that queues a valid task behind the broken entry and asserts that draining continues, that the unusable entry is not re-enqueued, and that no state record is rewritten for it: `test_dequeue_discards_payloads_that_cannot_be_dispatched`, `test_dequeue_ignores_non_string_task_id_in_every_discard_path`, plus `test_dequeue_dispatches_payload_without_args_field`, which pins the preserved default for a missing `args` field.

Verification:

- With only `redis_manager.py` reverted to the reviewed revision and the new tests kept: exactly 8 failures, all new — 6 × `AssertionError: 1 != 2` and 2 × `Expected 'patch_task' to not have been called. Called 1 times. Calls: [call(['task-1'], ...)]`.
- Against a real Redis instance (db 15, unmocked queue and unmocked `RedisState`), one broken entry followed by one valid task, `check_queue()` driving the drain: before, the empty value left `check_queue()` returning with `dispatched=[]`, `args: null` raised and grew the queue by re-enqueueing, and the list `task_id` raised `DataError` out of `dequeue()` with the valid task still queued. After, all three dispatch the valid task and leave the queue empty.
- Control: an expired-params entry with a valid **string** `task_id` is still marked `FAILED` through the real `RedisState` Lua script.
- `test_task_manager.py` 27 passed / 20 subtests; together with `test_controller_video.py` and `test_webui_task.py`, 69 passed / 46 subtests. Full suite 11060 tests with only the pre-existing Windows `symlink` case failing (`test_asgi_static_files`, fails standalone on this machine, passes on Linux CI). `ruff check` clean, no new `ruff format` hunk.

I left the two discard paths as separate call sites instead of extracting a helper, so the delta stays readable — the bug you found in the second path would have been caught by the new `subTest` for `stale params`.

harry0703:
Merged — thank you for the focused reliability improvement and for addressing all three additional malformed-payload cases!

We verified the updated implementation locally: 74 regression tests and 46 subtests passed, along with Ruff and all three CI checks. We also tested 18 scenarios against an isolated real Redis instance, confirming that unusable entries are discarded, the following valid task runs in a worker thread, concurrency slots are released, applicable failed states are persisted, and deleted task records are not recreated.

Thanks again for the thorough regression coverage and detailed follow-up!

PR #1384, opened 2026-09-16
fix(upload): reject C1 controls and bidi overrides in uploaded file names

c020627:
## Problem

Both upload validators reject C0 control characters, and nothing else from the same class:

```python
# app/services/bgm.py:122 (before this change)
or any(ord(character) < 32 for character in safe_name)
```

```python
# app/services/material_upload.py:119 (before this change)
or any(ord(character) < 32 for character in safe_name)
```

`ord(...) < 32` covers U+0000-U+001F only. The names below are accepted today by
`bgm.sanitize_upload_filename()` **and** `material_upload.sanitize_material_filename()`,
even though both functions are documented as returning a *display-safe* basename
(`material_upload.py:113`) and both values are written verbatim into log lines as
`original_name={safe_name}`:

| name | what it does |
| --- | --- |
| `song\x7f.mp3`, `song\x9b.mp3` | C1 controls, never valid in a filename |
| `song\x85.mp3` | U+0085 NEXT LINE — rendered as a line break by some log viewers, so one filename field forges an extra, independent-looking log line |
| `photo\u202egnp.mp3` | U+202E RIGHT-TO-LEFT OVERRIDE — everything after it is displayed reversed, so the name reads as a different extension in a file manager or a log |
| `song\u200e.mp3`, `song\u2069.mp3` | U+200E LRM / U+2069 PDI, the rest of the bidi control set |
| `song\u2028.mp3`, `song\u2029.mp3` | Unicode line/paragraph separators, same visual line-break problem as U+0085 |

The project already treats this character class as unsafe for anything that reaches
logs or responses: `base.normalize_task_id()` rejects it via `str.isprintable()`, and
`test/services/test_controller_base.py:52` lists `"unicode\u2028separator"` among the
values that must be replaced by a generated UUID. The two upload validators are the
outliers.

## Fix

Replace the C0-only predicate with an explicit `_UNSAFE_FILENAME_CHARACTERS` set that
covers C0 **and** C1 controls (U+0000-U+001F, U+007F-U+009F), the bidi formatting
controls (U+200E, U+200F, U+202A-U+202E, U+2066-U+2069) and U+2028/U+2029.

The set is deliberately character-specific rather than category-wide. `str.isprintable()`
would additionally reject every non-ASCII space (U+00A0 and friends), and rejecting all
of `Cf` would reject U+200D (ZWJ) and U+FE0F, which emoji file names legitimately use.
Both validators keep their own copy of the set, matching how
`_WINDOWS_RESERVED_FILENAMES` is already duplicated between the two upload services
(see #1381).

None of the rejected characters can be typed into a file-name field, so ordinary
uploads — including CJK, combining marks, curly quotes, U+00A0, U+FE0F and ZWJ — are
unaffected.

## Tests

`test/services/test_bgm.py` and `test/services/test_material_upload.py` each gain:

* `..._rejects_control_and_bidi_characters` — the eight newly rejected characters, plus
  the three C0 cases as regression protection.
* `..._keeps_printable_unicode_names` — CJK, combining marks, curly quotes, U+00A0,
  U+FE0F and ZWJ still round-trip, pinning the rule to specific characters.

Reverting only the two source files while keeping the new tests fails exactly the 16 new
sub-cases (`AssertionError: MaterialUploadError not raised`, `BgmUploadError not raised`);
the C0 cases keep passing, which shows the change is purely additive.

Full suite: `11001 passed, 19 skipped, 0 failed`. `ruff check` clean; the change adds no
`ruff format` delta to either file.

## Scope

`webui/Main.py:211` applies a separate pattern to *download* file names
(`[<>:"/\\|?*\x00-\x1f]`) derived from the video subject. That is a different entry point
and is not changed here.


harry0703:
Thank you for improving upload filename validation and adding regression coverage! Merged. We verified the related suites locally (76 tests and 146 subtests passed), plus 168 additional character checks covering all blocked code points and ordinary Unicode filenames. Thanks for keeping the validation focused while preserving multilingual and emoji names.

PR #1385, opened 2026-09-16
fix(video): apply the sibling clip-duration rules to Seedance and OFox

c020627:
fix(video): apply the sibling clip-duration rules to Seedance and OFox

The four paid text-to-video clients under app/services no longer agree on how a
clip duration is reported, to the caller and to the log.

An unparsable duration is a clean client error in two of them
(app/services/muapi.py:209-212 -> MuAPIError,
app/services/metaso_minimax.py:187-204 -> MetasoMiniMaxError) but leaks a raw
TypeError/ValueError in the other two (app/services/volcengine_seedance.py:227,
app/services/ofox.py:227). Callers and tests handle the client's own error class,
so a raw ValueError escapes every ``except VolcEngineSeedanceError`` /
``except OFoxError`` handler. app/services/material.py:1984-1993 already raises
"Seedance clip duration must be a positive integer" for exactly this condition
one level up, which is the intended contract.

A duration outside the configured range is replaced by a different one, and the
other paid sources say so before submitting the paid task: OFox at
app/services/ofox.py:230-237, MuAPI at app/services/muapi.py:215-220, Metaso at
app/services/metaso_minimax.py:227-232 and the WaveSpeed path at
app/services/material.py:742-749. Seedance is the only one that silently submits
a different duration, so a user who narrows volcengine_seedance_max_duration
(config.example.toml:119-120) gets shorter clips than requested with nothing in
the log to explain the substitution.

Behaviour is otherwise unchanged: a duration inside the range is still submitted
as-is, and non-positive values still clamp through max(int(value), 1) before
_duration_bounds(), matching MuAPI. The WaveSpeed path in app/services/material.py
has the same unguarded conversion and is left to a separate change so this diff
stays focused.

Tests: test_volcengine_seedance.py + test_ofox.py -> 80 passed, 87 subtests.
The three new cases fail on the pre-change sources (7 failed / 79 passed with
only app/services/*.py reverted) and pass after, so they really pin the change.
The clamp is asserted through a temporary loguru sink at INFO level, and a
matching case keeps an in-range duration from being reported as clamped.
Full suite: 1279 cases, 0 failures, 0 errors, 19 skipped. ruff check (0.15.21)
clean and no new ruff format hunks.


harry0703:
Thank you for the focused error-handling and logging improvements! Merged. Our related regression tests passed (243 tests and 110 subtests, with 3 skipped). Additional checks confirmed that invalid durations, including NaN and infinity, are rejected before any HTTP submission, and 18 before/after payload comparisons confirmed that existing duration clamping and request payloads remain unchanged. Thanks again for your contribution!

PR #1386, opened 2026-09-16
ci(smoke): cover the headless task-action tests on Windows

c020627:
ci(smoke): cover the headless task-action tests on Windows

test_webui_headless_task_actions.py asserts a POSIX path fragment in the
headless "open folder" toast. _open_task_path() builds that fragment with
os.path.relpath(), so on Windows the toast reads
./storage/tasks\headless-test and the literal could never match.

783321a repaired the assertion but left the file outside the Windows smoke
job. That job runs seven hand-picked files and is the only place a
Windows-only drift like this can surface: spoofing sys.platform is enough to
get the headless branch, but the separator still follows the host, so the
Linux job stays green either way.

#1260 hit the same shape of failure -- test_webui_task.py asserted the POSIX
form of a relative log path -- and added that file to the job "to keep the
regression covered on the platform where it appears". This is the other half
of that same repair.

Checked on Windows: the job's own file list passes with the file added (155
passed, 4 skipped, 44 subtests). Restoring the pre-783321a literal makes
test_headless_open_folder_shows_host_mapped_path fail on Windows while it
still passes on Linux, which is exactly the gap this closes.


harry0703:
Thank you for closing this Windows CI coverage gap! Merged. We confirmed in the Windows job log that the headless task-action tests were actually included and that the expanded smoke suite passed (155 tests and 44 subtests, with 4 skipped). This is a useful follow-up to the earlier portability fix and will help catch future platform-specific regressions.

PR #1387, opened 2026-09-16
fix(queue): coerce the configured task limits before scheduling

c020627:
fix(queue): coerce the configured task limits before scheduling

`max_concurrent_tasks` and `max_queued_tasks` reach the scheduler exactly as
TOML parsed them. A quoted number is a known way to write them: the repository
already documents the problem for `claude_code_timeout` ("TOML 既可能写成
`300`，也可能写成 `"300"`", `app/services/llm.py:95`) and `webui_task.py:21`
wraps `max_queued_tasks` in `max(1, int(...))` before constructing its manager.
Only the API path passes both values through unchanged (`v1/video.py:48-49`).

The string then survives into the comparisons in `add_task` as

    TypeError: '<' not supported between instances of 'int' and 'str'

For `max_concurrent_tasks` that is every request. For `max_queued_tasks` it is
worse: the queue guard is evaluated only once every concurrency slot is taken,
so the misconfiguration stays dormant while the server is idle and then fails
every request under load. `v1/video.py:220-225` deletes the task record and
re-raises, so the caller sees a 500 whose traceback never names the key.

Coerce in the constructor -- the single entry point every caller shares -- and
reject values that cannot be a limit with a message that names the key. Zero
and negative values keep their current meaning (queue without dispatching, then
reject), which `test_add_task_rejects_only_after_queue_limit` relies on.

Other values TOML can produce are now rejected instead of silently changing
meaning (reported by the reviewer, reproduced here):

* `0.5` / `1.5` were `int()`-truncated to 0, so a submitted task was accepted
  into the queue and no worker ever ran it;
* `inf` / `-inf` raised an uncaught `OverflowError` that never named the key,
  while `nan` reached only the generic branch.

All of them now go through the same named validation error, using
`math.isfinite` and `value.is_integer()` -- the shape `coerce_claude_code_timeout`
already uses. Integral floats such as TOML's `max_queued_tasks = 2.0` are still
accepted because no truncation is involved.

Checked with the source reverted and the new tests kept: 10 failing subtests
(`SUBFAILED` lines, 2 fields x {0.5, 1.5, nan, inf, -inf}) and nothing else
failing, so the change is a pure increment. Full suite: 11045 tests,
18 skipped, and the only failure is the pre-existing Windows-symlink case in
`test_asgi_static_files::test_does_not_serve_symlink_to_file_outside_tasks`,
which reproduces without this change. `ruff check` passes and
`ruff format --diff` reports the same hunk count as the unmodified files.


harry0703:
Thank you for the focused configuration fix and the regression tests! We verified the related suites locally (61 tests and 26 subtests passed). We also exercised real worker threads with both the in-memory queue and an isolated Redis instance: quoted limits of "2" correctly allowed two running tasks and two queued tasks, rejected the fifth task, and drained the queue afterward.

Before merging, could you please tighten two edge cases in _coerce_task_limit?

1. Non-integral floats are silently truncated by int(value). With the valid TOML syntax max_concurrent_tasks = 0.5, the PR converts the limit to 0: a submitted task is accepted into the queue but no worker runs it. We reproduced this with actual threads and compared it with the base revision, where the same input still dispatched a task. A fractional task limit is a misconfiguration, so please reject it with an error naming the setting rather than silently converting it into the existing zero-concurrency behavior. Please cover both limit fields, including fractional values such as 0.5 and 1.5.
2. TOML also permits inf and -inf. int(value) raises OverflowError for these, which is not handled by the new helper, so the error still lacks the configuration key. Please reject non-finite values through the same named validation-error path and add coverage for NaN and both infinities in both fields.

The intended quoted-integer behavior works well. Keeping existing integer values, including the deliberately supported zero/negative semantics, while rejecting invalid numeric inputs would make this safe to merge. Thanks again for improving the scheduler!

c020627:
Thanks for the reproduction -- both cases are real, and they are fixed in 85ad45a (still 2 files, `app/controllers/manager/base_manager.py` + `test/services/test_task_manager.py`, +96/-2).

| configured value | before this follow-up | now |
| --- | --- | --- |
| `0.5` / `1.5` | `int()` truncated to `0`: the task was accepted into the queue and no worker ever ran it | `ValueError: max_concurrent_tasks must be an integer, got 0.5` |
| `inf` / `-inf` | uncaught `OverflowError: cannot convert float infinity to integer`, no key name | `ValueError: max_concurrent_tasks must be a finite integer, got inf` |
| `nan` | `ValueError`, but only via the generic branch | `ValueError: ... must be a finite integer, got nan` |
| `"2"` (quoted) | coerced to `2` | unchanged |
| `2.0` | coerced to `2` | unchanged: no truncation is involved |
| `0` / negative | queue without dispatching | unchanged, pinned by `test_zero_concurrency_keeps_queueing_without_executing` |

What changed:

1. `float` values are handled explicitly now: non-finite first (`math.isfinite`, the same shape `coerce_claude_code_timeout` already uses at `app/services/llm.py:95`), then fractional (`value.is_integer()`).
2. `OverflowError` was added to the generic `int(value)` handler, so any other numeric type that cannot be converted reports the key name instead of leaking an unnamed error.
3. `test_non_integral_limits_are_rejected_by_name` covers both fields x {`0.5`, `1.5`, `"0.5"`, `nan`, `inf`, `-inf`} as subtests.

Verification -- source reverted to the reviewed revision, new tests kept:

```
10 failed, 20 passed, 2 subtests passed
SUBFAILED(field='max_concurrent_tasks', value=0.5) test_non_integral_limits_are_rejected_by_name
SUBFAILED(field='max_concurrent_tasks', value=nan) test_non_integral_limits_are_rejected_by_name
SUBFAILED(field='max_queued_tasks', value=-inf)    test_non_integral_limits_are_rejected_by_name
E   OverflowError: cannot convert float infinity to integer
app\controllers\manager\base_manager.py:25: OverflowError
```

Only the new cases fail (`SUBFAILED` lines -- the parent cases still report as passed, so the counts alone would be misleading); the 2 passing subtests are the quoted `"0.5"` ones, which `int()` already rejected. Standalone reproduction against the reviewed revision, using a real `InMemoryTaskManager` with `execute_task` recorded instead of a thread:

```
--- max_concurrent_tasks=0.5 -> max_concurrent_tasks=0
    task accepted: queue_size=1, tasks actually executed=[]
--- max_concurrent_tasks=inf -> OverflowError: cannot convert float infinity to integer
```

With the fix those become `... must be an integer, got 0.5` and `... must be a finite integer, got inf`.

Branch rebased on 552e0ee. `test/services/test_task_manager.py`: 20 tests, 12 subtests. Full suite: 11045 tests, 18 skipped, and the only failure is the pre-existing Windows-symlink case in `test_asgi_static_files` (it reproduces without this diff, and passes on the Linux CI). `ruff check` passes and `ruff format --diff` reports the same hunk count as the unmodified files.

harry0703:
Thank you for addressing both review points and expanding the regression coverage! Merged. We rechecked the updated revision: fractional and non-finite limits now produce clear errors naming the setting, while quoted integers and integral floats retain the intended behavior. Our local verification passed 67 tests and 38 subtests, plus 58 additional TOML boundary checks. Real worker-thread tests with both the in-memory queue and an isolated Redis instance also confirmed concurrency limits, queue-full rejection, queue draining, and slot release. Thanks again for the careful follow-up and contribution!

PR #1388, opened 2026-09-16
ci: lint and compile the agent skill helper

c020627:
ci: lint and compile the agent skill helper

`.github/workflows/ci.yml` runs `ruff check` and `compileall` over a
hand-written path list: `app cli.py main.py webui test`. That list was
introduced in a760e35 (2026-06-15) and only gained the `ruff` step later
(6b4dd6f, 2026-07-14); it has never been extended since.

`docs/skill/mpt_agent.py` landed on 2026-07-12 in b1588e1, so the 797-line
agent skill helper has always been outside both steps. It is the only
tracked Python file in the repository that the two steps do not cover.

Add `docs/skill` to all three lists so both steps really do cover every
Python source in the project. The `compileall` command is duplicated
between the `tests` and `windows-smoke` jobs, so both copies are updated
to keep the two lists identical.

Honest scope: there is no violation in the file today -- running the
widened command on the current tree prints `All checks passed!` -- so this
is a guard, not a fix. Syntax errors were already caught by pytest because
`test/services/test_mpt_agent_skill.py` loads the file through
`importlib`; only the lint half of the gap is real.

Checked that the gap is real by dropping a throwaway file with an unused
variable into `docs/skill/`:

    $ ruff check app cli.py main.py webui test          # today
    All checks passed!
    $ ruff check app cli.py main.py webui test docs/skill
    Found 2 errors.

The probe file was deleted afterwards, and `git status` then showed only
`.github/workflows/ci.yml` as modified.

Also verified:
- `ruff` is pinned to `==0.15.21` in `uv.lock`; the widened command was
  run with that exact version and passes.
- `python -m compileall app cli.py main.py webui test docs/skill` exits 0.
- `yaml.safe_load` parses the file, and every path in the three lists
  exists on disk.

Three lines changed in one file; no runtime behaviour is affected.


harry0703:
Thank you for closing this CI coverage gap! Merged. We ran the expanded lint and compilation commands locally, and all 35 agent-helper tests passed. We also confirmed from the Windows CI log that docs/skill/mpt_agent.py is now included in compilation. Thanks for the focused improvement!

PR #1390, opened 2026-09-16
feat(tts): add VoxCPM reference voice cloning

### LINKED ISSUE #1382: [Feature]: Add reference-audio voice cloning for VoxCPM TTS
## 需求描述 | Problem Statement

The existing VoxCPM provider supports standard text-to-speech with the configured/default voice. For recurring short-video series, creators may need a consistent branded narrator voice. ModelBest's Audio Speech API supports reference-audio voice cloning through an optional `ref_audio` field, but MoneyPrinterTurbo does not expose that capability today.

## 建议的解决方案 | Proposed Solution

If this is within project scope, add an optional VoxCPM-only reference-audio input in the WebUI:

- accept a local reference clip, validate it, and convert it to WAV when necessary;
- enforce ModelBest's documented decoded 5 MiB limit and submit it only as `data:audio/wav;base64,...` in `ref_audio`;
- keep the reference audio task/request scoped: do not persist it in `config.toml`, include it in logs, or retain it after the task;
- invalidate voice-preview caching when the reference clip changes;
- preserve the current standard VoxCPM TTS workflow when no reference audio is provided.

Potential advanced options such as `cfg_value` and `denoise` could remain out of the first iteration unless maintainers prefer to expose them.

Before preparing a PR, could maintainers confirm whether this scope is welcome and whether there are preferred UX, privacy, or file-lifecycle requirements?

lottshin:
Closes #1382

## Summary

- Add optional VoxCPM reference-audio voice cloning in the WebUI.
- Add high-fidelity delivery conditioning with `prompt_audio` and `prompt_text`.
  The default reuses the identity reference; an advanced control accepts a
  separate pacing and emotion example.
- Add optional local Whisper transcription to create an editable transcript
  draft for the active example audio.
- Keep reference audio and transcripts out of `config.toml`, presets, task
  history, and task logs. Temporary conversion and transcription files are
  cleaned up after use.
- Add input limits, conversion timeout, preview-cache invalidation, and
  regression coverage for task isolation and failure paths.

## Verification

- `1273 passed, 19 skipped`
- Ruff and `git diff --check` passed.
- Tested the ModelBest `VoxCPM2` endpoint with standard TTS, identity-only
  cloning, same-audio high-fidelity conditioning, and a separate delivery
  example.
- Tested the browser workflow for reference upload, cache invalidation,
  Whisper transcript drafting, transcript correction, and final playback.

## Scope

- Numeric `cfg_value` and `denoise` controls are intentionally not exposed.
- Whisper transcription is optional and uses the user's existing local Whisper
  configuration; it may download that configured model on first use.


harry0703:
Thank you for implementing this and for the detailed testing, especially the reference-audio handling, task isolation, and live ModelBest verification. This is a useful addition to VoxCPM.

We reviewed a616221 and reproduced two WebUI state-handling issues that we would like to address before merging:

1. **Replacing the example audio retains the previous transcript.** Upload clip A, enable high-fidelity delivery, and enter its transcript. Then replace it with clip B and request a preview. The TTS call receives clip B together with the transcript from clip A. We reproduced this both when reusing the identity reference as the delivery example and when uploading a separate delivery example. Could you bind the transcript to the effective example audio and clear it, or require explicit reconfirmation, when that audio changes? Switching between the shared and separate example modes should also be covered.

2. **An invalid reference upload still allows a plain-TTS preview.** After uploading an invalid WAV, the UI displays the validation error, but the preview button remains enabled. Clicking it calls TTS without reference audio, silently falling back to ordinary speech. Final video generation correctly blocks this error, so the two paths behave differently. Please apply the reference-upload validation to both preview buttons as well as generation, and avoid issuing a synthesis request while the upload is invalid.

For verification, 270 existing related tests passed locally, with 6 skipped, and Ruff passed. Real FFmpeg conversion worked for WAV, MP3, M4A, AAC, OGG, and FLAC. We reproduced the issues above with Streamlit AppTest and mocked the TTS call boundary; we did not make live ModelBest requests or assess cloning quality in this review.

Please add regression tests for these replacement and invalid-upload flows. Thanks again for the contribution and the care you have put into this feature!

lottshin:
Thanks for the detailed reproduction. Addressed both issues in 74730aa.

1. The transcript is now bound to the effective example audio.
   - Replacing the shared identity/example clip clears its transcript.
   - Replacing a separate pacing and emotion example clears its transcript.
   - Switching between shared and separate example modes clears its transcript.
   - Replacing identity audio while a separate example remains active preserves
     that transcript, because the effective example audio has not changed.

2. Both preview buttons now use the same provider validation as generation.
   An invalid VoxCPM reference upload disables sample and full previews, so no
   synthesis request can silently fall back to plain TTS.

Added regression coverage for the shared replacement, separate replacement,
mode-switch, and invalid-preview cases.

Verification:

- `1273 passed, 19 skipped`
- Ruff and `git diff --check` passed.

harry0703:
Merged — thank you for adding VoxCPM reference voice cloning and for promptly addressing the transcript and preview-validation issues!

We verified the fixes locally, including audio replacement, shared/separate example switching, invalid-upload handling, and recovery. The related regression tests and additional boundary checks passed, and all three CI checks are green.

Thanks also for the documentation, privacy considerations, and your live ModelBest testing. We appreciate the contribution!

wwwfalcooncom-lgtm:
نور

wwwfalcooncom-lgtm:
Yemen

PR #1391, opened 2026-09-16
fix(subtitle): preserve timestamps inside SRT cue text

haimuhaimu:
## Problem

`file_to_subtitles()` currently treats any line containing an SRT-looking timestamp as a timing header, including lines inside an active cue. A timestamp example in narration therefore disappears from the text and overwrites the original cue timing. Downstream correction code expects that timing field to contain a start/end pair.

Minimal SRT example:

```srt
1
00:00:00,000 --> 00:00:02,000
Jump to 00:01:23,456 in the recording.
```

Expected: keep the original 0–2 second timing and the complete sentence.
Before this patch: the sentence becomes the timing field and the cue text is empty.

## Change

Only recognize a timing header when no cue is active. Once a cue has started, keep its lines as text until the existing blank-line separator. The existing EOF flush and return structure remain unchanged.

The regression uses real temporary SRT files and covers:
- a timestamp embedded in a sentence;
- a timestamp-only body line;
- a complete timing range used as body text;
- multiline text, a following cue, and a final cue without a trailing blank line.

All three new cases failed against the original implementation and pass with the fix.

## Verification

Tested against upstream `3c9873014c774d73b8fb60531bae25b4b4ec0a87`, with Python 3.11 on macOS:

- `python -X utf8 -m pytest -q test/services/test_subtitle.py`: 12 passed, 3 subtests passed.
- `MPT_RUN_INTEGRATION_TESTS=0 python -X utf8 -m coverage run -m pytest -q test`: **1,273 passed, 19 skipped, 9,757 subtests passed**.
- `python -m coverage report`: **82%**, above the repository's 70% threshold.
- `ruff check app cli.py main.py webui test docs/skill`: passed.
- `python -m compileall -q app cli.py main.py webui test docs/skill`: passed.
- `git diff --check`: passed.

The initial sandboxed run blocked five existing localhost HTTP receiver tests. Re-running the complete suite with local networking allowed passed; no tests or assertions were weakened. Live provider tests stayed disabled, and no Redis test server was configured. The suite emits existing dependency/serialization warnings.

## Scope / review notes

Two files only; no dependency, configuration, generated-media, or release changes. This is a parser fix for correctly blank-line-separated SRT cues, not a rewrite of malformed-SRT recovery. No live TTS, model inference, or rendered-video acceptance is claimed.

Suggested review order: the parser's cue-state condition, then the complete tuple assertions in the regression test.

AI-assisted implementation and an independent automated code review; the reproduction and verification above were executed locally.

Contributor: Chen Quan / 陈全.


harry0703:
Merged — thank you for the focused fix, clear reproduction, and thorough regression tests!

We reproduced the bug locally using real SRT files and confirmed that the fix preserves both the original cue timing and timestamp-containing text. We also checked multiline cues, Windows line endings, EOF handling, and compatibility with the current main branch. All three CI checks have now passed.

The separate colon-splitting behavior in downstream subtitle correction is pre-existing and outside this parser fix, so it does not block this PR.

Thank you again for the contribution!

PR #1392, opened 2026-09-16
fix(asgi): normalize configured CORS origins to the Origin header shape

c020627:
`CORS_ALLOWED_ORIGINS` values are compared verbatim against the `Origin` request header, which RFC 6454 fixes to `scheme://host[:port]` — no trailing slash, no path, lowercase scheme and host, and the protocol's default port omitted. A value copied out of the browser address bar is `https://frontend.example/`, and `https://frontend.example:443` is what people write when they are being explicit about the port; neither can ever equal the `Origin` the browser actually sends.

The failure mode is silent rather than loud. The configured trusted frontend is refused with the same 403 as an attacker, the preflight is rejected with 400, the browser only reports a CORS error it does not show the response body for, and the single server-side trace is a `blocked untrusted browser origin` warning that points at the request instead of at the configuration. Nothing in that path says "your whitelist entry can never match".

`parse_cors_allowed_origins` now folds every item into the shape a browser actually sends and drops values that cannot match any `Origin` at all, with an explicit warning naming the offending value, instead of installing a whitelist entry that silently never matches. Normalization rebuilds the origin from the validated `hostname` and `port` rather than lowercasing `netloc`, so default ports are omitted, non-default ports and IPv6 brackets are preserved, and invalid ports are rejected.

A malformed entry can no longer take the process down with it. Parsing happens inside `try`, so the `ValueError` that `urlsplit()` raises on `https://[::1` — or that reading `port` raises on `:99999` — collapses to `None`: the entry is skipped with a warning and the remaining origins are kept. This matters because the parser runs at import time; before this, a single typo in the environment variable stopped the entire API from starting.

The existing handling of surrounding whitespace and empty comma-separated items is unchanged, as are the `*` compatibility mode and the default same-origin policy for an empty configuration, so the documented examples keep working unchanged.

### Evidence

**Parser reverted to the previous revision, new tests kept** — exactly the new cases fail and nothing else moves, so this is a pure increment:

```
3 failed, 9 passed
FAILED test/services/test_asgi_cors.py::TestASGICORS::test_default_port_configuration_admits_the_trusted_frontend
FAILED test/services/test_asgi_cors.py::TestASGICORS::test_explicit_default_port_is_folded_into_the_origin
FAILED test/services/test_asgi_cors.py::TestASGICORS::test_malformed_entry_is_dropped_without_losing_valid_origins
```

The first of them fails with the exact error reported from a fresh process: `ValueError: Invalid IPv6 URL`.

**Malformed entries in a fresh process**, with `CORS_ALLOWED_ORIGINS=https://valid.example,https://[::1,https://frontend.example:443,https://bad.example:99999`:

| | before | after |
| --- | --- | --- |
| `import app.asgi` | `ValueError: Invalid IPv6 URL` — the API cannot start | loads, `['https://valid.example', 'https://frontend.example']` |
| dropped entries | crashed on the first one | 2 warnings, one per entry |

**End-to-end through the real path** (environment value → `parse_cors_allowed_origins` → `configure_browser_access` → request), with `CORS_ALLOWED_ORIGINS=https://frontend.example:443/` and `Origin: https://frontend.example`:

| request | before | after |
| --- | --- | --- |
| cross-origin `GET` | 403, no `Access-Control-Allow-Origin` | 200, `Access-Control-Allow-Origin: https://frontend.example` |
| cross-origin preflight | 400, no `Access-Control-Allow-Origin` | 200, `Access-Control-Allow-Origin: https://frontend.example` |
| untrusted origin | 403 | 403 (unchanged) |

**Common shapes through the parser:**

| configured value | before | after |
| --- | --- | --- |
| `https://frontend.example/` | `['https://frontend.example/']` | `['https://frontend.example']` |
| `HTTPS://Admin.Example` | `['HTTPS://Admin.Example']` | `['https://admin.example']` |
| `https://frontend.example:443` | `['https://frontend.example:443']` | `['https://frontend.example']` |
| `http://frontend.example:80` | `['http://frontend.example:80']` | `['http://frontend.example']` |
| `https://frontend.example:0443` | `['https://frontend.example:0443']` | `['https://frontend.example']` |
| `https://frontend.example:8443` | kept | kept |
| `https://[::1]:3000` | kept | kept, brackets preserved |
| `https://frontend.example:bad` | kept, never matches | dropped + warning |
| `https://frontend.example:99999` | kept, never matches | dropped + warning |
| `https://[::1` (malformed) | `ValueError` at import | dropped + warning |
| `https://a.example/,https://a.example/` | both kept | one entry |
| `localhost:3000` (scheme missing) | kept, never matches | dropped + warning |
| `ftp://files.example` | kept, never matches | dropped + warning |
| `*` | `['*']` | `['*']` |

### Verification

- With the new tests kept and the parser reverted: `3 failed, 9 passed` — only the new cases.
- `test/services/test_asgi_cors.py`: 12 passed.
- Full suite: green on CI (ubuntu, Python 3.11 and 3.13 — run `35330109368`). Locally the only failures are pre-existing Windows environment issues that reproduce identically on unmodified `main` (`test_asgi_static_files.py::test_does_not_serve_symlink_to_file_outside_tasks`, which is also not part of the Windows smoke list), so they are unrelated to this change.
- `ruff check app cli.py main.py webui test docs/skill` with the locked `ruff==0.15.21`: passes.
- No new dependency, no new module; the branch is `2 files changed, 185 insertions(+), 2 deletions(-)` against `a76b61d`.


harry0703:
Thank you for the improvement and the detailed tests! We reproduced the trailing-slash issue locally and confirmed that this change fixes it: the trusted request and preflight now both succeed, while untrusted origins remain blocked. The 14 related CORS/static-file tests and Ruff checks also passed.

Before merging, could you please address two remaining edge cases in `_normalize_allowed_origin`?

1. **A malformed entry can now prevent the entire API from starting.** For example, `CORS_ALLOWED_ORIGINS=https://valid.example,https://[::1` causes `urlsplit()` to raise an uncaught `ValueError` during module import. We reproduced this in a fresh process; the previous parser still allowed the application to load. Please catch URL-parsing errors, warn and skip the invalid entry, and preserve the other valid origins.

2. **Normalization still leaves some entries unable to match a browser Origin.** For example, `https://frontend.example:443/` is retained as `https://frontend.example:443`, but the browser sends `https://frontend.example`; the request/preflight still return 403/400. The same applies to HTTP port 80 and ports with leading zeroes. Invalid ports such as `:bad` or `:99999` also pass through without a warning. This is an existing limitation rather than a new regression, but it falls within the normalization this PR introduces. Please validate the hostname/port, omit default ports, and normalize supported host forms rather than simply lowercasing `netloc`.

Regression coverage for malformed IPv6 entries mixed with valid origins, explicit default ports, and invalid ports would help keep this robust. Thanks again for working on this—we would be happy to take another look once these cases are covered!

c020627:
Both cases are fixed — thanks for the precise reproduction, the `ValueError` on import was the more serious of the two.

**1. A malformed entry can no longer stop the API from starting**

The whole parse now happens inside `try`. `urlsplit()` raising on a malformed IPv6 literal, and reading `hostname`/`port` raising on a non-numeric or out-of-range port, both collapse to `None` — the entry is skipped with the existing warning and the rest of the list survives. I checked `:bad` and `:99999` too: they came through unvalidated before, so they now take that same path.

Fresh process with `CORS_ALLOWED_ORIGINS=https://valid.example,https://[::1,https://frontend.example:443,https://bad.example:99999`:

| | before | now |
| --- | --- | --- |
| `import app.asgi` | `ValueError: Invalid IPv6 URL` | loads, `['https://valid.example', 'https://frontend.example']` |
| dropped entries | crashed on the first one | 2 warnings, one per entry |

**2. Normalization now produces the Origin the browser actually sends**

Instead of lowercasing `netloc`, the origin is rebuilt from the validated `hostname` + `port`:

| configured | before | now | browser sends |
| --- | --- | --- | --- |
| `https://frontend.example:443/` | `https://frontend.example:443` | `https://frontend.example` | `https://frontend.example` |
| `http://frontend.example:80` | `http://frontend.example:80` | `http://frontend.example` | `http://frontend.example` |
| `https://frontend.example:0443` | `https://frontend.example:0443` | `https://frontend.example` | `https://frontend.example` |
| `https://frontend.example:bad` | kept, silently | dropped + warning | — |
| `https://frontend.example:99999` | kept, silently | dropped + warning | — |
| `https://frontend.example:8443` | kept | kept | `https://frontend.example:8443` |
| `https://[::1]:3000` | kept | kept, brackets preserved | `https://[::1]:3000` |

End to end with `CORS_ALLOWED_ORIGINS=https://frontend.example:443/` and `Origin: https://frontend.example`: GET **200** with `Access-Control-Allow-Origin`, preflight **200** — previously 403 / 400. An untrusted origin still gets 403 with no `Access-Control-Allow-Origin`, so the boundary is unchanged.

**Tests** — the three cases you asked for:

- `test_malformed_entry_is_dropped_without_losing_valid_origins` — malformed IPv6 (`https://[::1`) in the middle of valid origins, plus the two invalid ports; asserts the valid origins survive and each dropped entry warns exactly once.
- `test_explicit_default_port_is_folded_into_the_origin` — `:443` / `:80` / `:0443` folded, `:8443` and `[::1]:3000` preserved.
- `test_default_port_configuration_admits_the_trusted_frontend` — the end-to-end case above.

With the parser reverted to `adf2095` and only the new tests kept: **3 failed, 9 passed**, the first failure being the exact `ValueError: Invalid IPv6 URL`. `test/services/test_asgi_cors.py` is **12 passed** after the change, and `ruff check` over the CI target list is clean.

I left `https://a.example/some/path` alone rather than rejecting it — the path is dropped when rebuilding from `hostname`, which keeps a value copied from a deep page working, and CORS cannot express path scoping anyway. Say the word if you would rather see that warned about instead.

harry0703:
Thank you for the fixes and the thorough regression coverage! We verified the malformed-entry handling in a fresh process and confirmed that default-port normalization allows trusted requests and preflights while keeping untrusted origins blocked. All 22 related local tests and the CI checks passed.

Merged! Keeping paths out of the resulting origin makes sense, since CORS is origin-based. We can handle the remaining uncommon host-normalization cases separately. Thanks again for your contribution and for following up on the review so carefully!

PR #1394, opened 2026-09-17
fix(llm): replace the retired Groq default model with openai/gpt-oss-120b

RooMax:
Groq no longer serves llama-3.3-70b-versatile. New Groq users who leave the model name empty get a rejected request, and the WebUI model dropdown shows no match for the registry default. Point the default at openai/gpt-oss-120b, which Groq currently lists, and update the test fixtures that referenced the old id.


harry0703:
Merged — thank you for updating the Groq default model and the related test fixtures!

We checked the replacement against Groq’s official deprecation notice and verified the default-model request path locally, including empty values and explicit model overrides. The related local regression tests passed, and all three CI checks are now green.

The change has been rebased and merged, preserving your original commit and authorship. Thanks again for helping keep the Groq integration working for new users!

